Issue #17, opened 2019-09-01
Scala Metals fails


mardukbp:
Scala Metals is a language server that communicates over JSON-RPC. It can be installed with

```
curl -Lo coursier https://git.io/coursier-cli && chmod +x coursier
./coursier bootstrap org.scalameta:metals_2.12:0.7.0 -o metals -f`
```

The contents of my `servers.yml`:

```
langservers:
  scala:
    - /Users/marduk/metals
```

I launched Jupyter Lab as follows:

```
node miniconda3/share/jupyter/lab/staging/node_modules/jsonrpc-ws-proxy/dist/server.js --port 3000 --languageServers servers.yml 

jupyter lab
```

I created a Scala notebook (using the [Almond kerne](https://almond.sh/)l) and in `~/Library/Caches/org.scalameta.metals/global.log` I got the following error:

```
INFO  tracing is disabled for protocol LSP, to enable tracing of incoming and outgoing JSON messages create an empty file at /Users/marduk/Library/Caches/org.scalameta.metals/lsp.trace.json
INFO  Starting Metals server with configuration: MetalsServerConfig(
  bloop-protocol=BloopProtocol(auto),
  glob-syntax=GlobSyntaxConfig(uri),
  status-bar=StatusBarConfig(off),
  slow-task=SlowTaskConfig(off),
  execute-client-command=ExecuteClientCommandConfig(off),
  show-message=ShowMessageConfig(on),
  show-message-request=ShowMessageRequestConfig(on),
  no-initialized=false,
  compilers=PresentationCompilerConfigImpl(false,None,None,Map(scala/collection/mutable/ -> mutable., java/util/ -> ju.),Ascii,true,true,true,true,true,20,SECONDS),
  http=false,
  input-box=false,
  icons=none,
  statistics=StatisticsConfig(default)
)
INFO  logging to file /.metals/metals.log
Sep 01, 2019 9:45:07 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.nio.file.AccessDeniedException: /.metals
```

So it seems that Metals cannot create the directory `.metals` in the current workspace. I did some research and it appears that the workspace for the language server is set by `jsonrpc-ws-proxy/dist/server.js`.

Metals works in all the major editors, but it would be awesome to have it available in JupyterLab. See [Integrating a new editor](https://scalameta.org/metals/docs/editors/new-editor.html) for details.

[...]

PR #35, opened 2019-09-19
[wip] add traitlets/entry_points-configurable proxy wrapper

bollwyvl:
@krassowski here's the more full-blown, traitlets-configurable version of the `jupyter-server-proxy`-based wrapper for the node thing. [Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-traitlets-proxy?urlpath=lab%2Ftree%2Fexamples%2FPython.ipynb) looks good. i did a lot of this without any internet, so bear with me! 

Some more opinions:
- dependencies
  - bump to `jupyterlab==1.1.4`
    - match more tooling, etc.
  - move most `@jupyterlab` and `@phoshpor` (and dependencies) into `peerDependencies` 
  - resolve `yarn.lock`
- running the linter more, tried to clobber some errors 
  - ended up adding some more `.catch(console.warn)` stuff due to aggresive promise catching
    - to be honest, I really haven't dug into how the code works much, hopefully no problems arise!
- scrapped the whole `binder/` folder, just moved everything to root (actually was able to get rid of a few things)
  - locally, i just run `postBuild`, so it's a tad more than what is needed in binder, but as i'm flitting about between projects, it really helps to have The One Command... which also happens to give us nice demos!

Lots more to do, but wanted to get something out there to see.

Known remaining stuff:
- [ ] add an `r` lsp connector (and figure out how to get it working in binder)
- [ ] more docs (though those were easy low-hanging fruit while off-line!)
- [ ] tests (especially of the config behavior, those are awful)
- [ ] do some looking on windows
- [ ] investigate whether the JSON-RPC can be done in python
  - but really, there are a lot of language servers written in js/ts, so you'd end up needing node anyhow, much of the time, but _I_ don't want to be the guy that decides that


src/adapters/jupyterlab.ts
```diff
@@ -175,19 +182,20 @@ export abstract class JupyterLabWidgetAdapter {
         'have been initialized'
       );
     });
-
   }
 
   protected async connect_document(virtual_document: VirtualDocument) {
-    virtual_document.foreign_document_opened.connect((host, context) => {
+    virtual_document.foreign_document_opened.connect((_host, context) => {
```
krassowski: I am not familiar with this convention, if it ain't WIP, would you mind pointing me to some material to educate myself?

bollwyvl:
It wasn't actually complaining about it (obviously), but might have
throwing something while i was working on appeasing the linter. the _ gets
rid of the "unused variable" warning (squiggly blue) that inevitably
happens when overloading things in exciting ways.

krassowski:
This looks very promising. Thanks for working on this :)

- I see that this won't fix the relative/absolute path problem (while a custom proxy, such as linked in #2 could). We would either need to modify `jsonrpc-ws-proxy` or expose the absolute path to the frontend. This is quite a high priority given that most of the servers do not work on 100% of their capacities without the ability to look around in the local filesystem to analyze imported modules (and the Scala server even refuses to start #17)
- Do I understand correctly that from the user perspective there will be two ways to get a server working: by autodetection (for a pre-defined subset of servers), and from the command line when starting jupyter lab (the `language_servers` argument); or is it populated via a config file?
- > ended up adding some more .catch(console.warn) stuff due to aggressive promise catching
  
   that's perfect, should be there in the first place!

src/adapters/jupyterlab.ts
krassowski: @bollwyvl, thanks for the explanation. My linter does not complain about such unused variables, but if it makes the code cleaner for others, that's perfectly reasonable to change it.

krassowski:
One more thought (thinking aloud): how crazy of an idea is to have the auto-detection logic defined in a form of a JSON file? I see that the pre-defined server connectors/detectors either start from `shutil.which` or `app.find_node_module` and then return a command composed from the detected paths. Having a JSON structure describing this process might make it easier for others to add the servers they need.

This way, we could make it configurable (hackable) on the frontend-side, in the advanced settings editor. There is a serious disadvantage though, it might be difficult to keep this secure if we allow for modifications from the frontend. Probably not worth the risk...

bollwyvl:
> I see that this won't fix the relative/absolute path problem

Indeed, the setup in this PR is really only tested for `localhost:8888/lsp`, and with binder, at `/user-13456/lsp`. It seems everything is a screaming mess once you get off the relatively quiet path of `pyls`, but I'll defer to you.

Rather than monkeying with `jsonrpc-ws-proxy`, it's indeed probably worth finding (or writing, if need be) an `async` python version that gives the notebook server and language kernel servers (see what i did there), but ultimately the user, more control. In the purest form of the Jupyter Architecture, you'd want the language servers to be users of the `ContentsManager`, as that is what the user sees. Or even more ridiculous, the `IDrive` content provider for lab google docs/github. But yeah, let's get files working first.

>  autodetection (for a pre-defined subset of servers), and from the command line when starting jupyter lab (the language_servers argument); or is it populated via a config file?

yes. The docs should make that more clear.

- if this repo were to migrate over to a jupyter org, it should cover (and test) the languages needed to contribute to jupyter: python (for server/extensions) and Web Stuff, which is pretty `node`-centric once you get out of the JSON standard(ish) things. 
  - but you'd still want to overload the defaults
- Python-adjacent languages can have the easiest packaging story, 
  - because `entry_points` just kind of work, and versioning Mostly Works. 
    - We didn't bring them into most core jupyter pieces for various reasons, and could indeed take that opinion out entirely, 
      - but they are pretty nice
- the `traitlets` config system from which `notebook` and `jupyterhub` get their config can be configured by:
  - changing some "well-known" files, e.g. `${PREFIX}/jupyter_config.py`, `~/.jupyter/jupyter_notebook_config.json`
  - putting unique `.py` and `.json` files into well-known locations, e.g. `~/.jupyter/`, `$PREFIX/etc/jupyter`, `$(cwd)`, etc.
  - passing traitlets directly in a the command line as JSON, e.g. `--NotebookApp.notebook_dir="/"` (bad idea)
    - encoding complex JSON  is not fun

> should be there in the first place!

bless the linters

krassowski:
Could this PR support a single pip-installable `jupyterlab-lsp` package which would enable to get rid of `jupyter_notebook_config.py`? I mean [this](https://github.com/jupyterhub/jupyter-server-proxy/blob/master/docs/server-process.rst#specifying-config-from-python-packages). Or maybe it already does?

krassowski:
> Rather than monkeying with `jsonrpc-ws-proxy`, it's indeed probably worth finding (or writing, if need be) an async python version that gives the notebook server and language kernel servers (see what i did there), but ultimately the user, more control. In the purest form of the Jupyter Architecture, you'd want the language servers to be users of the `ContentsManager`, as that is what the user sees. Or even more ridiculous, the `IDrive` content provider for lab google docs/github. But yeah, let's get files working first.

I see the point with `ContentsManager`. On the other hand, a quick fix, without touching the internals of the `jsonrpc-ws-proxy`, is to propose a new switch for `jupyterhub/jupyter-server-proxy`; let's call it `prepend_filesystem_path`; I need to look up the exact config variables but I hope that you get the idea - the LSP server would end up with an absolute path to files, and the path would never be exposed to the frontend.

bollwyvl:
Yeah, the entry_points part lets the core (or other pip installable
packages) not mess with the traitlets stuff... But the traitlets win, as
they are more user-serviceable.

krassowski:
But can we have both? Apologies if this is a question with an obvious answer, I have very little experience with traitlets.

bollwyvl:
Yes, both are supported in the prototype. Working kinda like:
https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html

krassowski:
Hi @bollwyvl - a breaking change on master:
- https://github.com/krassowski/jupyterlab-lsp/commit/23c19fc6e5f891cd037ec514737e5ff85f18d291
changes the endpoint name for "application/typescript" → to "typescript". I tried to use typescript servers to test some new functionality but stumbled upon the endpoint having a slash. I ended up rewriting entire language detection logic (no longer depends on the jumper); in short it defaults tothe second part of mimetype if the first is "application" or "text"; some known quirky mimetypes can be manually mapped (as it is now with R mimetypes, which are reported idfferently by the kernel and by the file editor)

you may be interested to know that I got typescript working (not sure if you managed it as well, I had some problems) using:

```yaml
  javascript:
    - node
    - '/usr/local/lib/node_modules/javascript-typescript-langserver/lib/language-server-stdio.js'
  typescript:
    - node
    - '/usr/local/lib/node_modules/javascript-typescript-langserver/lib/language-server-stdio.js'

(obviously the path should be adjusted/autodetected).
```

There might be something more - let me know if you would like help with resolving conflicts for this PR. I look forward to having this in.

*PS:* Some future outlook: it seems that a serverside mirroring of the directory structure might be required to make the LSP servers happy (if we stay with the canonical LSP and do not use the sourcegraph extensions). I see this as a temporary directory having the virtual files created and udpated by the server-side extension on requests from the frontend. The path to the temporary dir would be reported to the frontend and included in LSP requests (I assume this be safe as we would not be revealing the real path to the resources on the server).

bollwyvl:
> it defaults tothe second part of mimetype

This probably won't work, long term. The canonical truth is probably whatever the `languageId` actually ends up in the LSP messages, so we'd need a lookup table that some `jupyterlab-lsp-<language name>` package could register. If going jupyter-forward, thing best thing I can think of is `codemirror_mode`, e.g. the canonical name that codemirror uses for _it's_ lookup table. the kernels provide it as part of the kernelspec, and documents usually know it.

> language-server-stdio.js

good find!

> resolving conflicts for this PR

i think i'm mostly sorted. It would probably be very helpful to start using `lint:check` in CI, as pre-formatted source merges much better... trying that out presently.

> I look forward to having this in.

Yeah, sorry, a lot of stuff moving right now... this is up there, but there are So Many Things.

> temporary directory having the virtual files created and updated

yeah, we're a bit out of my ken at this point: if it doesn't at least work simply against files, and we can't make it work against the contents manager API somehow, we're in for a world of hurt. i started that discussion with the pygls folks, we'll see if something comes of it!

bollwyvl:
The irony of lint failing due to `.travis.yml` formatting: utter.

Having typescript and python working now is a really great demo.

bollwyvl:
[binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-traitlets-proxy?urlpath=lab%2Ftree%2Fexamples%2FPython.ipynb) is also apparently good to go.
Screenshot from 2019-09-24 22-06-29

src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -311,10 +317,9 @@ export abstract class JupyterLabWidgetAdapter
       serverUri: 'ws://jupyter-lsp/' + language,
       languageId: language,
       // paths handling needs testing on Windows and with other language servers
-      rootUri: 'file:///' + PathExt.join(this.server_root, this.root_path),
+      rootUri: 'file:///' + PathExt.join(this.server_root),
```
krassowski: Redundant join?

krassowski:
> This probably won't work, long term. [...] codemirror_mode [...] the kernels provide it as part of the kernelspec, and documents usually know it.

Agreed, CodeMirror's mode seems like a very good choice for now as it is already unified between the notebook and file editor (mimetypes are not, for R the file editor has a different mimetype than the notebook). Two points:
 - we still need a custom lookup table, as there is no guarantee that the mode name is the same as preferred LSP language name
 - In long-term, I feel uneasy about relying on the presentation layer metadata (i.e. CodeMirror's mode) to select the LSP language - I am afraid that there are languages which use the same syntax highlighting but would benefit from different language servers. I may be wrong but I would be cautious here.
- there is also "preferred kernel language" metadata in the FileEditor model (I guess used to start up a kernel/console attached to a file); I find this a bit unreliable (for typescript it gives javascript), but it may be the way to go
 - the version of the language may matter in the future (e.g. Python 2 vs Python 3). This, however, seems to be left out of the LSP specification, so nothing for us to do now.

I will leave the mime-type based detection for now, but I am happy to change it later (or accept a PR).

>  [if] we can't make it work against the contents manager API somehow

I was just thinking about this. Making a custom ContentsManager would overwrite others, wouldn't it? We would have to have $n$ custom ContentsManagers (for Google Drive etc), wouldn't we? If this is the case, maybe we could amend the API? CC @blink1073

bollwyvl:
Well, to be fair, kernel-backed things are often rightly a different
mimetype, because kernel builders can't help building in custom syntax.
It's so fun!

ContentsManager, like many other parts of the notebook architecture, are
extended by specialization, not composition, by default. Indeed, a
MetaContentsManager, kind of like nbviewer's URL scheme, would be a great
piece of technology. So you'd see something like this in your ui:

- localfiles/
- github/
- s3/
- ipfs/

GitHub, s3, ipfs would not be able to list their top level keys, so you'd
need some way to dig into those, and cache/bookmark ones you know existed.
Otherwise it should Just Work.

It could be making the language servers talk to such a beast is just a
matter of passing the contents URL, e.g
http://localhost:8888/api/rawcontents/localfiles/src/index.ts
<http://localhost:8888/api/contents/localfiles/> for the document URI (and
writing the handler without the JSON envelope). Haven't tried it!

bollwyvl:
Found this in the pyls examples:
https://github.com/palantir/python-jsonrpc-server/blob/develop/examples/langserver.py

bollwyvl:
Actually, the better example is here, basically performing the behavior of `jsonrpc-ws-proxy` (albeit for exactly one LS):
https://github.com/palantir/python-jsonrpc-server/blob/develop/examples/langserver_ext.py

Instead of being a `jupyter-server-proxy` configuration, it would become a `serverextension`, which means it should spin up a bit faster, in that it would be in-process with the rest of the `NotebookApp`. However, we'd _still_ likely want all the `nodejs` detection, etc. given how many things are implemented over on that side of the fence. But more importantly, we'd have the ability to inspect, and potentially mangle, the `documentURI`s coming over the wire, and potentially at some future time allow splitting/injecting things to hand off to kernels.

Adopting this would:
- remove the dependency on `jupyter-server-proxy` (and its ten transitive/direct dependencies, `p-j-s` only depends on `future`)
- remove the hard dependency on `jsonrpc-ws-proxy` (and therefore `nodejs`)
- allow us to re-use the existing notebook test infrastructure, such as it is

I think i need to at least investigate this, though I'd likely replace their extra loop and thread with the [asyncio native](https://docs.python.org/3.5/library/asyncio-subprocess.html) syntactic sugar, if it can be made to work. 

Indeed, if we find configurations become more complex, we might end up wanting to implement something akin to the [kernelspec](https://jupyter-client.readthedocs.io/en/stable/api/kernelspec.html) approach.

krassowski:
> But more importantly, we'd have the ability to inspect, and potentially mangle, the documentURIs coming over the wire, and potentially at some future time allow splitting/injecting things to hand off to kernels.

This would be cool!

I would like to start working on releasing the next version soon, so ideally I would see this in a subsequent PR, rather in this mega-PR, but I agree that this would be great to have, ideally before 1.0.

krassowski:
Okay, this is strange - I did not expect GitHub to show this PR as merged after merging #46...

bollwyvl:
Yeah, based it off that and it contained all the commits. You could roll
back and squash...

PR #36, opened 2019-09-21
Split the LSP features into separate modules

### LINKED ISSUE #23: Split the CodeMirrorAdapterExtension into smaller modules
So far I have tried to follow the patterns from the upstream CodeMirrorAdapter of [wylieconlon/lsp-editor-adapter](https://github.com/wylieconlon/lsp-editor-adapter). Their design is alright for the number of LSP features implemented; however, the added complexity of the virtual document model introduced to support notebooks and nested foreign code makes it no longer sustainable to keep everything in one large class. As I plan to implement more LSP functions, it does not seem to be reasonable to keep inheriting from the upstream class as there are too many divergent needs when it comes to JupyterLab integration.

I think that having one file per LSP function and using a manager pattern (as used for the extensions in JupyterLab) would be a good idea. A similar idea for the reduction of complexity of the index was suggested in #22.

The initial interface idea:
```typescript
interface ILSPFunction {
  /** Connect event handlers to the editor, virtual document and connection(s) **/
  register(
    virtualEditor: IVritualEditor,
    virtualDocument: IVritualDocument,
    connection: ILSPConnection
  );
  /** Will allow the user to disable specific functions **/
  isEnabled: boolean;
  /** Return JupyterLab commands to be registered;
  * intended for single-use in index.ts (during extension registration)
  **/
  get commands(): IJupyterLabCommand;
  /** Return the context menu commands to be added during extension registration.
  * The comands would be grouped by target context menu (like Cell or FileEditor).
  **/
  get contextMenuCommands();
  /** Remove event handlers on descruction **/
  remove();
}
```

Using such a pattern would also make it easier to unit test specific functions.

There is a small performance penalty, as some things which share logic in a single handler will need to be repeated for the individual functions and multiple event handlers will be needed; I do not think that this will be significant.

krassowski:
Fixes #23. While the CodeMirrorAdapterExtension mega class is no longer there, there is still some work to be done in moving the commands a context menu entries into the feature-specific modules.


krassowski:
Ready to go!

PR #38, opened 2019-09-22
Add "rename" action

krassowski:
General notes:
 - It only works when using absolute paths and not for synced documents (e.g. rope in pyls seems to require the file to exists):
   - we may need to mirror the virtual documents in the file system to allow for seamless integration of rename action. Those files could be hidden; however, there are OS-compatibility, safety and import systems (varying across languages) concerns with such a solution;
  - notebooks are not supported for now as those would require the virtual documents to be saved on the disk (e.g. as hidden files)

TODO:
- [ ] reflect the change in the frontend editor
  - [x] for FileEditor
  - [ ] <s>for Notebook</s> not in this PR
- [x] pre-populate the renaming dialogue with the existing token value
- [ ] <s>make use of the [prepare rename request](https://microsoft.github.io/language-server-protocol/specification#textDocument_prepareRename) to validate the new name</s> not in this PR

This PR does **not** implement workspace-wide edits (rename action fails if multiple files need to be edited). A shared mechanism for opening files in background is needed for this as well as for #41.

### DIFF
--- a/src/adapters/codemirror/feature.ts
+++ b/src/adapters/codemirror/feature.ts
@@ -18,6 +18,19 @@ export enum CommandEntryPoint {
   FileEditorContextMenu
 }
 
+function toDocumentChanges(changes: {
+  [uri: string]: lsProtocol.TextEdit[];
+}): lsProtocol.TextDocumentEdit[] {
+  let documentChanges = [];
+  for (let uri of Object.keys(changes)) {
+    documentChanges.push({
+      textDocument: { uri },
+      edits: changes[uri]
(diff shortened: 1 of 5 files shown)


krassowski:
For future reference: I had an annoying issue with prettier/ts-ignore intersection here:

Screenshot from 2019-10-01 11-47-55
Screenshot from 2019-10-01 11-47-33
so it was always "wrong".

Soultion (fixed on master):
Screenshot from 2019-10-01 11-49-42

bollwyvl:
Looking forward to playing with that!

PR #50, opened 2019-10-01
Add R to Binder

krassowski:
Adds R kernel and R languageserver specs config to Binder environment; all 100% conda-forge based.


krassowski:
Relevant binder repositories:
- https://github.com/binder-examples/r_with_python
- https://github.com/binder-examples/multi-language-demo

krassowski:
Tests on Mac, and the binder startup takes very long with installation of the language server...

I wish there was something like wheels for R's CRAN...

krassowski:
@bollwyvl I cannot get this working - after one or two initial messages being passed ok, I keep getting BrokenPipeError:

```
  File "[...]/pyls_jsonrpc/streams.py", line 100, in write
    self._wfile.flush()
BrokenPipeError: [Errno 32] Broken pipe
```

(it is nicely handled in `_write_to_lsp`). I did lots of debugging but had to give up after a few hours. I would appreciate if you could take a fresh look.

bollwyvl:
I had a similar experience with the Julia language server, and didn't even
try it on binder! It would be great if we could avoid going the
only-works-in-this-one-special-docker route, but we might eventually end up
there.

The other tack would be to put a request out on conda-forge/staged-recipes
for someone to attempt packaging the language server... Even if everyone
doesn't use conda, when it works, it works well, especially for these kinds
of cross-language use cases.

I'll take a look, but we may have to wait until someone that really cares
comes along.

bollwyvl:
Somebody has at least started the conda approach here:

https://anaconda.org/gwerbin/r-languageserver

I'd have to dig into their recipe, some, but seems like a thing we could
likely land on conda-forge.

bollwyvl:
I've done very little R work, so not sure how much help I can be! But I can test things more!

So if it passes the tests, but then dies after a while: perhaps we need some more substantial per-language server fixtures? Are we talking hundreds of messages or just initialization? Perhaps something akin to http://hackage.haskell.org/package/lsp-test, but with hypothesis?

setup.cfg
```diff
@@ -49,6 +49,7 @@ jupyter_lsp_spec_v0 =
     vscode-html-languageserver-bin = jupyter_lsp.specs:html
     vscode-json-languageserver-bin = jupyter_lsp.specs:json
     yaml-language-server = jupyter_lsp.specs:yaml
+    r-language-server = jupyter_lsp.specs:r
```
bollwyvl: seems to often be written `r-languageserver`... but not a big deal.

krassowski:
I found the culprit: the first read from the stdin of R process intercepts the executed command (`'"languageserver::run(debug=TRUE)"\n`), which is [blindly (and therefore "wrongly") interpreted by the python-jsonrpc-server as a header](https://github.com/palantir/python-jsonrpc-server/blob/develop/pyls_jsonrpc/streams.py#L41-L56) which leads to the [read loop being broken](https://github.com/palantir/python-jsonrpc-server/blob/develop/pyls_jsonrpc/streams.py#L26-L27) as there is no additional blank line, nor the contents following.

krassowski:
@bollwyvl what is the command / server for Julia that you have tried?

krassowski:
And that is not enough, because the `python-jsonrpc-server` sends the header in UTF8 while is it should be sent as ASCII, and in a separate package to exactly mimic the behavior of the vscode implementation. Additionally, there is the encoding issue - the encoding headers are not present in vscode and lead to a failure of R language server (and possibly others as well)

krassowski:
The headers issue is [fixed on master of R languageserver](https://github.com/REditorSupport/languageserver/issues/77#issuecomment-537409247).

Also, looking at https://pypi.org/project/json-rpc/ to replace `python-jsonrpc-server`, as the latter does not get too much of attention recently (and requires palantir contributors agreement to contribute which stalls some of the PRs).

setup.cfg
krassowski: Right. Changed it, thanks!

krassowski:
Travis was failing due to: `ERROR: this R is version 3.2.3, package 'languageserver' requires R >=  3.4.0` with newer version of languageserver (which has the necessary fixes). I gave up on this, having not the best experiences with enabling newer version of R on Travis (this is a pain). Let's focus on having a nice demo in binder first, using conda ecosystem - and if it works we can try to use it for Travis (or other CI) as well.

krassowski:
Binder finally works with R. I use conda only because the binder runtime was too slow (it installed RStudio, IRKernel, etc); it is still slow, but at least can start up. Maybe getting more dependencies of R languageserver from conda could speed this up a bit.

.travis.yml
```diff
@@ -5,6 +5,7 @@ cache:
     - .yarn-packages
     - $HOME/.cache/yarn
     - $HOME/.cache/pip
+    - $HOME/lib/R/site-library
```
krassowski: probably no longer needed cache conda instead?

krassowski:
It works in R files, though it seems that the extractors do not work on master currently; this is a separate issue.

py_src/jupyter_lsp/specs/r_languageserver.py
```diff
@@ -0,0 +1,7 @@
+from .utils import ShellSpec
+
+
+class RLanguageServer(ShellSpec):
+    key = cmd = "R"
```
bollwyvl: you'd proabably want:
```
    key = "r-languageserver"
    cmd = "R"
```

bollwyvl:
> Maybe getting more dependencies of R languageserver from conda could speed this up a bit.

conda-forge is on the case :female_detective: :
- https://github.com/conda-forge/staged-recipes/issues/9707
  - https://github.com/conda-forge/staged-recipes/pull/9709
    - https://github.com/conda-forge/r-languageserver-feedstock
    - https://github.com/conda-forge/r-collections-feedstock
    - https://github.com/conda-forge/r-xptr-feedstock

There's a bit of a hitch in the :robot: so they aren't published yet, but that will get resolved within a countable number of hours, no doubt... I'd hold off on merging this PR until we can at least evaluate whether this path is viable.

>  nice demo in binder first, using conda ecosystem - and if it works we can try to use it for Travis 

Yeah, that's a part where I _can_ come in handy, as I have been refining my conda-based stuff to get to browser testing [robotlab on azure](https://github.com/robots-from-jupyter/robotlab/blob/master/azure-pipelines.yml) (and indeed, we'll definitely be figuring out how to ship a robot language server over there, soon enough).

bollwyvl:
>  blindly (and therefore "wrongly") interpreted by the python-jsonrpc-server as a header

Everything is in process now, so we can try a subclass, and/or contribute it upstream. I'll wait until i can actually get one running that exhibits the condition.

bollwyvl:
> https://pypi.org/project/json-rpc/ to replace python-jsonrpc-server

sounds good. I picked p-j-s because it would be a transitive dependency of p-l-s, and therefore already installed (at least in our demos and tests). But if it's not implementing the spec we should switch, for sure. If it's just how vscode implements it I'm kinda meh on it, but the more libre the better, no?

bollwyvl:
> what is the command / server for Julia that you have tried?

I tried a couple-step approach to get a demo working:
```
# env.jl
using Pkg
Pkg.add("IJulia")
Pkg.add("LanguageServer")
```
And then tried some stuff from [julia-vscode](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/languageserver/main.jl)

Never really got that far...

krassowski:
I tried the conda-forge package on another branch ([binder](https://mybinder.org/v2/gh/krassowski/jupyterlab-lsp/full_conda?urlpath=lab)) and it works, but linting in R does not work (I guess because the fixes we need were not released yet).

bollwyvl:
Bravo! Next chance I get, I'll look into getting an azure PR out so we can
start using this.

Is there anything we can do on our end to overcome the issue that's fixed
on r-languageserver master, e.g. subclassing the writer and changing some
messages?

I had a look at json-rpc, and it's not as straightforward as might be hoped.

krassowski:
> Is there anything we can do on our end to overcome the issue that's fixed
on r-languageserver master, e.g. subclassing the writer and changing some
messages?

Yes, we can: https://github.com/krassowski/jupyterlab-lsp/commit/2511731ae6e7924e941bd3dd93fcb0d2ac43423b, but it seems that R in the "full_conda" branch still does not work on binder :(

A nice thing to have would be a way to pass the LSP and backend errors to the frontend to make debugging easier on binder; I think that it just worked with the node implementation.

bollwyvl:
> A nice thing to have would be a way to pass the LSP and backend errors to the frontend to make debugging easier on binder; 

Right. I think this fits into the "control" channel mentioned a few times. Perhaps we add some more routes like:
- `/lsp/r` LSP-over-websocket (as currently)
- `/lsp/r/status` plain-old-JSON REST
- `/lsp/r/log` logs-over-websocket

> I think that it just worked with the node implementation.

We can set up some testing to find cases where we return different answers from the outside (e.g. the websocket) and try to fix them. I'm just not interested in maintaining any more node-based stuff, and/or forcing end users to have to deal with it... unless they opt/are opted into it with their choice of language server.

bollwyvl:
I think you were really close: this PR-on-a-branch-off-a-PR works with R on binder.

Screenshot from 2019-10-03 21-45-23

https://github.com/krassowski/jupyterlab-lsp/pull/52

I don't know which order you want to do this in: it looks like the total delta is actually pretty small vs `master`, but it's a little hard to track. Squash merge is probably appropriate in this case!

I _think_ I can just start on the azure setup so we can actually start testing `r`.

bollwyvl:
Update on the follow-on work:

Things are not pretty in windows land. The r server works (everything
works) when run interactively, but when testing, even in the thread,
deadlocks when talking to the r server after the first message is received,
and the tests timeout. Without more servers to test, I can't tell if this
is unique to r, or will show up on more servers.

Trying to go to the happy path of ripping out pyls and just using asyncio
subprocesses fails, because the zmq loop (needed for talking to kernels) is
incompatible with awaiting file handles on windows. This will come to a
head more broadly for notebook/tornado with py38, but we're not going to
solve it from here.

Options:

- send more messages so we detect at least one message returned
  - should unblock the readline

- don't unit test r in windows
  - bad
  - but maybe a browser-based integration test will make me feel okay about
it

- also/only offer tcp-based connections in addition to stdio, use by
default with r
  - wanted to do this anyway
  - sockets are more cross-platform
  - but more open ports, bad?
  - but maybe make it the default, if we're not sure this is going to be an
issue with other servers
  - would make the spec more complicated
    - going to happen eventually (env vars, etc)
   - for servers that don't support tcp, wrap? Ugh.

krassowski:
I will give it some thought later on  (I need to go under for a while - starting a PhD programme now). If we can keep stdio-only interface that would be great IMO (the simpler the easier to maintain).

R was always uneasy to interface with from Python, especially when the threads come in, the rpy2 project has a lot of experience with this. I hoped that subprocess/stdio based approach would work well. I do not remember what rpy2 uses but I think this is a lower level communication protocol.

Is there a chance of using something else than tornado here? Like - for example - Flask? I know that jupyter stack is tornado-oriented, just thinking aloud that maybe it would be possible to solve some of the issues if we used Flask instead?

bollwyvl:
Best of luck with the PhD work!

Yes, everything is simpler on Linux... except for getting people to use it.
"Simple on Linux" but "doesn't really work on windows" is a pretty raw
compromise. We've ended up with tornado because, once it works, it works.

Yes, flask is "simpler," but also blocking and requires things like gevent
and another server (like tornado, for example) to actually be a
production-grade server of websockets, etc. Indeed, some of my earliest
comments to jupyter (then ipython) notebook were similar, but in the end,
thoughtful asynchronous has proven to really be the only way to handle
things like multiple kernels/language servers/content providers/completions
without a massive process/thread footprint.

I'll keep chugging... Don't lose hope.

PR #52, opened 2019-10-04
More work on R/conda/binder

bollwyvl:
Screenshot from 2019-10-03 21-45-23

Builds on the upstream, apparently works. [Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/full_conda?urlpath=lab) is solid.

Also applied the python formatters, etc. And, because `r-lintr` was working, I even linted the helper script, without even knowing r! Hooray dogfood!

As I was perusing upstream, i see that `r-languageserver` also supports rmarkdown. Definitely a nice value proposition, but apparently codemirror isn't telling anybody, so it comes back `plain`. 

I'm not sure what ordering of things made a difference but it just started working against my local dev env (also linux on conda).

Let's squash merge this, if we get it working!

### DIFF
--- a/src/adapters/jupyterlab/file_editor.ts
+++ b/src/adapters/jupyterlab/file_editor.ts
@@ -10,12 +10,10 @@ import { ICompletionManager } from '@jupyterlab/completer';
 import { LSPConnector } from './components/completion';
 import { CodeEditor } from '@jupyterlab/codeeditor';
 import { VirtualFileEditor } from '../../virtual/editors/file_editor';
-import { LSPConnection } from '../../connection';
 
 export class FileEditorAdapter extends JupyterLabWidgetAdapter {
   editor: FileEditor;
   jumper: FileEditorJumper;
-  main_connection: LSPConnection;
   virtual_editor: VirtualFileEditor;
   protected current_completion_connector: LSPConnector;
 
--- a/src/adapters/jupyterlab/jl_adapter.ts
+++ b/src/adapters/jupyterlab/jl_adapter.ts
@@ -138,12 +138,6 @@ export abstract class JupyterLabWidgetAdapter
     });
   }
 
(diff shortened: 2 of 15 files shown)


src/adapters/jupyterlab/file_editor.ts
```diff
@@ -10,12 +10,10 @@ import { ICompletionManager } from '@jupyterlab/completer';
 import { LSPConnector } from './components/completion';
 import { CodeEditor } from '@jupyterlab/codeeditor';
 import { VirtualFileEditor } from '../../virtual/editors/file_editor';
-import { LSPConnection } from '../../connection';
 
 export class FileEditorAdapter extends JupyterLabWidgetAdapter {
   editor: FileEditor;
   jumper: FileEditorJumper;
-  main_connection: LSPConnection;
```
krassowski: This

src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -138,12 +138,6 @@ export abstract class JupyterLabWidgetAdapter
     });
   }
 
-  get main_connection(): LSPConnection {
-    return this.connection_manager.connections.get(
-      this.virtual_editor.virtual_document.id_path
-    );
-  }
-
```
krassowski: And this

bollwyvl:
This is the confusion i expressed over on the other pr: this targets `full_conda` rather than `master`, so if you want to bring it in there and do whatever surgery that's great. Then merge that into #50? I'm not really sure how you want to track the history, if at all.

By the by, the azure stuff is looking pretty good, and may have finally actually found an issue in windows: https://dev.azure.com/nickbollweg/nickbollweg/_build/results?buildId=367

krassowski:
Ok, I will play with that - the worst thing that can happen is one or two minor cleanup commits could be lost.

Issue #39, opened 2019-09-22
Initialization problem?


krassowski:
Both binder and my local version work now perfectly (after refactoring all the features and removing the dependency on third-party `CodeMirrorAdapter`), but they appear to only work after refreshing the window. It seems that there is a problem with initialization after connection/registration of features.

PR #53, opened 2019-10-05
Use conda installation on the R branch

krassowski:
Merging back, up the path. See #50 and #52.


krassowski:
Works on binder (except for #39).

PR #55, opened 2019-10-05
Fix #31 (LSP commands in context menu when clicked outside of editor)

### LINKED ISSUE #31: LSP context menu entries in notebook should be hidden if user clicked outside of a cell
This probably happens because the `activeCell` is checked without checking the focus of the notebook.

krassowski:
A trivial fix.

### DIFF
--- a/src/command_manager.ts
+++ b/src/command_manager.ts
@@ -131,7 +131,7 @@ export abstract class ContextMenuCommandManager extends LSPCommandManager {
 
 export class NotebookCommandManager extends ContextMenuCommandManager {
   protected tracker: INotebookTracker;
-  selector = '.jp-Notebook .jp-CodeCell';
+  selector = '.jp-Notebook .jp-CodeCell .jp-Editor';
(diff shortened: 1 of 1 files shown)


krassowski:
Before:
Screenshot from 2019-10-05 15-31-26

After:
Screenshot from 2019-10-05 15-38-40

An unintended side-effect: the LSP commands are now on the top (if the menu is invoked within the editor):
Screenshot from 2019-10-05 15-38-58

bollwyvl:
Given that they are token-based (as opposed to cell based) this is probably
appropriate.

bollwyvl:
Not in this pr, but: I'd also like to see a toolbar item that let you
enable/disable different language server ui features for the current file.
It could be right aligned, but left of the kernel name/status icon, and
also have an indicator of whether there is pending stuff... This would be
particularly useful in the typescript case which takes _upwards of a
minute_ to fully resolve all the references in, say, index.ts.

krassowski:
IMO this would belong to the statusbar, but yes, this feature would be nice to have.

PR #57, opened 2019-10-09
Add azure pipelines (part iii)

bollwyvl:


bollwyvl:
sry, still cooking on this (but did figure out the windows deal)...

PR #58, opened 2019-10-09
Add azure pipelines, remove python-jsonrpc-server

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-azure-pipelines?urlpath=lab)

Adds CI on Azure Pipelines as demonstrated over on https://github.com/bollwyvl/jupyterlab-lsp/pull/3.

The pipeline uses conda-based excursions for win/osx/linux on python 3.6/3.7. Travis is now really only uniquely running the 3.5/3.8 cases (not served well by conda, at present) and could be moved as well, using the same deadsnakes ppa, but it seemed easier to leave them both for the time being. It's about a five-minute setup to get azure going... though if you end up wanting to do automated releases, I usually prefer to have a bot account with very limited scope.

In getting the tests to run (with `r-languageserver`) on windows, I encountered a hard limitation on how windows was working with some subprocess pipes. While easy to reproduce, it was maddening to debug. I ended up cargo-culting some stuff (with attribution) to set some crazy flags, and things are running great. I've even taken out the `pytest-timeout` dependency (as it didn't generate very good test reports).

In all the deugging and source reading, I was also able to reimplement the parts of the stdio handler we need, so we can drop a dependency. Also, because the io is no longer blocking, we dispense with the threading altogether, which is on the whole simpler.

This would also be a good candidate for a squash merge! Of course no rush, with your other commitments, but I'm pretty confident in where the implementation is right now.


setup.cfg
```diff
@@ -22,7 +22,6 @@ classifiers =
 
 [options]
 install_requires =
-    python-jsonrpc-server
```
bollwyvl: Hooray!

py_src/jupyter_lsp/stdio.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class Reader(StdIOBase):
+    """ Language Server stdio Reader
+
+        Because non-blocking (but still synchronous) IO is used, rudimentary
+        exponential backoff is used.
+    """
+
+    max_wait = Float(2.0, help="maximum time to wait on idle stream").tag(config=True)
```
krassowski: Seconds?

krassowski:
It's working! Fun fact if you change the name of the pipeline while it is executed, the result gets reported back to GitHub under the new name, but GitHub still thinks the "old" pipeline is still running (see "jupyterlab-lsp-CI (1)" at https://github.com/krassowski/jupyterlab-lsp/runs/261231379).

bollwyvl:
Hooray!

Yeah, sorry the first unasked-for pr was a surprise. when you get to
deploying stuff (even to gh-pages), it _really_ wants lots of perms, so
I've been setting up dedicated gh accounts for that.

Once I get the browser testing stuff stable I'll follow up with that pr...
I've mostly gotten through diagnostics and jump (within a file)... Seems
like hover shouldn't take too long after that.

PR #60, opened 2019-10-15
Set up CI with Azure Pipelines

krassowski:


krassowski:
Azure is so invasive... it takes my account without asking if I want to use it, it creates a PRs when only asked for a branch... Not the best first impression.

PR #62, opened 2019-10-17
Add Browser Tests

bollwyvl:
This adds acceptance tests with screenshots for diagnostics and jump features. It uses [Robot](https://robotframework.org/) Firefox ESR and geckodriver from conda-forge. The overall runtime of course increases, but I feel there's a lot of value being delivered.

Other things:
- increases instrumentation, uploading more coverage/test reports from jest and acceptance tests
  - the coverage reports are still a touch wonky
- found some more issues introduced by the reader/writer implementation from #58
- add support for jsx/tsx
- some minor frontend fixes without unit tests
  - normalize case of drives on windows paths
  - handle non-list jump targets (observed with css)
  - can move these to a separate PR if desired

Caveats:
- markdown is not tested. I haven't dug into what's going on.
- browser/selenium flake is inevitable... still researching things to make it more robust (rather than just slower)
- again, probably a squash merge (even after i discarded a lot of trial-and-error commits)
- i haven't tried doing a "hard" install of all these things without conda (mostly because i help maintain all of the stuff there)
- i swear i'll calm down and stop making such large PRs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,7 @@ venv.bak/
 
 node_modules
 .yarn-packages
+
+atest/output/
+junit.xml
+coverage/
--- a/.prettierignore
+++ b/.prettierignore
@@ -28,3 +28,5 @@ envs
 **/.ipynb_checkpoints
 **/*.egg-info
 .yarn-packages
+atest
+coverage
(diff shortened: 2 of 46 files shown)


krassowski:
Thank you!

PR #67, opened 2019-10-25
Compactify the language servers tables

krassowski:
Up for discussion, but I think that this is a bit easier to read.

Also, corrected second css to sass.

### DIFF
--- a/py_src/jupyter_lsp/README.md
+++ b/py_src/jupyter_lsp/README.md
@@ -14,7 +14,7 @@ them if they _are_ installed and we know about them.
 > You can disable this behavior by configuring [`autodetect`](#autodetect)
 
 Use a package manager to install a [language server][lsp-implementations]
-(also [this list][langserver]) from the table below: these implementations are
+(also [this list][langserver]) from the tables below: these implementations are
 tested to work with `jupyter-lsp`.
 
 Don't see an implementation for the language server you want? You can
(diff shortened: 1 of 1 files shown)


krassowski:
By the way, do we want to brand the python package as "jupyter-lsp" (as in readme) or "jupyterlab-lsp" @bollwyvl?

py_src/jupyter_lsp/README.md
```diff
@@ -23,18 +23,21 @@ Don't see an implementation for the language server you want? You can
...
+| bash                      |              `bash-language-server`              |
+| css<br/>less<br/>sass     |         `vscode-css-languageserver-bin`          |
+| docker                    |       `dockerfile-language-server-nodejs`        |
+| html                      |         `vscode-html-languageserver-bin`         |
+| javascript<br/>typescript |        `javascript-typescript-langserver`        |
+| json                      |         `vscode-json-languageserver-bin`         |
+| markdown                  |            `unified-language-server`             |
+| yaml                      |              `yaml-language-server`              |
+
+| language | `conda install -c conda-forge` |         language-specific package manager         |
```
bollwyvl: Yeah, that's fine... just haven't gotten around to the headache of conda packaging the other ones. No doubt they have conflicting dependencies all over the place, which can be tough to manage sanely without packaging everything down to the last `left-pad`...

bollwyvl:
>  "jupyter-lsp" (as in readme) or "jupyterlab-lsp" 

I think `jupyter-lsp` is more accurate, as there's nothing inherently lab-specific about the backend piece, and someone _could_ (and no doubt _should_) re-use it for classic, nteract, or some other client. It's crucial to have _a_ frontend developed together with the backend, especially at this stage, but i could image someday splitting them.

That being said, I also do import `jupyterlab` to check its `app_dir`, but that should be made optional, and we should likely also support [jupyter_server](https://github.com/jupyter/jupyter_server) which is getting some attention right now. I could imagine certain kinds of `voila` interactive experiences being very interesting with some of the features. Will PR.

py_src/jupyter_lsp/README.md
krassowski: Once those are conda-packaged I fully supported having another column for conda package names here :+1: 

The pattern I see here is that there is a large group of markup/style/spec languages with a simple syntax, which can be reasonably implemented in JS/TS - and those are in the first table (+JS/TS itself). The more complex, turing-complete languages are harder to analyze without the dedicated toolset which is often written in the same language or distributed in language-specific channels.

PR #68, opened 2019-10-25
Fix a description issue which caused PyPI to reject

krassowski:
It seems that PyPI does not like line breaks in the description.
This PR also silences a warning about "author_email" being required if "author" is defined.

However, this still uses the main README (see https://test.pypi.org/project/jupyter-lsp/0.6.0a3/). I think that we may prefer to use the Python one instead. What do ou think @bollwyvl?

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,15 @@
+# Releasing `jupyterlab-lsp` and `jupyter_lsp`
+
+Releases may require building both the python package and nodejs packages.
+
+## Updating Versions
+
+The version for PyPI must be updated in two places:
+
+- `py_src/jupyter_lsp/_version.py`
+- `azure-pipelines.yml`
+
+The version for npm must be updated in two places:
+
+- `package.json`
+- `azure-pipelines.yml`
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,7 +2,7 @@
 
 Releases may require building both the python package and nodejs packages.
 
-## Updating Versions
+## Updating Version Strings
 
 The version for PyPI must be updated in two places:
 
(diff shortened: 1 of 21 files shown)


krassowski:
It fails on an unchanged variable with the version number in pipelines. I tried to read it from the _version.py file but could not get it working yet

krassowski:
Any ideas? Or do we have to stick with changing it in two places?

RELEASE.md
```diff
@@ -0,0 +1,15 @@
+# Releasing `jupyterlab-lsp` and `jupyter_lsp`
+
+Releases may require building both the python package and nodejs packages.
+
+## Updating Versions
+
+The version for PyPI must be updated in two places:
```
bollwyvl: I'm a little torn: on the one side, having the single-source-of-truth is really good, but on the other hand, having it in one "real" place and one "test" place is kind of attractive. In the past, I've added an "integrity" script that checks the two (or n) places rather than doing heroic things to make them all read from one place.

bollwyvl:
Linux fail is temporary conda flake, just needs rekick, but am away from
keyboard...

bollwyvl:
Hooray, back to green.

Some paths forward:

If we indeed want to have exactly the two version strings (which I feel pretty strongly _should_ remain separately floating... the server side will eventually be _done_, once it's extensible) we can:
- [ ] make a `scripts/install_dev.py` which will install the built packages in ci, based on what it finds in the canonical versions

If we want to keep them split, but use one "test" place to test one "truth" place, we can: 
- [ ] make a `scripts/integrity.py` (or just make it a test) that checks they agree
  - this can also check other things like requirements agreement, etc.
  - it can also check the latest git tag, which can be nice

krassowski:
> If we indeed want to have exactly the two version strings (which I feel pretty strongly should remain separately floating... the server side will eventually be done, once it's extensible)


I agree. Merging this for now in order to keep the things moving and release the first version with PyPI.

PR #69, opened 2019-10-26
Robot flake fixes 🤖❄️ 

bollwyvl:
Been noticing some flake on #67, #66: sorry to make so many :x: .

- [x] always sleep before attempting to start feature tests
- [x] connect to an unused port, rather than hardcoding `18888`
- [x] some docs for robot vars
- [ ] ???

I'd say if this passes, it can be merged... it will be an on-going battle to keep the browser tests happy.


bollwyvl:
First of linux builds passed :trumpet:

Issue #70, opened 2019-10-26
remove hard dependency on jupyterlab


bollwyvl:
Responding to https://github.com/krassowski/jupyterlab-lsp/pull/67#issuecomment-546533942, we should probably keep the `jupyter-lsp` name for the python side, and not have a hard dependency on lab... it would be great if other clients could use it!

bollwyvl:
closed by #71

PR #71, opened 2019-10-26
Remove hard lab dependency

bollwyvl:
For #70.

This removes the `jupyterlab` dependency, leaving `notebook` (and adding the missing `setuptools`). Lab will still get installed when following the dev instructions (which I've also revisited a bit).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,17 +21,16 @@ classifiers =
     Programming Language :: Python
 
 [options]
-install_requires =
-    notebook
 package_dir =
     = py_src
+
 packages = find:
 include_package_data = True
 zip_safe = False
 
-tests_require =
-    pytest
-    pytest-cov
+install_requires =
+    notebook
+    entrypoints
 
 [options.packages.find]
 where =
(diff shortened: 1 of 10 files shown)


setup.cfg
```diff
@@ -42,7 +41,7 @@ jupyter_lsp_spec_v0 =
     bash-language-server = jupyter_lsp.specs:bash
     dockerfile-language-server-nodejs = jupyter_lsp.specs:dockerfile
     javascript-typescript-langserver = jupyter_lsp.specs:ts
-    pyls = jupyter_lsp.specs:py
+    python-language-server = jupyter_lsp.specs:py
```
krassowski: This was not reflected in the pyls specs. Do we want to keep this change or was it accidental?

bollwyvl: It actually doesn't matter a _whole_ lot, but `python-language-server` is the full name of the distribution. just a thing i noticed while working through it. It just means something else can't _also_ be called `python-language-server` in entry_points.

krassowski: Well, I think that we tend to refer to `pyls` as `pyls`, because there is also https://github.com/microsoft/python-language-server (which is much less capable). Ironically, they do not have the `python-language-server` on PyPI, which is used by https://github.com/tartley/pyls.

bollwyvl: It's true, all very confusing, but i'll favor the pypi being the canonical:
https://pypi.org/project/python-language-server/

The csharp-based one isn't really installable any way other than as part of vscode to my understanding. Also re: `python-jsonrpc-server`. But then, naming this is one of the hard things, no?

PR #72, opened 2019-10-26
More ci flake fixes 🤖📦❄️

bollwyvl:
For #68 flake... will squash merge once passing.
- [ ] Reloads the page (and re-disable alerts) after resetting application state 
  - [x] instead, just try to make more sure all tabs are closed
    - [x] heavy reloading of lab triggers a bunch of `nodejs` processes which are bad on windows
- [x] always try to open the file list
- [x] retry troublesome package manager requests :three: times
- [x] also bring back yarn cache
- [x] shut down all kernels (slowing down windows, probably)
- [ ] ???

### DIFF
--- a/atest/01_Editor.robot
+++ b/atest/01_Editor.robot
@@ -11,6 +11,7 @@ ${CM CURSORS}     css:.CodeMirror-cursors:not([style='visibility: hidden'])
 
 *** Test Cases ***
 Bash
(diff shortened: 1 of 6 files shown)


bollwyvl:
Ok, two consecutive passing builds seems good enough to help support #73 etc.

Issue #61, opened 2019-10-15
Selectively enable/disable LSP features


krassowski:
Extracting discussion from #55, quote by @bollwyvl :

> Not in this pr, but: I'd also like to see a toolbar item that let you
enable/disable different language server ui features for the current file.
It could be right aligned, but left of the kernel name/status icon, and
also have an indicator of whether there is pending stuff... This would be
particularly useful in the typescript case which takes _upwards of a
minute_ to fully resolve all the references in, say, index.ts.

I think that a statusbar would be a good place for it (so that we have it for both files and notebooks).

bollwyvl:
Very good point (though I usually turn it off)... I _thought_ the editor
also exposed a toolbar entry point like the notebook, but I could be
imagining things...

It's too bad there's no way to know if the language server is still
"busy"... Responses just seem to come back whenever. Sure would make
testing easier!

bollwyvl:
Riffing on the status bar idea, perhaps we have it show, e.g.:

(icons for placement only
```
  [⏳ Python]
  [🎧 Python] # ready
```

krassowski:
I like this idea. It could be in a popover but a new tab/sidebar widget would work as well.

krassowski:
I am working on a basic status indicator right now.

bollwyvl:
Cool! Excited to see what you come up with.

My thinking on the ugly ASCII mockup: I think more
popups/hovers/(context)menus, aside from being harder to test, can make it
harder to inform people what choice they are making when it comes to
settings. "What the heck is a diagnostic and why would I hide it?" Hence I
usually encourage anything remotely settings related (e.g would persist
after restart) to err on the side of being more verbose and discoverable.

dclong:
Agree that there should be a way to disable LSP (completely) just in case things are broken or completion is slow so that users to want to temporary disable it without uninstalling the extension. 

Also, it's best it there's a way to restart lsp without restarting jupyterlab. This is critical for Docker based users.

PR #73, opened 2019-10-26
Add an empty properties to settings schema

bollwyvl:
Even though we don't have any settings to configure, we need an empty `properties` so that the settings editor can draw it without failing (and mucking up the UI).

Will also add a test...

### DIFF
--- a/atest/00_Smoke.robot
+++ b/atest/00_Smoke.robot
@@ -1,6 +1,17 @@
 *** Settings ***
+Suite Setup       Set Screenshot Directory    ${OUTPUT DIR}${/}screenshots${/}smoke
 Resource          Keywords.robot
 
 *** Test Cases ***
 Smoke
     Capture Page Screenshot    00-splash.png
+
+Settings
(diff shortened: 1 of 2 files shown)


bollwyvl:
Still flaking on conda... needs #72

bollwyvl:
Well, #72 helped, but still getting some flake (windows R and JSX or something).

I think we need the status/feature indicator, e.g. https://github.com/krassowski/jupyterlab-lsp/issues/61#issuecomment-542370182, to give us something concrete to wait on. This will get especially important for diagnostics, as they can take a _very_ long time if they actually resolve references.

bollwyvl:
We should probably get this in before the next npm release... The
almost-empty settings page looks pretty boring, but breaking the settings
page is pretty bad.

krassowski:
Oh, I get it now (could not understand what was the problem so set it aside for now). Thanks for the additional info. Merging.

bollwyvl:
Hooray!

Issue #76, opened 2019-10-28
Reorganize client source with lerna and typescript projects


bollwyvl:
To easing proposing the client-facing parts of this codebase to @jupyterlab/jupyterlab, the repository could be re-organized as a monorepo with smaller, independently testable, extensible, and replaceable components with well-defined inter-package interfaces.

This would not block any feature development, but should help keep PRs more focused and less likely to create conflicts.

### step  1️⃣
- move `src` into `packages/jupyterlap-lsp`
- add `tsconfigbase.json` to be used by sub-package `tsconfig.json`s
- create new private lerna `package.json` 
  - owns common activities, like `lint` and `build`
- create metapackage as a composite of the other packages
  - this is the only place `tsc` will be invoked
- change `build` to use the metapackage
- otherwise no substantive code changes

### step  2️⃣
- break up `jupyterlab-lsp` into `language-server` and `language-server-extension`
  - these will need new `name`s at this point
  - `language-server`
  - add a `tokens.ts` and an `ILanguageServerManager`
  - centralize `CommandIds`
- identify opportunities for lazy-loading to reduce impact on `vendor-main` with `await import`
  - the `extension` will end up being in `main`, and just about everything from the `manager` will likely be lazy-loadable, e.g. the first time a notebook/document is loaded. on subsequent loads, it will already be cached.

### step  3️⃣
- break out features, as relevant, into smaller sub-packages, adopting `Manager` registration APIs
  - especially per-language or per-activity features, e.g. `language-server-ipython`, `language-server-notebook`
  - this will make it more obvious how to contribute new language/features
  - this also paves the way to contribution by extension, e.g. gutters and codeactions from @deathbeds/lintotype, hierarchical symbols (perhaps as part of @jupyterlab/toc)

bollwyvl:
remember when i said i would be better about doing more manageable chunks of work? well...

I started taking a look at what it would take to add a hierarchical symbol viewer, and it got a little more... involved than I thought... basically a good chunk of [2] and [3]:

https://github.com/krassowski/jupyterlab-lsp/compare/master...bollwyvl:add-hierarchial-symbol-viewer

here's the yak shave, each one of which should probably be a pr:
- add a new package boilerplate, but can't talk to lsp
  - add a token on `lsp` for the new package to require
    - refactor how the connection manager work so that it could allow adding a capabilit
      - bring in `lsp-ws` to start hacking on it
[...]

PR #77, opened 2019-10-28
upgrade typescript, use composite, remove require usage (76  0️⃣ )

bollwyvl:
> This can wait for review until after the next release, but will be the starting point for some additional PRs.

To simplify the changes in step 1 of #76, this prepares things a bit, such that the move into `packages` will be straight moves, rather than move-and-changes.

Concretely:
- [x] upgrades typescript to what lab is currently using 
  - [x] resolve `yarn.lock`
  - [x] appeases some `empty block` lint fails
- [x] cleans up some imports (use of `require`, re-exporting of CodeMirror in a few places)
- [x] add `composite: true`
- [x] adds the `-b` flag to `build` and `watch`


bollwyvl:
One windows flake, but otherwise good!

krassowski:
Okay, the manual Azure Pipelines restart does not work due to https://developercommunity.visualstudio.com/content/problem/784369/rerun-failed-jobs-transitions-task-to-canceled.html.

bollwyvl:
Yeah. Bummer. Can you "enqueue new build"? I don't have perms (which is
fine! More looking for more chores) It might not get picked up in gh,
though....

krassowski:
It went through smoothly at https://dev.azure.com/krassowskimichal/jupyterlab-lsp/_build/results?buildId=84&view=logs&j=a1481901-c0b1-5ce5-074c-5d9d313a786f. I reviewed this step [0] (no comments) and I'm going to merge it now so we have this in before the presentation.

I am not sure if I want to merge step [1] already or first finish off other PRs and then merge it.
I will take a closer look tonight.

PR #78, opened 2019-10-28
Statusbar

krassowski:
Currently "uninitialized virtual documents" messages are false alarms as the "`serverInitialized`" event is not emitted (fixing requires creating a fork for the connection implementation - got the code, but not the time to publish)


krassowski:
It would be helpful if we could know if there is an LSP server for given language (from the frontend). Maybe the dear endpoints could return 404 or something similar?

bollwyvl:
> fixing requires creating a fork for the connection implementation

While contributing to the upstream is always good, I concur, it may be best, at least in the near term, to fork, at least until we figure out what needs to change and/or have a larger API surface exposed... for example, I wanted to look into adding hierarchical symbol support, but the initialization message is hard-coded inside the function.

This would be another compelling thing to do on #76 at any point after stage 1 (currently working on it, but again, should in no way block the 0.6.0 line).

> It would be helpful if we could know if there is an LSP server for given language (from the frontend). Maybe the dear endpoints could return 404 or something similar?

Hm, websockets are funny, and not expecting to 404... i think the connection would already open.

We could have `/lsp/` be a plain-old REST endpoint that returns a list of all the available specs, likely something we'd want to do anyway.

bollwyvl:
Wooee, that's quite a status thing. Also for settings icons, we're gonna need an icon. How do these hit you? (helps to zoom way out, as the status bar is only 24px)

| ![logo][] | ![logo_2][] |
|-|-|

> font is [FiraCode](https://github.com/tonsky/FiraCode

The color ones could play a bit more (like the underlines could be another color), but for status bar icons you pretty much just get tone...

[logo_2]: https://user-images.githubusercontent.com/45380/67822576-305e4180-fa97-11e9-99a1-bd754402b608.png
[logo]: https://user-images.githubusercontent.com/45380/67821904-0572ee00-fa95-11e9-9a1e-a45ad52387c0.png

krassowski:
I would prefer the leftmost one. In the rightmost one, it is not so obvious that the pipe on the left represents 'L'. I think that actually this reflects that we configure _language servers'_ features, rather than the _language server protocol_ itself. On the other hand, "LSP" acronym is what ppl are already familiar with (so not sure what is the best course of action).

In the long term, I was also thinking about changing the branding more in the direction of "IntelliSense" (TM of Microsoft) as this is essentialy what we provide (literally depending on small pieces of VSCode) - and more explanatory to a novice user. Something like "JupyterLab code inteligence", "JupyterLab language intelligence" etc. But this is not very important now, and if you think this is better to have an icon already, I would go with the leftmost one!

Is FiraCode already used by JupyterLab or would it be a new dependency?

bollwyvl:
> IntelliSense

:-1: there's basically no reason to introduce that, not least because of :tm:. If anything, it would be in the docs, 

>> _In addition to Jupyter Kernel features like dynamic completion, you can enhance the JupyterLab Notebook and File Editor with code insights akin to the Intellisense:tm: features found in VSCode, including static code completion, diagnostics, symbol-based navigation, etc. by installing the appropriate and configuring a [Language Server](https://langserver.org/)_

But nah, we shouldn't have that string in the code anywhere.

> ppl are already familiar with

Consider the new user using hub and lab for a class and learning to code for the first time: they won't know any of the acronyms or branding, but would still substantially benefit from the features. the challenge is to reveal the features as simply and naturally as possible... and we won't have to worry about branding at all once we get into jupyterlab/jupyterlab :monkey: 

>  if you think this is better to have an icon already

yeah, if only for testing, and making it easier for people to suggest a new one (single file PR). I want to be able to find it on the page, and write docs about it. otherwise in _Advanced Settings_ we just get a gear. Also we're probably going to need a _family_ of icons, very possibly one per language server, but that's further down the road.

> Is FiraCode already used by JupyterLab or would it be a new dependency?

not without [jupyterlab-fonts](https://github.com/deathbeds/jupyterlab-fonts) :ghost:. we'd cook it down to svg anyway.

krassowski:
@bollwyvl I got an initial version of the statusbar ready.

- I moved the detailed text message into a popup; if we get a dedicated set of status icons (representing states such as "waiting", "connecting", "initializing", "initialized") we could further reduce the verbosity (but I would keep an explanatory word or two).
- the icons are placeholders - feel free to provide suggestions/swap them for something better
- I will implement your idea of having a new tab/sidebar for more detail configuration (this is a good one) later on - for this PR I only got the status message.
- I see the popup as an intermediate step between quick actions and the full-blown configuration tab. I would like to see three-four buttons here, like:
   - "Restart LSP connections/reconnect" (after an internet connection issues for users working remotely, i.e. JupyterHub/binder),
  -  "Pause the LSP functionality, my laptop battery is dead and the node_modules indexing is driving my CPU crazy!"
   - "Show me the full config, please!"

A typical picture from a single file editor:
Screenshot from 2019-10-31 11-04-44

An advanced case of a notebook with cells including other languages:
Screenshot from 2019-10-31 11-04-54

krassowski:
Merging as-is, as it provides a huge usability improvement already (users get a visual indication that the extension is enabled and my deduce why it might not be working as intended and can provide us with some useful info for debugging).

I gotta go and do some science, will be back on Saturday.

bollwyvl:
Science hard, my friend!

PR #79, opened 2019-10-28
add lerna, metapackage, integrity checks, and move src (76   1️⃣  )

bollwyvl:
For #76, continuing on #77, this does the Big Move of `src` and starts setting up the tooling for further work. It may or may not be appropriate to squash merge this: i tried to be concise but ended up creating a few more commits than I would have liked, but it's still reasonable to start working on step 2.

- [x] Moves `src` into a new sub-package
 - as few other changes as i could manage
- [x] creates new, unpublished "dev-only" root package
  - [x] move all the test-time language servers here
  - [x] move most `scripts` (not `test`) here
- [x] adds new, unpublished `metapackage`
  - [x] includes `references` in `tsconfig` to `jupyterlab-lsp`
    - keeping these versions in-line is a bit of a chore, and should probably be managed with an integrity script a la lab
- [x] changes the lab build strategy for ci (`pack`, `install`, `build`) vs binder (`link`, `install`, `build`)
  - this will likely have to change some as we introduce packages that aren't labextensions
- [x] add integrity script to check `metapackage` versions and refs (also other things)
- [x] update docs


bollwyvl:
Boo, need to chomp the line break on the variables in the yaml...

bollwyvl:
Huzzah, mostly-working pass. Despite my desire to keep this a manageable review, the `metapackage` demanded that I start thinking more about the integrity stuff, so I've gone ahead and added that to CI... still want to do some stuff on here, will start a punchlist above.

bollwyvl:
I had another go through the docs... given there is just the one "real" package and the metapackage, there isn't a lot to say that's new, but I did catch (and add a test for) the version of the package listed in `/README.md`. If this goes green, it's probably ready for review...

bollwyvl:
Yep, probably done on this for the time being. No rush to review. I'll check out #78 and see if there's a way to do server discovery that makes sense.

README.md
```diff
@@ -120,7 +120,7 @@ jupyter labextension update @krassowski/jupyterlab-lsp
 Use install command (update does not seem to work) appending `@version` to the extension name, like this:
 
 ```bash
-jupyter labextension install @krassowski/jupyterlab-lsp@0.5.0-rc.0
```
krassowski: Well, the point here was to demonstrate the RC installation. It's a pity I rushed through with soft release, though I do not feel that we need this integrity check for this part - other than this example, we do not need the version in the README. Though I am also happy for it to stay - as-is.

bollwyvl:
Either way! Sometimes it's great to have the version in the readme, others,
more so much...

PR #81, opened 2019-10-30
Add server status endpoint

bollwyvl:
For https://github.com/krassowski/jupyterlab-lsp/pull/78#issuecomment-546907659

This adds a REST-ish (whatever) JSON endpoint for getting the current state of the server sessions at `/lsp` (no trailing slash).

<details>
<summary>example response</summary>
<pre>
{
  "version": 0,
  "sessions": [
    {
      "languages": [
        "bash",
        "sh"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "css",
        "less",
        "scss"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "dockerfile"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "html"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "javascript",
        "jsx",
        "typescript",
        "typescript-jsx"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "json"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "markdown",
        "ipythongfm"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "python"
      ],
      "handler_count": 1,
      "status": "started",
      "last_server_message_at": "2019-10-30T04:30:53.069124+00:00",
      "last_handler_message_at": "2019-10-30T04:30:52.219183+00:00"
    },
    {
      "languages": [
        "r"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    },
    {
      "languages": [
        "yaml"
      ],
      "handler_count": 0,
      "status": "not_started",
      "last_server_message_at": null,
      "last_handler_message_at": null
    }
  ]
}
</pre>
</details>

Not sure what else we're looking for at this point.

There is a JSON schema (for testing, not presently hot validating) which we could generate typescript with, if that helps.

Everything it reports is already evented through `traitlets`, so this could relatively easily be Yet Another WebSocket, but we'll cross that bridge when we have to, I suppose, as it would require a breaking change to the URL schema, probably. Perhaps `lsp/server/python` and `lsp/servers` or something.

It also currently doesn't detect error states in a meaningful way, but we could potentially infer them by looking at the queue lengths, perhaps.

### DIFF
--- a/py_src/jupyter_lsp/schema/servers.schema.json
+++ b/py_src/jupyter_lsp/schema/servers.schema.json
@@ -1,0 +1,48 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$ref": "#/definitions/servers-response",
+  "definitions": {
+    "servers-response": {
+      "type": "object",
+      "properties": {
+        "sessions": {
+          "type": "array",
+          "items": { "$ref": "#/definitions/session" }
+        }
+      },
+      "required": ["sessions"]
+    },
+    "nullable-date-time": {
+      "oneOf": [{ "type": "string", "format": "date-time" }, { "type": "null" }]
+    },
+    "session": {
+      "additionalProperties": false,
+      "required": [
+        "languages",
+        "handler_count",
+        "status",
+        "last_server_message_at",
+        "last_handler_message_at"
+      ],
+      "properties": {
+        "languages": {
+          "type": "array",
+          "items": {"type": "string"},
+          "uniqueItems": true,
+          "minItems": 1
+        },
+        "handler_count": { "type": "integer", "minValue": 0 },
+        "status": {
+          "type": "string",
+          "enum": ["not_started", "starting", "started", "stopping", "stopped"]
+        },
+        "last_server_message_at": {
+          "$ref": "#/definitions/nullable-date-time"
+        },
+        "last_handler_message_at": {
+          "$ref": "#/definitions/nullable-date-time"
+        }
+      }
+    }
+  }
+}
--- a/py_src/jupyter_lsp/schema/servers.schema.json
+++ b/py_src/jupyter_lsp/schema/servers.schema.json
@@ -1,21 +1,32 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$ref": "#/definitions/servers-response",
+  "title": "jupyter_lsp server status response",
+  "description": "describes the current state of (potentially) running language servers",
   "definitions": {
     "servers-response": {
       "type": "object",
       "properties": {
         "sessions": {
           "type": "array",
+          "description": "a list of servers that are, could be, or were running",
           "items": { "$ref": "#/definitions/session" }
+        },
+        "version": {
+          "type": "integer",
+          "description": "the version of the schema",
+          "enum": [0]
         }
       },
-      "required": ["sessions"]
+      "required": ["sessions", "version"]
     },
     "nullable-date-time": {
+      "description": "a date/time that might not have been recorded",
       "oneOf": [{ "type": "string", "format": "date-time" }, { "type": "null" }]
     },
     "session": {
+      "title": "Language Server Session",
+      "description": "a language server session",
       "additionalProperties": false,
       "required": [
         "languages",
@@ -26,20 +37,29 @@
       ],
       "properties": {
         "languages": {
+          "description": "languages supported by this Language Server",
           "type": "array",
-          "items": {"type": "string"},
+          "items": { "type": "string" },
           "uniqueItems": true,
           "minItems": 1
         },
-        "handler_count": { "type": "integer", "minValue": 0 },
+        "handler_count": {
+          "title": "handler count",
+          "description": "the count of currently-connected WebSocket handlers",
+          "type": "integer",
+          "minValue": 0
+        },
         "status": {
+          "description": "a string describing the current state of the server",
           "type": "string",
           "enum": ["not_started", "starting", "started", "stopping", "stopped"]
         },
         "last_server_message_at": {
+          "description": "date-time of last seen message from the language server",
           "$ref": "#/definitions/nullable-date-time"
         },
         "last_handler_message_at": {
+          "description": "date-time of last seen message from a WebSocket handler",
           "$ref": "#/definitions/nullable-date-time"
         }
       }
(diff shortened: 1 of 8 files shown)


krassowski:
Can we get a list of all auo-detected + manually configured language servers, e.g. under "available_servers"?

krassowski:
Also, I'm not sure how to identify the sessions on the frontend. Can we use intercept the filename and use it as an identifier?

bollwyvl:
Yeah, so this is revealing a few things deeper in the model that I haven't really documented (which this will further complicate). I should probably draw a picture of it. Any qualms about going "full lab" and using [jupyterlab-drawio](https://github.com/QuantStack/jupyterlab-drawio)?

>  get a list of all auo-detected + manually configured language servers

this _is_ all the detected (potential) servers
 - they only thing you're not getting is `argv`, which won't be super helpful to the user
 - a display name would be good, but some of them are "cute", or just confusing, but frankly, mostly people care about the language
- there is only one session per language server, as there's only one process per language server

> Also, I'm not sure how to identify the sessions on the frontend. 

there isn't much to identify. Each websocket (`handler_count`) gets _all_ of the messages from _all_ of the like-languaged servers it can talk to
- there should _probably_ only ever be zero or one websocket per language server
  - exactly one `initialze` request should be made
  - the request ids should probably be unique across the whole system
  - everything else should _probably_ be managed on the client, based on the request ids (when appropriate)
  - some things (like diagnostics) don't even have a request id, because they just _are_... hence the uri check

> Can we intercept the filename and use it as an identifier?

- we don't do any message inspection (much less parsing!)
  - i want to add that feature, once we have the first compelling use case for it (this might be it)
  - so the hook would look something like

```py
def load_jupyter_server_extension(nbapp):
    # obviously we wouldn't do it this way, but what would someone else do?

    # another extension developer might need to IOLoop.call_later
    mgr = nbapp.language_server_manager

    @mgr.on_message(languages=".*", methods=r"textDocument/did(Open|Close)")
    def do_thing_with_message(
        # would be nice if i could generate this from a JSON schema
        message: LSPMessage, 
        # it would be nice if these had common read/write API 
        sender: Union[LanguageServerSession, LanguageServerWebSocketHandler]
    ):
        open_file_uris = list(mgr.open_file_uris)

        try:
            method = message["method"]
            uri = message["textDocument"]["uri"]
        except:
            return

        if method == "textDocument/didOpen" and url not in open_file_uris:
            open_file_uris.append(uri)
        elif method == "textDocument/didClose" and url in open_file_uris:
            open_file_uris.remove(uri)
        else:
            return

        mgr.open_file_uris = open_file_uris
        # ...and then it would go off an do something on the reporting channel
```

krassowski:
Okay, I get it now, I misread the example. I though that sessions correpond to connections, but they reallly correspond to language servers (and those are actually sessions for the backend). Sorry about that!

krassowski:
Now I see that you got this covered in schema ("`a list of servers that are, could be, or were running`"), so that is fine no need to rename.

However, I think that it would be useful for the users to know the name of the language server (and to have it here so we can expose it in the UI). For example, is the `pyls` running or Microsoft's `python-language-server`? This is of course not so important yet, but once other power-users will start playing around with settings and swapping the LSP servers to their liking this could become an annoyance.

Even better, we could require (or allow) an additional filed in the LSP specification with the URL to the repo or documentation of the server - this way we could hint the users (directly in the UI) where to go to discover server-specific configuration.

bollwyvl:
And further, this could be the canonical place where we put the mime type
mappings.

display_name: pyls
docs_url: https://...
help_url: https://...
mime_types:
  - text/python
  - text/x-ipython

krassowski:
Sounds good. Btw, jupyterlab-drawio requires some attention, I use it daily but there are some annoying bugs I wish I had time to fix.

krassowski:
Thank you! I will use it for the status bar tonight!

bollwyvl:
groovy.

I'll catch up a bit and see where we can go... Please do shout out if there
are things that would be helpful to add aside from what we've already
mulled over...

py_src/jupyter_lsp/schema/servers.schema.json
```diff
@@ -0,0 +1,68 @@
+{
```
krassowski: I am experimenting with augmenting the statusbar with the server extension status; I was thinking about generating typescript interfaces from this schema e.g. with https://www.npmjs.com/package/json-schema-to-typescript. Any experiences or thoughts on this @bollwyvl?

bollwyvl:
Yes, absolutely, I put that as a to-do on the wip branch that has all of
the good stuff like display name and mime types. Indeed, there's probably
an opportunity to reduce the amount of special typescript we have to write
for each language if we can capture it in a declarative way where the
server is created... For example, what will the complexity be like to
generate the synthetic Scala files we need from notebooks? If we can get
that part out of typescript AND python and into JSON, then we're that much
closer to being able to scale the contribution process.

For now, I'd favor the back end holding the canonical copy, as it is the
part that is has to actually generate the stuff... Or we raise it to a
top-level "schemas/" and copy it out to both places at build time. This is
another one that should eventually be _done_, and should change more slowly
than either the front end or back end, for everyone's sanity.

I haven't explored the various options in a while for which actual tool to
use, having recently been interested in the opposite direction... I guess
it's a matter of determining which one gives us the features and naming
we'd most like to use.

I think the form it would take would be a new, separate
"packages/language-server-schema/src" that contains a .ts or .d.ts (as well
as a copy of the schema in raw form) and can be imported by whichever
packages need them. In the past, I've often found it useful to have the
main imported index.ts _not_ be generated, and expose a few key types
simply (e.g LanguageServerSpec), with the rest available under an
importable namespace, (e.g Schema.InstallHelp).

If you want to take this on, by all means, I can react to whatever you end
up with... But you may wish to already look at what's over on the pr rather
than what's on master.

bollwyvl:
#90 is ready for review: if you like, I can take a crack at generating some types into a new package, or you can go after it closer to the work where you want it...

PR #84, opened 2019-10-30
Migrate from lsp-editor-adapter to (stripped) lsp-ws-connection.

krassowski:
This allows us to:
- remove some dependencies, especially conflicting CodeMirror versions
- remove lots of unused code, since we got the adapter part rewritten
- start customizing the connection behavior to our liking

Here is the link to the repo: https://github.com/krassowski/lsp-ws-connection. We may want to pull this in as a package, though there is a slight difference in the license (ISC), so not sure about this.

I also removed `jsonrpc-ws-proxy` - high time to have the transition finished off.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "eslint-plugin-prettier": "^3.0.1",
     "husky": "^3.0.9",
     "javascript-typescript-langserver": "^2.11.3",
-    "jsonrpc-ws-proxy": "0.0.5",
     "lerna": "^3.13.2",
     "precise-commits": "^1.0.2",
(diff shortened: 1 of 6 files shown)


krassowski:
I will follow with the removal of harcoded options later in the other repo, just want to get the statusbar in first (and already got the required adjustments).

bollwyvl:
Ha, barely even noticed this!

Yeah, there's no problem including it inside `packages`, and even keeping it under the same license and re-publishing it.

I think a big thing is getting the initialization message out of the closure: as we evolve what we support, we'll need to be able to add stuff in a type-safe manner.

Issue #91, opened 2019-11-03
Provide canonical rootURI in server route


bollwyvl:
Proposed on #17.

Whatever we provide in the server route introduced in #81 (and extended upon in #90) we should also provide a canonical `rootURI` so we aren't doing freaky string replacement in the client.

About the best way to do it today would be to pull it off the `nbapp.ContentsManager` at manager startup... making it work for any-old-contents-manager e.g. postgres, s3, will be really challenging, but we can probably make it work for the `FileContentsManager` and its subclasses.

We may also wish to add a whole separate endpoint with extensible paths to handle out-of-notebook-dir paths, to obviate the need for the symlink approach... I guess my inclination would be to reuse all the existing file machinery and set it up as another `IDrive`, with it being read-only by default.

PR #92, opened 2019-11-04
Add rootUri to PageConfig

bollwyvl:
For #91.

This is a lightweight approach which adds a (hopefully) canonical `rootUri` to `page_config`, where it probably belongs in the first place, rather than putting it on something language server specific. This simplifies a lot of behavior... we shall see (especially r on windows)!

### DIFF
--- a/py_src/jupyter_lsp/tests/test_paths.py
+++ b/py_src/jupyter_lsp/tests/test_paths.py
@@ -1,0 +1,14 @@
+import pytest
+
+from ..paths import normalize_uri
+
+
+@pytest.mark.parametrize(
+    "root_uri, normalized",
+    [
+        ["file:///c:/Users/user1", "file:///c:/Users/user1"],
+        ["file:///C:/Users/user1", "file:///c:/Users/user1"],
+    ],
+)
+def test_normalize_path(root_uri, normalized):
+    assert normalize_uri(root_uri) == normalized
--- a/py_src/jupyter_lsp/tests/test_paths.py
+++ b/py_src/jupyter_lsp/tests/test_paths.py
@@ -8,7 +8,7 @@ from ..paths import normalized_uri
 
 WIN = platform.system() == "Windows"
 HOME = pathlib.Path("~").expanduser()
-PY35 = sys.version_info[:2] < (3, 5)
+PY35 = sys.version_info[:2] == (3, 5)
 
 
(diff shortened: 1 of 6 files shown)


py_src/jupyter_lsp/tests/test_paths.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from ..paths import normalize_uri
+
+
+@pytest.mark.parametrize(
+    "root_uri, normalized",
+    [
+        ["file:///c:/Users/user1", "file:///c:/Users/user1"],
+        ["file:///C:/Users/user1", "file:///c:/Users/user1"],
```
krassowski: Can we also have a test case with `~/`?

bollwyvl:
Hm, the tilde gets injected by some notebook machinery... I don't think
FileContentsManager will accept a tilded path as an input for root_for, nor
will cwd ever return it, but I'll look into it!

bollwyvl:
Ah, yes, pathlib wasn't quite fully baked on 3.5... guess we need to ensure
the path actually exists to test it being expanded properly.

krassowski:
`mkdir foo`? or we could create a temporary directory with TemporaryDirectory?

I am really excited about this PR - if it works out we would be really heading in the right direction.

bollwyvl:
Yeah, but we should definitely be using vscode-uri on the front end...

py_src/jupyter_lsp/tests/test_paths.py
```diff
@@ -1,14 +1,17 @@
...
 import platform
+import sys
 
 import pytest
 
 from ..paths import normalized_uri
 
 WIN = platform.system() == "Windows"
 HOME = pathlib.Path("~").expanduser()
+PY35 = sys.version_info[:2] < (3, 5)
```
krassowski: `==`? (or `<=`)

bollwyvl:
Yup. Afk, add if you can!

krassowski:
Passed and merged. Thank you!

bollwyvl:
Hooray!

Some warnings: I didn't really document much about what it's doing (lab
does the same kinds of stuff, though), and there is no provision for
rootUri coming back unset in the front end.... Probably just give up and go
home, as pretty much nothing is going to work.... Maybe a status bar icon
kind of thing, e.g. [ls!].

Issue #99, opened 2019-11-07
Listener API


bollwyvl:
A number of issues (#95, #17, probably more) have identified a need for writing 
concrete files to disk in order for language servers to act on them. This is 
sufficient to start the conversation about allowing the pipeline of language 
messages to be intercepted by any number of agents. Currently the flow looks like:

```
                 --> queue --> stdin
                |                |
                |                v
client  <--> handler         language server
                ^                |
                |                v
                 --  queue  <-- stdout

``` 

Proposed is to allow any number of listeners to react to the messages.

```
                 --> listener --> queue --> stdin
                |                            |
                |                            v
client  <--> handler                   language server
                ^                            |
                |                            v
                 --  listener <-- queue <-- stdout

```

Not proposed is to actually _change_ or _prevent_ the delivery of messages, 
nor to encourage direct access to the handlers or servers, but rather to delay
delivery of a message until some external side effect has been completed, in the
initial case, the writing of a file to disk.

This particular case _could_ be achieved through direct use of the existing 
`ContentsManager` API, but in this case, LSP provides a change-based approach
over WebSockets that is likely more performant.

_notional API_

```py
from jupyter_lsp import on_lsp_message, LanguageServerManager

def load_jupyter_server_extension(nbapp):
    # another extension developer might need to IOLoop.call_later
    mgr = nbapp.language_server_manager

    @on_lsp_message(languages=".*", methods=r"textDocument/did(Open|Close)")
    async def do_thing_with_message(
        message: LSPMessage, 
        manager: LanguageServerManager
    ) -> None:
        # do something with a message
```

As these cannot change the message, they can be scheduled to run in "parallel".

_notional manager changes_
```py
import asyncio

class LanguageServerManager
    async write_server_message(message, language):
        await asyncio.gather([
            listener(message, self)
            for listener in listeners
            if listener.should_listen(message, {"language": language})
        ])
        for session in self.get_sessions(language):
            await session.write_message(message)
```

krassowski:
I guess we can close this one?

krassowski:
Please reopen if needed!

bollwyvl:
Oops, yes, thanks. We'll learn more as people think of things to do with
it. Probably need to expose a client-side one, too, at some point...

PR #100, opened 2019-11-08
Add message listener API

### LINKED ISSUE #102: Pin to notebook dependency to same as JupyterLab
Follow-on to: https://github.com/krassowski/jupyterlab-lsp/issues/93#issuecomment-554646608

JupyterLab 1.x requires `notebook >=4.3.1`. We should just adopt that opinion, as it streamlines the install procedure with the `conf.d` serverextension API.

bollwyvl:
This adds a listener API, mostly following #99.

- [x] decorators for (un)listening to session and handler messages
- [x] tests
- [x] configurablity?
  - [x] traitlets?
  - [x] entry_point?
- [x] docs
- [x] changelog

Definitely need to chew on this a bit more, but wanted to get it out there once it (apparently) started working. ~~For example, there's a lot of duplicated code for the two types of listeners... there could be a single decorator with a `sender="session"`, with a default of listening to _all_ messages, which would have its advantages, I suppose. I _don't_ anticipate actually adding any listeners on this PR, or at least not any that would run by default.~~

Fixes #102.

### DIFF
--- a/EXTENDING.md
+++ b/EXTENDING.md
@@ -20,7 +20,15 @@ as many Language Servers work out of the box as possible, consider
 Message listeners may choose to receive LSP messages immediately after being
 received from the client (e.g. `jupyterlab-lsp`) or a language server. All
 listeners of a message are scheduled concurrently, and the message is passed
-along once all processing has completed.
+along **once all listeners return** (or fail). This allows listeners to, for example,
+modify files on disk before the language server reads them.
+
+If a listener is going to perform an expensive activity that _shouldn't_ block
+delivery of a message, a non-blocking technique like
+[IOLoop.add_callback][add_callback] and/or a
+[queue](https://www.tornadoweb.org/en/stable/queues.html) should be used.
+
+[add_callback]: https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.IOLoop.add_callback
 
 #### Add a Listener with `entry_points`
 
--- a/EXTENDING.md
+++ b/EXTENDING.md
@@ -22,7 +22,43 @@ received from the client (e.g. `jupyterlab-lsp`) or a language server. All
 listeners of a message are scheduled concurrently, and the message is passed
 along once all processing has completed.
 
-#### Python API
+#### Add a Listener with `entry_points`
+
+Listeners can be added via `entry_points` by a package installed in the same
+environment as `notebook`:
+
+```toml
+# setup.cfg
+
+[options.entry_points]
+jupyter_lsp_listener_all_v1 =
+  some-function = some.module:some_function
+jupyter_lsp_listener_client_v1 =
+  some-other-function = some.module:some_other_function
+jupyter_lsp_listener_server_v1 =
+  yet-another-function = some.module:yet_another_function
+```
+
+#### Add a Listener with Jupyter Configuration
+
+Listeners can be added via `traitlets` configuration, e.g.
+
+```yaml
+# jupyter_notebook_config.jsons
+{
+  'LanguageServerManager':
+    {
+      'all_listeners': ['some.module.some_function'],
+      'client_listeners': ['some.module.some_other_function'],
+      'server_listeners': ['some.module.yet_another_function'],
+    },
+}
+```
+
+#### Add a listener with the Python API
+
+`lsp_message_listener` can be used as a decorator, accessed as part of a
+`serverextension`.
 
 This listener receives _all_ messages from the client and server, and prints them
 out.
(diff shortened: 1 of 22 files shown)


bollwyvl:
Also, hooray :100: !

packages/jupyterlab-lsp/src/connection_manager.ts
```diff
@@ -91,13 +91,15 @@ export class DocumentConnectionManager {
 
     const wsBase = PageConfig.getBaseUrl().replace(/^http/, 'ws');
     const rootUri = PageConfig.getOption('rootUri');
-    let socket = new WebSocket(PathExt.join(wsBase, 'lsp', language));
+    const documentUri = URLExt.join(rootUri, virtual_document.uri);
```
bollwyvl: Oh yeah, this was my bad from before: `URLExt`, not `PathExt`!

bollwyvl:
Kinda where i'm going:
image

bollwyvl:
Finally did a much-needed pass with mypy (it had been a while) to tighten some of the signatures and names up... now looking at how to handle the configurability pieces...

bollwyvl:
traitlets is in... just need to add `entry_point`s....

EXTENDING.md
```diff
@@ -0,0 +1,86 @@
...
+
+Listeners can be added via `entry_points` by a package installed in the same
+environment as `notebook`:
+
+```toml
+# setup.cfg
+
+[options.entry_points]
+jupyter_lsp_listener_all_v1 =
+  some-function = some.module:some_function
```
krassowski: The identifier on the left is any unique name, right?

setup.cfg
```diff
@@ -72,3 +72,39 @@ multi_line_output = 3
...
+ignore_missing_imports = True
+
+[mypy-ruamel.*]
+ignore_missing_imports = True
+
+[mypy-ruamel_yaml]
+ignore_missing_imports = True
+
+[mypy-jupyter_lsp.non_blocking]
+ignore_errors = True
```
krassowski: What is it need?

EXTENDING.md
```diff
@@ -0,0 +1,86 @@
...
+    @lsp_message_listener("all")
+    async def my_listener(scope, message, languages, manager):
+        print("received a {} {} message about {}".format(
+          scope, message["method"], languages
+        ))
+```
+
+`scope` is one of `client`, `server` or `all`, and is required.
+
+#### Listener options
```
krassowski: Do I understand correctly that the listeners are blocking by default (the message won't pass through until the listener is done due to await)? Should we mention this in the docs / demonstrate how to make a non-blocking listener? I don't have strong intuition on async/await as this is relatively new in Python and I think that many contributors may be in a similar place.

EXTENDING.md
bollwyvl: yes. I think we have links to the canonical docs, but we could perhaps urge that a little more strongly. thus far, none of the entrypoints names _do_ anything, other than be unique, but if they become server routes it might be more important

setup.cfg
bollwyvl: i cargo culted that, it's evil dark magic, and uses windows-specific things that may never get types.

EXTENDING.md
bollwyvl: Yes, they need to block, as the initial use case is ensuring a file exists on disk before the message is delivered to the server. This will also make it easier to implement transformers in the future, though then we would have to run them all in sequence instead of scheduling them all together.

I'd rather wait to do more docs until we have an implementation that works, as working code always finds issues!

Anyhoo, the basic idea would be:
```
import tornado.ioloop
@listen
async def listener(..):
   IOLoop.add_callback(lambda: do_some_expensive_but_not_relevant_to_server())
```
But there's fifty ways you might use it: there might be a queue involved. It's one of those things where we'll just have to see what people can build with it!

bollwyvl:
>  tests and the type checking support 

Yeah, since this is a part-part-time project, when I get the time to work on it, aggressive testing and static analysis is the only way i have a chance of remembering what the heck i was working on when i pick it up again. Of course, there's still a TON of implicit any, due to all those skips, so i have to keep that in my head, a bit. As another pyls (and then jlab-lsp) plugin, i'd love some kind of unobtrusive mark that showed it.

We're really not cashing in yet on the full value of the type annotations, as we could get nice docs out with something like [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints). To that end, I've been looking into [sphinx-js](https://pypi.org/project/sphinx-js) (which also supports ts), [sphinx-jsonschema](https://pypi.org/project/sphinx-jsonschema/), and [nbsphinx](https://github.com/spatialaudio/nbsphinx), for various reasons... perhaps we can do some piloting there.

EXTENDING.md
bollwyvl: updated!

bollwyvl:
actually working up a retrying thing for #103, so despite the most recent success...

EXTENDING.md
bollwyvl: added a touch more explanation, without doing a full example...

scripts/atest.py
```diff
@@ -15,32 +15,53 @@
...
+            print("Error deleting {}, hopefully harmless: {}".format(out_dir, err))
+
+    return robot.run_cli(list(map(str, args)))
+
+
+if __name__ == "__main__":
+    attempt = 0
+    rc = -1
+
+    retries = int(os.environ.get("ATEST_RETRIES") or "0")
```
bollwyvl: For #103

bollwyvl:
Well, it looks like all of them passed, so no real test of #103, but hey, they passed! Happy for this to be merged at this point...

krassowski:
Merged, thank you!

bollwyvl:
woo!

bollwyvl:
I'll try to monitor the `master` build, and clean up any issues i see...

https://dev.azure.com/krassowskimichal/jupyterlab-lsp/_build/results?buildId=153

bollwyvl:
well, the retry thing didn't work at all! i'll try again. always fun to track down something to fix something you don't know when will happen...

bollwyvl:
Well, it didn't flake this time, but the fix I applied has been known to work in the past. Green `master` is good anyway!

Issue #98, opened 2019-11-06
Generate typescript types from schema


bollwyvl:
We're quickly iterating on the spec and server schema. We need a reasonable way to build typescript types for the frontend as we learn more about integrating more servers and shifting concerns back and forth between the server and client.

Tool options
- https://www.npmjs.com/package/dtsgenerator (6k weekly downloads, 3 new deps)
  - i've used this before, but it's been ages
- https://www.npmjs.com/package/json-schema-to-typescript (42k weekly downloads, 18 new deps)
  - seems a bit better documented

I'll kick the tires on both of them and see what we get. There's also nothing wrong with making choices in the schema itself to get better types out.

References
- #81
- #90 
- #22 

I'll do a bit of research

bollwyvl:
`dtsgenerator` didn't like it, but js2ts rendered:
```ts
/* tslint:disable */
/**
 * This file was automatically generated by json-schema-to-typescript.
 * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
 * and run json-schema-to-typescript to regenerate this file.
 */

/**
 * describes the current state of (potentially) running language servers
 */
export type JupyterLspServerStatusResponse =
  | ServersResponse
  | LanguageServerSpecsMap;
/**
 * the count of currently-connected WebSocket handlers
 */
export type HandlerCount = number;
/**
 * the arguments to start the language server normally
 */
export type LaunchArguments = string[];
/**
 * the arguments to start the language server with more verbose output
 */
export type DebugArguments = string[];
/**
 * name shown in the UI
 */
export type DisplayName = string;
/**
 * This interface was referenced by `EnvironmentVariables`'s JSON-Schema definition
 * via the `patternProperty` "[^ ]+".
 */
export type AnEnvironmentVariableMayContainPythonStringTemplateEvaluatedAgainstTheExistingEnvironmentEG$HOME = string;
/**
 * the install commands or description for installing the language server
 *
 * This interface was referenced by `Installation`'s JSON-Schema definition
 * via the `patternProperty` ".+".
 *
 * This interface was referenced by `Installation1`'s JSON-Schema definition
 * via the `patternProperty` ".+".
 */
export type InstallHelp = string;
/**
 * known extensions that can contribute to the Language Server's features
 */
export type Extensions = LanguageServerExtension[];
/**
 * languages supported by this Language Server
 */
export type LanguageList = [string, ...(string)[]];
/**
[...]

PR #106, opened 2019-12-01
Statusbar with server status

krassowski:
Implement mockup from #78. Still, a lot to be done, but a lot of improvement compared to the previous statusbar.

## References

#78

## Code changes

Temporarily added hardcoded interfaces for server specs (until #98 is here).

## User-facing changes

Screenshot from 2019-12-01 16-43-12

None

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


krassowski:
Actually, we could move the textual description "Fully connected but..." to a tooltip, and reclaim some more space. We could place a link to the documentation at the very bottom instead.

PR #107, opened 2019-12-07
Add lsp-ws-connection to repo

bollwyvl:
## References

- #76 (step 3)
  - turns out we kind of need this before step 2

## Code changes

- [x] brings `lsp-ws-connection` into the repo
- [x] ts, tslint, yarn, etc.
- [x] also using `chai` instead of `expect` because **OH MY GOD SOME DAYS I CAN'T STAND JS STUFF AT ALL**
  - but seriously, it didn't seem to work properly with ts3.7
- [x] moves `initializeParams` into an overloadable `protected` function
  - not for any good reason, yet, but it was the first thing i wanted to do

## User-facing changes

None

None yet. In a follow on, I'll probably be taking a harder look at a) whether we need it at all (MessageConnection seems _just fine_) and b) how we might make it more configurable by some means other than subclassing and of course c) entirely removing its knowledge of a documentUri, because we only need one websocket per language server.

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry


.gitignore
```diff
@@ -112,3 +112,4 @@ node_modules
 atest/output/
 junit.xml
 coverage/
+.vscode/
```
bollwyvl: i _also_ got a new ubuntu box, and am trying out vscode: for example, i hadn't found the "automatically insert trailing newline" setting yet...

package.json
```diff
@@ -32,21 +31,34 @@
...
     "prettier": "prettier --write '**/*{.ts,.tsx,.js,.jsx,.css,.json,.md,.yml}'",
-    "test": "lerna run test",
     "tslint": "tslint --fix -c tslint.json --project tsconfigbase.json 'packages/**/*{.ts,.tsx}'",
     "tslint:check": "tslint -c tslint.json --project tsconfigbase.json 'packages/**/*{.ts,.tsx}'"
   },
   "workspaces": {
     "packages": [
       "packages/*"
+    ],
+    "nohoist": [
```
bollwyvl: might be able to clean these up, now that they are no longer fighting...

packages/lsp-ws-connection/README.md
```diff
@@ -0,0 +1,30 @@
+# lsp-ws-connection
```
bollwyvl: oops, this is now out of date.

packages/lsp-ws-connection/src/ws-connection.ts
```diff
@@ -0,0 +1,528 @@
...
+
+  public getDocumentUri() {
+    return this.documentInfo.documentUri;
+  }
+
+  /**
+   * Initialization parameters to be sent to the language server.
+   * Subclasses can overload this when adding more features.
+   */
+  protected initializeParams(): protocol.InitializeParams {
```
bollwyvl: the one thing i actually changed

packages/lsp-ws-connection/src/ws-connection.ts
```diff
@@ -0,0 +1,528 @@
...
+   */
+  public isImplementationSupported() {
+    return !!this.serverCapabilities?.implementationProvider;
+  }
+
+  /**
+   * Does the server support find all references?
+   */
+  public isReferencesSupported() {
+    return !!this.serverCapabilities?.referencesProvider;
```
bollwyvl: well, some of these, too...

packages/lsp-ws-connection/README.md
bollwyvl: updated

package.json
bollwyvl: still didn't look at these yet...

.gitignore
bollwyvl: updated.

bollwyvl:
Hopefully have fixed the version detection stuff which was failing on windows. Still need to figure out why i'm having issues with menu selection, but haven't gotten my windows vm back up and running yet.

packages/lsp-ws-connection/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2019, jupyter-lsp contributors
```
krassowski: I guess we should not stop at the fork mention, but also add "Copyright (c) 2018, William Conlon, ISC" in here. 

Legalities aside I would like to credit @wylieconlon as it would not be possible to start this project without the leading work of @wylieconlon (and obviously lots of code in this PR is @wylieconlon code with slight modifications). @wylieconlon would you like any specific backlinks or mentions? Are you ok with the current mentions?

bollwyvl: Sure, happy to give credit where due!

I think we may end up moving further and further from what remains of the code from upstream: for example, using [tsee](https://www.npmjs.com/package/tsee) would let us keep the full event typing information, and remove the conflicts we're seeing between `@types/node` and `@types/event`. Or, as mentioned, just use `MessageConnection` more directly.

Of course, because we don't control the end-user jupyterlab `webpack.config.js`, we'll need to keep webpacking the part that interacts with `fs`, so this package (and its special needs) will continue to exist.

bollwyvl:
Huzzah, looks like all the CI passed, provided it can actually upload the final report bundle!

krassowski:
>  c) entirely removing its knowledge of a documentUri, because we only need one websocket per language server.
before that we would want to have multi-document tests; I saw a regression with inspections being displayed in a wrong file...

Otherwise, ready to merge?

bollwyvl:
Yeah, if we're good on license stuff. Still trying to unwind all the stuff
I did on the heirarchical symbol branch... This is the first step, once we
have it in place for hacking, we can start the process of moving the
ownership of an lspwsconnection to a manager singleton. That might also be
the right time to give up trying to guess language names, and just go to
implementation-keyed sockets. Or something.

We will likely end up wanting a pattern not dissimilar to how
@jupyterlab/services hands out sessions for kernels and terminals. we'd
likely further abstract the connection.

At any rate, it'll be good to have the windows tests working!

krassowski:
Can we synchronize the `lsp-ws-connection` versions between `jupyterlab-lsp/package.json` and `metapackage/package.json`? or should we have it in the former in the first place?

I stuck trying to modify lsp-ws-connection` but the changes are not reflected since it is usinga an old version.

krassowski:
I really like the idea of the monorepo (it improves workflow greatly), but I got lost with the complexity of  the building/watch system and can't figure out how to get the changes in `lsp-ws-connection` to be reflected in mu builds... Maybe I need to `rm -rf` everything and install from scratch, but if there is any secret knowledge I am missing, please do help @bollwyvl.

bollwyvl:
Hm, just what I updated in contributing... Though clearing out your built
lab will be required the first time after the merge. But yes, bumping the
version would also be a good idea, too. I think that's one thing I didn't
bring over from the symbol tree branch.

wylieconlon:
I'm glad you found the project useful! I haven't really done a full review of your usage, but the license allows forking/vendoring like you've done. Were there any bugs you encountered or other issues that could make their way back into the main repo?

krassowski:
Thank you, @wylieconlon! No, not many issues significant enough to contribute back yet, but there is a thing that I am looking into right now, related to sending messages before initialization has been completed by the server (according to the specification client cannot send any messages before that; there is a check but it is on being connected, not being initialized). Again, this is a minor thing but maybe changing:

```typescript
    if (!this.isConnected) {
      return;
    }
```
to 
```typescript
    if (!this.isConnected || !this.isInitialized) {
      return;
    }
```
in a few places would bring the behaviour closer to the LSP specification. Also, having didClose would be nice (but I am not sure if this is relevant to the upstream project).

Also, I am experimenting on sending partial text updates and other extensions (rather than bugfixes). The ability to flexibly extend (and move away from node-based and callback-based constructs) was the primary reason to fork.

PR #108, opened 2019-12-10
lsp-ws-connection in repo (follow up)

bollwyvl:
## References

- #107 didn't quite clean up after itself on the metapackage side, resulting in some [confusing behavior](https://github.com/krassowski/jupyterlab-lsp/pull/107#issuecomment-563489362) and some downright _wrong_ dependency resolution. mea culpa.

## Code changes

- [x] Use relative `file:../<pkg>` links in metapackage dependencies.
- [x] Use extremely loose `*` dependency between `jupyterlab-lsp` and `lsp-ws-connection` _for now_
  - we'll have to tighten this up when we start making substantive changes, and certainly at release
  - it may well have not been getting used previously... luckily it still appears to work :stuck_out_tongue_closed_eyes: 
- [x] remove watch from `jupyterlab-lsp` (so there's only the one `tsc -w` running in metapackage, and the one `webpatch --watch` running in `lsp-ws-connection`)
  - this probably wasn't creating a problem... yet... but _was_ contributing even more to the watched/open file limit, which is probably very real threat now that we have `lsp-ws-connection` in there. I'm glad I'm not developing on windows.

I do recommend a `jupyter lab clean`, and full run of `postBuild` to ensure everything has been cleaned right after we add a new sub-package (hopefully not _too_ many more for a while). Also, due to the wackiness of the webpack build(s), there can be some thrashing after a change to `lsp-ws-connection` as built by `jlpm watch` and `jupyter lab --watch`... i usually end up making a few whitespace changes and then reverting them so it catches up (one way or the other). Not ideal, but there's kind of a lot going on down in there, and having to do a whole release to try out a new thing Is Not Fun.

## User-facing changes

None

None

## Chores

- [x] linted
- [x] tested
- [ ] documented (nothing yet)
- [ ] changelog entry (nothing yet)


bollwyvl:
Also, by fixing the dependencies, the `lsp-ws-connection` tests get run first by `jlpm test` (it does a topological sort for serial commands). Neat! That it wasn't before should have been a tip-off that something wasn't right. :woman_shrugging:

bollwyvl:
Test reporting wouldn't have caught anything in particular, but it's nice to have and luckily didn't pull in _too_ many additional dependencies.

packages/metapackage/package.json
```diff
@@ -26,8 +26,8 @@
     "watch": "tsc -b --watch"
   },
   "dependencies": {
-    "@krassowski/jupyterlab-lsp": "^0.6.1",
-    "lsp-ws-connection": "^0.2.0"
+    "@krassowski/jupyterlab-lsp": "file:../jupyterlab-lsp",
+    "lsp-ws-connection": "file:../lsp-ws-connection"
```
krassowski: Yeah, I was thinking about it but feared the outcome for the publishing process... But I guess it's better to start with than nothing. I will read up on how JupyterLab solves this (sure someone from the team may have some ideas on how to do it better as well).

yarn.lock
```diff
@@ -7207,12 +7221,10 @@ lru-cache@^5.1.1:
...
 
-lsp-ws-connection@~0.1.1:
-  version "0.1.1"
-  resolved "https://registry.yarnpkg.com/lsp-ws-connection/-/lsp-ws-connection-0.1.1.tgz#5e259a5079aa91f9bb55453d03df14f3e526fa6b"
-  integrity sha512-JLGsBbJc5NlNkvo8hyjVqOG2jZrCtXnz65d1pPPnp6mRl++J3pa0M0e2/jOGeDx+emDNk31/pUlLDOSHhphE4w==
+"lsp-ws-connection@file:packages/lsp-ws-connection":
+  version "0.2.0"
   dependencies:
-    vscode-jsonrpc "^4.0.0"
+    vscode-jsonrpc "^4.1.0-next"
```
krassowski: Note on `next` in here.

krassowski:
It's much better now, I see that all the files in `node_modules` in my real filesystem are ok, but the webpack is just reluctant to update to reflect the changes  - unless I manually restart `jupyter lab --watch` (obviously not ideal).

PR #109, opened 2019-12-10
add keys to list-like things to squash react console errors

bollwyvl:
## References
- #106 looks great, but react complains a lot about some list stuff  

## Code changes

- [x] adds `key={i}` inside of the loops, as well as some other list-like places to squash the warnings

## User-facing changes

None (well, they might get a bit less of a show when checking the console for errors, I suppose)

None

## Chores

- [x] linted
- [ ] tested (am starting to dig into how to use this to better inform the :robot: tests, but that shouldn't block this)
- [ ] documented (probably not needed)
- [ ] changelog entry (probably not needed)

### DIFF
--- a/packages/jupyterlab-lsp/src/adapters/jupyterlab/components/statusbar.tsx
+++ b/packages/jupyterlab-lsp/src/adapters/jupyterlab/components/statusbar.tsx
@@ -26,7 +26,9 @@ interface IServerStatusProps {
 }
 
 function ServerStatus(props: IServerStatusProps) {
-  let list = props.server.spec.languages.map(language => <li>{language}</li>);
+  let list = props.server.spec.languages.map((language, i) => (
+    <li key={i}>{language}</li>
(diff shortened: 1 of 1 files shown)


krassowski:
> am starting to dig into how to use this to better inform the robot tests, but that shouldn't block this

Caveat: I consider the current state of the statusbar/popover as work in progress (I don't like the weak/potentially non-intuitive distinction between "connected" and "initialized", and I would like to use more info from the `/lsp` endpoint; also the there is the issue of the icon). A further refactor is coming, this was just to give users and ourselves a working and friendly status panel.

But I agree, that this could be nice place to give to robot info on the state of the connections.

krassowski:
Thank you and sorry about not keying the react components myself.

bollwyvl:
No worries! Just trying to pin down quick things I _do_ understand... not
much of a react nerd, but had run into those errors before.

PR #110, opened 2019-12-10
generate initial typings from server response json schema

bollwyvl:
## References

- #98 pretty much like said there, but haven't added the "public" `schema.ts` which would pare down the results a bit

## Code changes

- [x] adds a first build stage for schema with `json-schema-to-typescript`
  - 8.0 just dropped, seemed like an interesting thing to check out
  - doesn't presently have a `watch` command
- [x] applies it to status bar
- [x] squashes a few `any`s

## User-facing changes

None

None

## Chores

- [x] linted
- [ ] tested (again, no tests of it yet, though still reading through the code)
- [x] documented (probably does need a blurb in CONTRIBUTING)
- [x] changelog entry (probably does need an entry)


packages/jupyterlab-lsp/src/adapters/jupyterlab/components/statusbar.tsx
```diff
@@ -7,6 +7,8 @@ import React from 'react';
 import { VDomRenderer, VDomModel } from '@jupyterlab/apputils';
 import '../../../../style/statusbar.css';
 
+import * as SCHEMA from '../../../_schema';
```
krassowski: Is there a convention to use uppercase `SCHEMA` for this import?

packages/jupyterlab-lsp/package.json
```diff
@@ -30,7 +30,8 @@
     "url": "https://github.com/krassowski/jupyterlab-lsp.git"
   },
   "scripts": {
-    "build": "tsc -b",
+    "build": "jlpm build:schema && tsc -b",
+    "build:schema": "json2ts ../../py_src/jupyter_lsp/schema/schema.json --unreachableDefinitions | prettier --stdin-filepath _schema.d.ts > src/_schema.d.ts",
```
krassowski: Wow, this was as simple as this!

bollwyvl:
Yeah, the state of schema to/from _x_ is getting pretty good these days. It
can also handle multiple files with $ref, which might end up being
important. There's a few more knobs, but they don't seems to provide too
much value (maybe to trigger more aggressive strict null checks on index
properties, which I some we don't have turned on).

No particular convention for UPPERCASE: it will just end up with lots of
names that sounds like other (non-generated) names, so importing individual
types would get hairy. Happy for whatever convention you prefer.

not checking it in will help us stay aware of when changes will have
impact. I'm willing to live with not having a watch command, for now. it
will tightly couple changes between front end and back unless we add
versioned schema and routes (blah) and that could get a little onerous, but
schema changes need to be really intentional, anyway.

I think we can certainly start using it now, either as-is, or through a
named file that either just re-exports, or import-and-extends the key types
needed for status bar.

bollwyvl:
If CI passes, we're good to go (got the docs in)

Issue #111, opened 2019-12-11
Adding lsp_message_listener("all") prevents server messages from being sent to the client


krassowski:
## Description

Using the example from the docs, all the intercepted messages are from the client. Frontend stops receiving server messages (including the very first response to initialise request) 

## Reproduce

In `load_jupyter_server_extension` of `jupyter_lsp/serverextension.py` add:

```python
    lsp_message_listener = LanguageServerManager.register_message_listener

    @lsp_message_listener("all")
    async def my_listener(scope, message, languages, manager):
        print("received a {} {} message about {}".format(
            scope, message["method"], languages
        ))
```

Observe no server messages being printed out and LSP features no longer working in the frontend. Remove the listener and everything is back to normal.

If we change:
 - `scope='client'` everything seems to work fine.
 - scope='server' leads to the same bug, but no prints (but I see that the client messages go to the server in the server logs)

Note to self: remember to manually restart the lab after changes, the watch does not update python changes.

## Context

- Operating System and version:
- Browser and version: unrelated
- JupyterLab version: 1.2.3
- `jupyter-lsp` version: master
- `@krassowski/jupyterlab-lsp` version: master

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

## Coordination

@bollwyvl I will try to write a test for this tonight (London time) but feel free to have your own look.

bollwyvl:
Oh no! Perhaps I missed an await somewhere...

bollwyvl:
Nah, it's that not every message has a `method`, because some are replies. Makes sense! opened up #114, probably ready to go if all the robots are appeased...

krassowski:
Closed by #114

PR #114, opened 2019-12-11
Capture listener exceptions, avoid expecting method

bollwyvl:
## References

- #111 

## Code changes

Wraps the user-provided listener call in a try/catch block with some logging (message content, traceback).

## User-facing changes

Failing listeners should not break delivery of messages to client.

None

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/py_src/jupyter_lsp/types.py
+++ b/py_src/jupyter_lsp/types.py
@@ -20,7 +20,7 @@ from typing import (
 )
 
 from notebook.transutils import _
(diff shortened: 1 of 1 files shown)


bollwyvl:
Realized this would let through a message that didn't have a `method` to a listener that was expecting a method. Should be fixed... but probably needs a test.

krassowski:
Thanks!

Issue #95, opened 2019-11-05
R: completion not available for tidyverse packages


rongmu:
`tidyverse` is a collection of some of the most popular packages in R which all play within the "tidy data" framework. `library(tidyverse)` reduces the effort to load various tidyverse packages by loading all the core packages at once: `ggplot2`, `tidyr`, `dplyr`, etc.  It should be noted that `tidyverse` uses a rather [sophisticated method][tidyverse-load-comment] to load the dependencies. 

**The problem**: Currently it seems that completion for tidyverse packages is not available on JupyterLab if they are loaded with `library(tidyverse)` rather than loaded individually. For example, no completion is suggested for ggplot2's plotting functions such as `geom_*`. 

However, developers of R's `languageserver` package have [already addressed][tidyverse-solved-comment] `tidyverse`'s problem, and I can confirm that completion for tidyverse packages works nicely on Vim with the LanguageClient-neovim plugin. So the current issue might be specific to jupyterlab-lsp.

randy3k, `languageserver`'s developer, [suggests][tidyverse-randy-suggestion] that the problem might have something to do with the following issues:
- REditorSupport/languageserver/issues/15
- REditorSupport/languageserver/issues/27

My environment:

- `jupyter-lsp`: v0.6.0b0
- `jupyterlab-lsp`: v0.6.1
- `languageserver` (R): v0.3.2

[tidyverse-load-comment]:https://github.com/REditorSupport/languageserver/issues/20#issuecomment-416348442
[tidyverse-solved-comment]: https://github.com/REditorSupport/languageserver/issues/20#issuecomment-549422900
[tidyverse-randy-suggestion]:https://github.com/REditorSupport/languageserver/issues/20#issuecomment-549657714

krassowski:
Thanks for checking or extension and opening this detailed issue. Does it
happen in notebook only or in the file editor as well?

If it's notebook only, is probable because they do not support some
functionalities for synced documents (a pity, many servers assume that they
have access to a file on the filesystem when they should not). Either way,
I have an interest in fixing it and will get back to you tonight (in ~10
hours time).

rongmu:
Thank you for the reply!
Unfortunately, completion for tidyverse packages didn't work in either notebook or file editor.
Support for `tidyverse` would be great because these packages are virtually indispensable for many R users.

krassowski:
[...]

PR #115, opened 2019-12-13
Add shadow filesystem, rename in notebooks

krassowski:
This PR adds the ability to mirror the virtual documents on the filesystem along with the primary benefit of doing so, i.e. renaming of variables in notebooks becomes possible in pyls (and in the future, we can have formatting, and other code modifying features).

There will be also benefits for R language server (#95), with `didSave` implementation to make R language server happy (another PR).

Pitfalls of the current approach:
- cross-module renaming may not always work as we put the files in a different place (the matter of the path). We could have a setting which would make the "shadow" files be written as hidden files in the same directory as the notebooks thay are from, but this would be optional. Alternatively, we could try to modify the python path variable for pyls.

TODO:
- [x] remove everything in '.virtual_documents' on start to purge unused files?
   - [ ] also, remove virtual files with `didClose` signal? (for future PR)
- [x] test cases for (jest or browser):
   - [ ] nested virtual documents (magics, `%%R`, `%%timeit,` etc)
   - [x] substitutied line magics 
- [x] adapt the `file_uri_to_path` tests from https://stackoverflow.com/a/57463161/6646912
- [x] rename and move `my_listener` (I am not sure how to hook it up so it is picked up - should I just import a module where I put it?)
- [x] apply changes to multi-cellular notebooks
- [x] R magics
   - [x] exclude R magics (code and tests) to separate files
   - [x] implement reverse transform for cell magics
   - consider if the approach based on regex is correct (it's nice for simple patterns, but a higher grammar - or just simple functions - might be better); revise API if needed (or draft it for future releases).
- [x] maybe display an error message when renaming fails in the statusbar?
- [ ] streamline the text displayed in the statusbar:
    - [ ] no plural for single rename,
    - [ ] responsive positioning (left if on a small screen, centre if widescreen)

Future PRs:
 - implement `didSave`
 - add `%Rpush` and `%Rpull` rpy2 magics

## References

 #99

## Code changes

Backend:
- adds `virtualDocumentsUri` to the `pageConfig`
- adds shadow filesystem which intercepts `didOpen`, `didChange` and `didSave` messages for files which are placed in the `virtualDocumentsUri`  (the choice on the placement is the responsibility of the frontend)
- minor adjustments to listener tests and others to allow having the shadow filesystem listener in the core.

Frontend:
- accepts `virtualDocumentsUri`
- declares the location of  all virtual documents for notebooks and extracted foreign documents as `virtualDocumentsUri`
- adds a file extension to names of all virtual files sent to `virtualDocumentsUri` (pyls will not rename unless filename ends with `.py`)
- magic replacements were rewritten to support  "reverse" regexp which is used to map the magic transformed to a valid source code back to the magic form
- implements rename for notebooks:
   - uses virtual documents (concatenated cells where magics are replaced with valid source code representation) as input for pyls rename
   - once response (edit specification) arrives:
      - uses virtual line => editor mapping to get the appropriate cell
      - slices and transforms the virtual source back to the cells of origin, transforming magic's code representations back to the magic (`%`, `%%`, or `!`) form.

## User-facing changes
rename in notebook first cell

- The magics interface was extended and now requires a "reverse" regexp which is used to map the magic transformed to a valid souce code back to the magic form.

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
Demo GIF:

rename in notebook any cell

Now testing, improvements, code refactoring, more testing, and docs.

krassowski:
The statusbar message looks ugly in such a small window, I need to improve the css.

krassowski:
And it does not work yet - we need to count the number of actual changes and add a guard so that the non-changed cells are neither counted as changes nor modified (so that we do not have to press "undo" repetitively with no actual changes).

krassowski:
Will need to modify the API for magic substitution to allow inverse transformation (from equivalent python code back to the form magic) to propagate the changes. But for now, we can simply block modifications affecting magics directly...

krassowski:
@bollwyvl this is kind-of ready. Major missing components are browser and jest tests for rename, but I am already thinking about merging this one to publish an alpha version, and postponing tests for a subsequent PR (i.e. next week, this weekend is over for me).

Two packages could be extracted:
 - rpy2 magics (definitely something which should be optional), 
 - the "shadow" filesystem module (I have mixed feelings, a part of me wants it to stay in our "core"),

but I gave up on this for now, given the scale of the changes elsewhere.

krassowski:
Final (?) demo:

rename in notebook any cell with R

The virtual document for this notebook (after the edits captured on the GIF) looks like this:

```python
test_single_cell = 1

test_single_cell


def a_function():
    pass


used_in_rpy2_cell_magic = 1


a_function()


used_in_rpy2_line_magic = 1

local_number = 1  # not really

def func():
    local_number = 2
    print(local_number)


rpy2.ipython.rmagic.RMagics.R("", "", used_in_rpy2_line_magic)


rpy2.ipython.rmagic.RMagics.R("""
test_single_cell <- c(1, 2)
print(test_single_cell)""", "", used_in_rpy2_cell_magic)


not_passed_in_rpy2_magic = 1


func
rpy2.ipython.rmagic.RMagics.R(" not_passed_in_rpy2_magic", "")


x = get_ipython().run_line_magic("ls", "")
```

krassowski:
Would be ready for review, but (no idea why) the CSS and JS tests fail on the azure (everything passes locally). Also, all rename tests fail on Windows, but it seems that Windows just requires more sleep for the async actions to work.

bollwyvl:
Yeah, everything's harder in ci. Will peruse when I get a chance! If you
can link to any stuff you couldn't get working still in/commented out/in
history I can take a look...

krassowski:
It would be great if we could have a preview of console logs from the robot browser tests. In this particular case, if I knew the output of a console log here:
https://github.com/krassowski/jupyterlab-lsp/pull/115/files#diff-30dc64680d2a59b62ad013d8a0942858L245 (removed in this PR, but lets say we can get it back for debug purposes) I could narrow down the issue (by following with a test case in jest against problematic server response; or maybe we would learn that the server response is the same as on my PC and the issue lies elsewhere).

bollwyvl:
Looks like we can, but might have to do some fiddling: we need to set
devtools.console.stdout.content = true in the firefox profile.

bollwyvl:
Hm, from a clean build, trying to rename `square` to `square_me` gives the following errors on the frontend:
```
index.js:8 Uncaught (in promise) Error: IndexError: list index out of range
    at index.js:8
    at index.js:8
    at index.js:8
    at run (setImmediate.js:40)
    at runIfPresent (setImmediate.js:69)
    at onGlobalMessage (setImmediate.js:109)
```

and this on the backend:

```
2020-01-03 14:11:11,591 UTC - ERROR - pyls_jsonrpc.endpoint - Failed to handle request 13
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls_jsonrpc/endpoint.py", line 113, in consume
    self._handle_request(message['id'], message['method'], message.get('params'))
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls_jsonrpc/endpoint.py", line 182, in _handle_request
    handler_result = handler(params)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls_jsonrpc/dispatchers.py", line 23, in handler
    return method(**(params or {}))
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls/python_ls.py", line 340, in m_text_document__rename
    return self.rename(textDocument['uri'], position, newName)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls/python_ls.py", line 283, in rename
    return self._hook('pyls_rename', doc_uri, position=position, new_name=new_name)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls/python_ls.py", line 155, in _hook
    return hook_handlers(config=self.config, workspace=workspace, document=doc, **kwargs)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/hooks.py", line 289, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/manager.py", line 87, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/_tracing.py", line 80, in __call__
    return outcome.get_result()
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/callers.py", line 52, in from_call
    result = func()
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/_tracing.py", line 78, in <lambda>
    outcome = _Result.from_call(lambda: self.oldcall(hook, hook_impls, kwargs))
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/manager.py", line 81, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/pyls/plugins/rope_rename.py", line 21, in pyls_rename
    document.offset_at_position(position)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/refactor/rename.py", line 25, in __init__
    evaluate.eval_location2(this_pymodule, offset)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/base/evaluate.py", line 19, in eval_location2
    return pyname_finder.get_primary_and_pyname_at(offset)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/base/evaluate.py", line 81, in get_primary_and_pyname_at
    holding_scope = self.module_scope.get_inner_scope_for_line(lineno)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/base/pyscopes.py", line 132, in get_inner_scope_for_line
    return self._scope_finder.get_holding_scope(self, lineno, indents)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/base/pyscopes.py", line 229, in get_holding_scope
    line_indents = self.get_indents(lineno)
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/base/pyscopes.py", line 222, in get_indents
    self.lines.get_line(lineno))
  File "/home/weg/projects/lsp/envs/default/lib/python3.7/site-packages/rope/base/codeanalyze.py", line 61, in get_line
    self.starts[lineno] - 1]
IndexError: list index out of range
```

krassowski:
Interesting. What python code did you use?

bollwyvl:
The python example notebook

krassowski:
Thank you! Indeed renaming in Python.ipynb will not work because there is a syntax error. After commenting it out it should work fine, see the gif below... Wait, what??!!!

What is going on?
So it seems that renaming a function named circle is ok, but if it is named square - it fails. The failure is on the level of rope not reading the shadow virtual file. I see the file and it is ok. No idea what is going on. Well, definitely more testing (and coffee) is needed.

krassowski:
The `square` issue seems to have been some kind of a fluke, it works ok now (but only if there are no syntax errors).

For the `IndexError` we need to file an issue on rope/pyls, ubt I am not sure what to expect from pyls here. Maybe a more user-friendly error message? Something like "RenameError: could not parse Python source (SyntaxError: XYZ)".
 
For now, I added error handling which displays the server error on the status bar (`Rename failed: Error: IndexError: list index out of range`) and a troubleshooting entry with an explanation to `README.md` (please feel free to improve/move around if you feel that it needs improvements)

krassowski:
@bollwyvl so, I added a message listener (instead of figuring out how to observe `console.log`) to figure out what is wrong with the CI browser tests, [here is the diff](https://github.com/krassowski/jupyterlab-lsp/commit/1a5c3f4fe569aab2537f8f63de6c45a0710cfc8a) but now it started passing. My favourite...

Does Azure Pipelines have anything which may have caused some weird caching behaviour?

krassowski:
I intercept correct messeges:

```
[D 13:57:40.347 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'id': 7, 'method': 'textDocument/rename', 'params': {'textDocument': {'uri': 'file:///home/vsts/work/1/s/atest/output/linux_37_1/home/example.js'}, 'position': {'line': 4, 'character': 1}, 'newName': 'new_name'}} | client
[D 13:57:40.387 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'id': 7, 'result': {'changes': {'file:///home/vsts/work/1/s/atest/output/linux_37_1/home/example.js': [{'range': {'start': {'line': 0, 'character': 9}, 'end': {'line': 0, 'character': 12}}, 'newText': 'new_name'}, {'range': {'start': {'line': 1, 'character': 15}, 'end': {'line': 1, 'character': 18}}, 'newText': 'new_name'}, {'range': {'start': {'line': 1, 'character': 24}, 'end': {'line': 1, 'character': 27}}, 'newText': 'new_name'}, {'range': {'start': {'line': 4, 'character': 0}, 'end': {'line': 4, 'character': 3}}, 'newText': 'new_name'}]}}} | server
[D 13:57:40.408 LabApp] [javascript] Handling a message
[D 13:57:40.408 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'method': 'textDocument/didChange', 'params': {'textDocument': {'uri': 'file:///home/vsts/work/1/s/atest/output/linux_37_1/home/example.js', 'version': 1}, 'contentChanges': [{'text': 'function new_name(n) {\n  return n<2?n:new_name(n-1)+new_name(n-2);\n}\n\nnew_name(5);\n\nwindow.location /;\n'}]}} | client
```

The server message is basically the same what we have in jest unit tests: https://github.com/krassowski/jupyterlab-lsp/pull/115/files#diff-f2bbd96d4c1db749aef785d505a3c4b0R80-R141 (but this is not unexpected, as this time the tests succeeded).

I will roll back the changes that I introduced, re-run the pipelines on the experimental side branch and see what happens...

krassowski:
The Windows failure is not due to any filesystem differences (virtual filesystem is not engaged for simple files, only for notebooks) but due to a delay in the execution on Windows (46 us, vs 21 us on Linux):

```
[D 13:58:01.333 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'id': 7, 'method': 'textDocument/hover', 'params': {'textDocument': {'uri': 'file:///d:/a/1/s/atest/output/windows_37_1/home/example.js'}, 'position': {'line': 6, 'character': 18}}} | client
[D 13:58:01.489 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'id': 6, 'result': {'changes': {'file:///d:/a/1/s/atest/output/windows_37_1/home/example.js': [{'range': {'start': {'line': 0, 'character': 9}, 'end': {'line': 0, 'character': 12}}, 'newText': 'new_name'}, {'range': {'start': {'line': 1, 'character': 15}, 'end': {'line': 1, 'character': 18}}, 'newText': 'new_name'}, {'range': {'start': {'line': 1, 'character': 24}, 'end': {'line': 1, 'character': 27}}, 'newText': 'new_name'}, {'range': {'start': {'line': 4, 'character': 0}, 'end': {'line': 4, 'character': 3}}, 'newText': 'new_name'}]}}} | server
[D 13:58:01.489 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'id': 7, 'result': {'contents': []}} | server
[D 13:58:01.536 LabApp] [javascript] Handling a message
[D 13:58:01.536 LabApp] [lsp] intercepted a message: {'jsonrpc': '2.0', 'method': 'textDocument/didChange', 'params': {'textDocument': {'uri': 'file:///d:/a/1/s/atest/output/windows_37_1/home/example.js', 'version': 1}, 'contentChanges': [{'text': 'function new_name(n) {\n  return n<2?n:new_name(n-1)+new_name(n-2);\n}\n\nnew_name(5);\n\nwindow.location /;\n'}]}} | client
[D 13:58:02.211 LabApp] [javascript] Handling a message
```

After wrapping the content comparison into `Wait Until Keyword Succeeds`, [everything passes](https://dev.azure.com/krassowskimichal/jupyterlab-lsp/_build/results?buildId=230&view=results).

krassowski:
And it still fails [here](https://github.com/krassowski/jupyterlab-lsp/runs/373740719), whereas using identical source it succeeded [here](https://github.com/krassowski/jupyterlab-lsp/runs/373719771). Two options:
- there is some weird caching enabled on Azure pipelines which affects Linux and Mac but not Windows
- there is something very different in running pipelines manually rather than from a PR

bollwyvl:
yeah, let's kill all the yarn cache stuff. will push directly to the branch.

ci/job.test.yml
```diff
@@ -39,11 +39,12 @@ jobs:
               - script: conda info && conda list -n jupyterlab-lsp
                 displayName: list conda packages and info
 
-              - task: CacheBeta@0
-                inputs:
-                  key: yarn | $(Agent.OS) | yarn.lock
-                  path: $(YARN_CACHE_FOLDER)
-                displayName: restore cached yarn packages
+              # TODO: determine how can bring this back to get more robust installs (see #115)
```
bollwyvl: perhaps this was caching the built tarballs of _our_ packages during yarn build?

bollwyvl:
> there is some weird caching enabled on Azure pipelines which affects Linux and Mac but not Windows

Indeed, very possible. You know my trust level of windows + node is very low. i have commented out the `CacheBeta@0` step.

> there is something very different in running pipelines manually rather than from a PR

Yeah, there may be variables that are only set during pr/master/tag builds...

bollwyvl:
Another thing we could do (not a huge fan of it, though) would be to somehow inject the commit of the build into the npm source in a way that can be gotten from the frontend. Since this is probably going to continue to be an issue, perhaps we have an (expensive) `versions` endpoint which returns all the version stuff:
- `nodejs --version`
- `*/node_modules`
- `python --version`
- `os.platform`
- `conda --version`
- `jupyter lab --version`
- `jupyter_lsp --version`
- `{% for server in servers %}{{server --version}}{% endfor %}
...and augment it with browser stuff, so we can get one screenshot/copy pasta that has everything.

The server version thing is probably a good idea anyway, though I'm not 100% certain all of them even implement such a thing.

krassowski:
> yeah, let's kill all the yarn cache stuff. will push directly to the branch.

Seems to be passing so far. Thank you!

> > there is something very different in running pipelines manually rather than from a PR
> 
> Yeah, there may be variables that are only set during pr/master/tag builds...

I tried a "clean" branch just before (`git checkout -b rename_for_notebooks_branched`) and it gave a [full success](https://github.com/krassowski/jupyterlab-lsp/pull/125/checks?check_run_id=373860533), so it was the caching issue.

bollwyvl:
The whole `feature X is only enabled in state Y` is a very interesting problem. I don't think `IndexError` would help a lot of users, unfortunately, even if we could promote that to the front end. I wonder, how we might allow a frontend feature to capture (and humanely describe) such highly contextual relationships between two features?

krassowski:
We could make an exception and check Diagnostics from Rename feature, pseudocode:

```python
if error and diagnostics.content.has("SyntaxError|invalid syntax").any():
     status = "Rename failed: You have a syntax error, which prevents the rename action"
```

technically this is easy, not sure if this is the right thing to do.

krassowski:
@bollwyvl if you do not have major comments, I would merge this and follow up with smaller PRs to address the outstanding issues which are:
- [ ] UX improvements:
  - [ ] a nicer error message for python when rename fails due to syntax error
  - [ ] better notifications on the status bar
  - [ ] warnings about invalid identifier when trying to rename
- [ ] tests:
  - [ ] testing what happens when the server rejects due to an invalid identifier
- [ ] editing code improvements:
  - [ ] allowing for edits which add or remove lines (those are tricky for notebooks) and testing those heavily
  - [ ] allowing for edits in embedded virtual documents (like R in Python, or HTML in Python) - this is again quite tricky but should be manageable with the same algorithm I got in place for notebooks edits.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - generates types for server data responses from JSON schema (
   [#110](https://github.com/krassowski/jupyterlab-lsp/pull/110)
   )
+- add 'rename' function for notebooks
```
bollwyvl: perhaps link to pr

packages/jupyterlab-lsp/src/connection.ts
```diff
@@ -88,7 +97,7 @@ export class LSPConnection extends LspWsConnection {
   private _sendChange(
     changeEvents: lsProtocol.TextDocumentContentChangeEvent[]
   ) {
-    if (!this.isConnected) {
+    if (!this.isConnected || !this.isInitialized) {
```
bollwyvl: perhaps `!(this.isConnected && this.isInitialized)`? If this ever appears anywhere else, perhaps collapse into a `get isReady`?

packages/jupyterlab-lsp/src/virtual/editor.spec.ts
```diff
@@ -47,18 +50,57 @@ describe('VirtualEditor', () => {
     language: 'R',
     pattern: '(^|\n)%R (.*)\n?',
     extract_to_foreign: '$2',
-    keep_in_host: true,
     is_standalone: false,
     file_extension: 'R'
   });
 
+  PageConfig.setOption('rootUri', '/home/username/project');
```
bollwyvl: should probably always include a `file://`?

packages/jupyterlab-lsp/src/virtual/editors/file_editor.ts
```diff
@@ -18,7 +19,7 @@ export class VirtualFileEditor extends VirtualEditor {
     // TODO: for now the magics and extractors are not used in FileEditor,
     //  although it would make sense to pass extractors (e.g. for CSS in HTML,
     //  or SQL in Python files) in the future.
-    super(language, file_extension, path, {}, {});
+    super(language, file_extension, path, {}, {}, true);
```
bollwyvl: this signature is getting completely opaque at this point: perhaps moving to an `VirtualEditor.IOptions` pattern would increase readability/discover?

packages/jupyterlab-lsp/src/virtual/editors/notebook.ts
```diff
@@ -38,9 +38,9 @@ class DocDispatcher implements CodeMirror.Doc {
       );
   }
 
-  getValue(seperator?: string): string {
-    return this.virtual_editor.getValue();
-  }
+  // getValue(seperator?: string): string {
```
bollwyvl: why just commenting?

packages/lsp-ws-connection/src/ws-connection.ts
```diff
@@ -198,13 +198,18 @@ export class LspWsConnection extends events.EventEmitter
...
       .sendRequest<protocol.InitializeResult>('initialize', message)
-      .then(this.onServerInitialized.bind(this), e => {
-        console.warn(e);
-      });
+      .then(
+        params => {
+          this.onServerInitialized(params);
+        },
+        e => {
+          console.warn(e);
```
bollwyvl: given how hard it is to debug this layer, perhaps a `console.warn('lsp-ws-connection inialization error', e, params)` would make it clearer?

packages/lsp-ws-connection/src/ws-connection.ts
```diff
@@ -198,13 +198,18 @@ export class LspWsConnection extends events.EventEmitter
...
+        }
+      );
   }
 
   public sendChange() {
-    if (!this.isConnected) {
+    if (!this.isConnected || !this.isInitialized) {
```
bollwyvl: ah, hah! here's that second use of these two together. definitely recommend creating an accessor that combines them.

py_src/jupyter_lsp/paths.py
```diff
@@ -15,3 +17,19 @@ def normalized_uri(root_dir):
         RE_PATH_ANCHOR, lambda m: "file://{}".format(m.group(1).lower()), root_uri
     )
     return root_uri
+
+
+def file_uri_to_path(file_uri):
+    """ Return a path string for give file:/// URI.
```
bollwyvl: haven't gotten that far yet, but these smell like something that should be tested and not `pragma: no cover`ed by either using `monkeypatch` for `os.name` or allowing an `is_posix` flag.

py_src/jupyter_lsp/serverextension.py
```diff
@@ -21,9 +22,16 @@ def load_jupyter_server_extension(nbapp):
 
     # try to set the rootUri from the contents manager path
     if hasattr(contents, "root_dir"):
-        page_config["rootUri"] = normalized_uri(contents.root_dir)
-        nbapp.log.debug("[lsp] rootUri will be %s", page_config["rootUri"])
+        root_uri = normalized_uri(contents.root_dir)
+        page_config["rootUri"] = root_uri
+        nbapp.log.debug("[lsp] rootUri will be %s", root_uri)
+
+        virtual_documents_uri = root_uri + '/.virtual_documents'
```
bollwyvl: probably use `url_path_join` from notebook utils rather than adding strings.

also, this really smells like it should be configurable, as in:
- completely disable-able, e.g. `c.LanguageServerManager.virtual_documents_path = None` or
- `c.LanguageServerManager.virtual_documents_path = Path.home() / '.my_virtual_documents`

I'd probably also think about prefixing `.jupyter_lsp_virtual_documents` to it: `virtual_documents` is really vague. Or, if we ancipate there being more craziness, `.jupyter_lsp/virtual_documents`

py_src/jupyter_lsp/serverextension.py
```diff
@@ -21,9 +22,16 @@ def load_jupyter_server_extension(nbapp):
...
+        page_config["virtualDocumentsUri"] = virtual_documents_uri
+        nbapp.log.debug("[lsp] virtualDocumentsUri will be %s", virtual_documents_uri)
     else:  # pragma: no cover
+        page_config["rootUri"] = ''
```
bollwyvl: :-1: I think _not_ having it set, or `None` is better than an empty string, as the errors would make more sense if it isn't set for some reason.

py_src/jupyter_lsp/serverextension.py
```diff
@@ -36,3 +44,5 @@ def load_jupyter_server_extension(nbapp):
             json.dumps(manager.language_servers, indent=2, sort_keys=True)
         )
     )
+
+    setup_shadow_filesystem(virtual_documents_uri=virtual_documents_uri)
```
bollwyvl: should probably live in manager (or as an extension to the manager)

py_src/jupyter_lsp/types.py
```diff
@@ -130,9 +138,8 @@ def register_message_listener(
     ):
         """ register a listener for language server protocol messages
         """
-
         def inner(listener: "HandlerListenerCallback") -> "HandlerListenerCallback":
-            cls.unregister_message_listener(listener)
+            cls.unregister_message_listener(listener)   # Why?
```
bollwyvl: We only want one copy of a particular listener.

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+def extract_or_none(obj, path):
+    for crumb in path:
+        try:
+            obj = obj[crumb]
+        except (KeyError, TypeError):
+            return None
+    return obj
+
+
+class EditableFile:
```
bollwyvl: at lest some code comments

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def __init__(self, path):
+        self.path = path
+        self.lines = self.read_lines()
+
+    def read_lines(self):
+        lines = ['']
+        try:
+            with open(str(self.path)) as f:  # Python 3.5 relict
```
bollwyvl: why not `path.read_text()`?

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class EditableFile:
+
+    def __init__(self, path):
+        self.path = path
+        self.lines = self.read_lines()
+
+    def read_lines(self):
+        lines = ['']
```
bollwyvl: why an empty string?

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+        try:
+            with open(str(self.path)) as f:  # Python 3.5 relict
+                lines = f.read().splitlines()
+        except FileNotFoundError:
+            pass
```
bollwyvl: perhaps a debug message

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+            lines[side] = trimmed
+        return needs_glue
+
+    @staticmethod
+    def join(left, right, glue: bool):
+        if not glue:
+            return []
+        return [(left[-1] if left else '') + (right[0] if right else '')]
+
+    def apply_change(self, text: str, start, end):
```
bollwyvl: go ahead and add the rest of the type signature, or get rid of all of it...

bollwyvl: if we need to bring in `TypedDict`, this is a fine time.

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+        inner = text.split('\n')
+
+        self.lines = (
+            before[:-1 if needs_glue_left else None]
+            + self.join(before, inner, needs_glue_left)
+            + inner[1 if needs_glue_left else None:-1 if needs_glue_right else None]
+            + self.join(inner, after, needs_glue_right)
+            + after[1 if needs_glue_right else None:]
+        ) or ['']
```
bollwyvl: again why empty string

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def write(self):
+        Path(self.path).parent.mkdir(parents=True, exist_ok=True)
+
+        with open(self.path, 'w') as f:
+            f.write('\n'.join(self.lines))
```
bollwyvl: `path.write_text()`

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+        end = {
+            'line': len(self.lines),
+            'character': len(self.lines[-1]) if self.lines else 0
+        }
+        return {'start': start, 'end': end}
+
+
+def setup_shadow_filesystem(virtual_documents_uri):
+
+    assert virtual_documents_uri.startswith('file:/')
```
bollwyvl: don't rely on runtime assertions: check and raise (custom subclass of) `ValueError`

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+    shadow_filesystem = Path(file_uri_to_path(virtual_documents_uri))
+    # create if does no exist (so that removal does not raise)
+    shadow_filesystem.mkdir(parents=True, exist_ok=True)
+    # remove with contents
+    rmtree(str(shadow_filesystem))
```
bollwyvl: `rmtree(..., ignore_errors=True)` rather than creating dir?

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+    # create again
+    shadow_filesystem.mkdir(parents=True, exist_ok=True)
+
+    @lsp_message_listener("client")
+    async def shadow_virtual_documents(scope, message, languages, manager):
```
bollwyvl: this is very buried...

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+        """Intercept a message with document contents creating a shadow file for it.
+
+        Only create the shadow file if the URI matches the virtual documents URI.
+        Returns the path on filesystem where the content was stored.
+        """
+
+        write_on = [
```
bollwyvl: probably promote to constant, documented

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+            'textDocument/didOpen',
+            'textDocument/didChange',
+            'textDocument/didSave'
+        ]
+
+        if not ('method' in message and message['method'] in write_on):
```
bollwyvl: `if message.get('method') not in WRITE_ON`

bollwyvl: or, `@lsp_message_listener("client", method="textDocument/did(Open|Change|Save)")` but sometimes explicit is better than implcit...

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+            return
+
+        document = extract_or_none(message, ['params', 'textDocument'])
+        if document is None:
+            raise ValueError('Could not get textDocument from: {}'.format(message))
```
bollwyvl: Probably subclass these errors...

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+        uri = extract_or_none(document, ['uri'])
+        if not uri:
+            raise ValueError('Could not get URI from: {}'.format(message))
+
+        if not uri.startswith(virtual_documents_uri):
+            return
+
+        path = file_uri_to_path(uri)
+        file = EditableFile(path)
```
bollwyvl: `file` is a reserved word

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+
+        text = extract_or_none(document, ['text'])
+
+        if text is not None:
+            changes = [{'text': text}]
+        else:
+            changes = message['params']['contentChanges']
+
+        if len(changes) > 1:
+            print(      # pragma: no cover
```
bollwyvl: use `manager.log.warn`

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -0,0 +1,145 @@
...
+                'LSP warning: up to one change'
+                ' supported for textDocument/didChange'
+            )
+
+        for change in changes[:1]:
```
bollwyvl: why not the last one?

tsconfigbase.json
```diff
@@ -14,6 +14,7 @@
     "preserveWatchOutput": true,
     "resolveJsonModule": true,
     "sourceMap": true,
+    //"strictNullChecks": true,
```
bollwyvl: remove (or better, uncomment). I know it's JSON+comments, but bleah

bollwyvl:
Whole slew of comments: a number of them are probably better to handle before merge, some could wait until later.

The virtual file thing should probably be either a) be merged into the manager proper or b) be its own separate, subordinate of the manager, e.g. `VirtualDocumentManager`. If _anything_ else is going to want a piece of this, it will probably need some events or something that people can extend/listen to.

We may wish to consider moving some of the heavy file i/o to a thread, as this is precisely the kind of thing where we want to block the message of interest, but _not_ the whole event loop, and lots of `read_text`, `write_text`, `mkdir` and `rmtree` could start to add up to an unresponsive frontend.

packages/jupyterlab-lsp/src/virtual/editors/notebook.ts
krassowski: Removed in a #127

packages/lsp-ws-connection/src/ws-connection.ts
krassowski: Right, these guys need to be together (almost) everywhere, as the LSP spec does not allow the client to send anything before initialization is completed. This is a remnant from a crazy debugging phase (turned out to be unrelated) but I kept it here as it improves the compliance with specs. Also, the sendChange was the only one which was eagerly sent early others are not in danger of being sent before init (but this is just a happy coincidence, we should not relay on this as this may change).

Will address in a subsequent PR with a helper accessor or something.

py_src/jupyter_lsp/serverextension.py
krassowski: Well, actually the extension would still work.. to an extent. I will think about it, there must have been a reason I set it (but probably nothing that cannot be turend around).

tsconfigbase.json
krassowski: well served it's purpose. we probably want to adopt the strictNullChecks at some point but this would be too much for this PR.

packages/jupyterlab-lsp/src/virtual/editor.spec.ts
krassowski: Yeah, I think so. Will leave it as is for now as I heavily this fragment in a subsequent PRs.

packages/jupyterlab-lsp/src/virtual/editors/file_editor.ts
krassowski: Agreed, a good idea, will address later.

packages/lsp-ws-connection/src/ws-connection.ts
krassowski: good idea, making it more explicit. unfortunately no `params` here.

py_src/jupyter_lsp/serverextension.py
krassowski: Initially tried using Path but as these are URIs (not paths) there were some issues on Windows if I remember correctly. I should have commented. `url_path_join` could work here.

Agree on configurability. I would avoid very long names such as `.jupyter_lsp_virtual_documents`, this can annoy ppl (more than having a hidden file in the first place). `.jupyter_lsp/virtual_documents` sounds good, but I will refreain from attempts on refactoring (burnout on this piece for this weekend)

py_src/jupyter_lsp/serverextension.py
krassowski: I failed so hard trying to move it out :( Feel free move it around.

py_src/jupyter_lsp/paths.py
krassowski: It's all tested [here](https://github.com/krassowski/jupyterlab-lsp/pull/115/files/6390dc6a25d2e503931cb1226ccb1f536d9d3777#diff-0fb2b1f5fdab934c16136c38e7c08a79R48-R70). As you say, monkeypatch or a flag would be alternative approaches.

py_src/jupyter_lsp/virtual_documents_shadow.py
krassowski: Surprised that its supported in 3.5

py_src/jupyter_lsp/virtual_documents_shadow.py
krassowski: needed by the apply change/glue algorithm (could work around but wold need a few more checks for the special case of no lines)

py_src/jupyter_lsp/virtual_documents_shadow.py
krassowski: No, it's not. It was builtin of Python 2, but was dropped in Python 3. Formally, it was [not even a reserved word in Python 2.7](https://docs.python.org/2/reference/lexical_analysis.html#keywords).

krassowski: We are just stuck with syntax highlighters written for Python 2 in 2020...

py_src/jupyter_lsp/virtual_documents_shadow.py
krassowski: I choose the former (still not exactly the same, but indeed nicer)

py_src/jupyter_lsp/virtual_documents_shadow.py
krassowski: Right (I do not expect it to be a problem now, but once it gets configurable it might slip)

krassowski:
Thank you @bollwyvl for the review, I addressed as much as reasonable with the commits above. Merging this now as this PR gets a bit too large to load in my browser, but the work will continue ;)

PR #116, opened 2019-12-14
Add a command to run specific frontend tests only

krassowski:
If there is a better way, please let me know!


krassowski:
A problem with this approach is that the karma kind of ignores the `-t` argument (but does not complain). It works fine with jest though. Would it be worth rewriting all tests to jest?

Issue #42, opened 2019-09-25
Add an option to display diagnostics results summary in a panel


krassowski:
Should display the inspection results: the full diagnostic text, position and should link to a relevant line in code.

bollwyvl:
The inspector _ne_ Contextual Help would be a an interesting initial place to send this.

PR #127, opened 2020-01-04
Implement a UX workaround for rope/pyls rename issue

krassowski:
Implement a UX workaround for rope/pyls rename issue for the case of syntax errors in the source code.

In #115 an issue with rename for Python (when using pyls) was identified:
rename was failing with an obscure message when the source code could not be parsed correctly by rope (due to a user's syntax error).

This workaround detects such a condition using diagnostics feature provides a nice error message to the user (UX first).

## References

#115

## Code changes

`ICommandContext` now includes `VirtualEditor`. `Diagnostics` feature got a new, public database of diagnostics (`diagnostics_db: IEditorDiagnostic[]`) which will be also used to implement the diagnostics Panel widget (#42).

## User-facing changes

Screenshot from 2020-01-04 22-02-53

Note: the error is in the 4th cell (1-based indexing) because above the code cells there are two markdown cells; we could change the code to display the ordinal number for code cells only (but not sure how it would work in future if we decide to analyze markdown cells too).

None

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics.ts
+++ b/packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics.ts
@@ -1,13 +1,26 @@
 import * as CodeMirror from 'codemirror';
 import * as lsProtocol from 'vscode-languageserver-protocol';
 import { PositionConverter } from '../../../converter';
-import { IVirtualPosition } from '../../../positioning';
+import { IEditorPosition, IVirtualPosition } from '../../../positioning';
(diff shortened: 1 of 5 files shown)


krassowski:
Note: even though I filled an issue in pyls (and will probably fill another in rope), even if we contribute a fix upstream, we would like to have a wrapper on this case of our own making - the concept of cell does not exist in rope/pyls so we would not be able to direct user to a specific cell (as we do here).

PR #132, opened 2020-01-06
Release jupyterlab-lsp 0.7.0-beta0, jupyter-lsp 0.7.0b, lsp-ws-connection 0.3.0

krassowski:
Zeroth beta release so that I (and others) can test it while working on real projects (also, because it fixes some important bugs and I need it working ;). I will write note down any potential issues with the release process here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,23 +2,31 @@
 
 ## `@krassowski/jupyterlab-lsp 0.7.0-beta.0`
 
-- reduced space taken up by the statusbar indicator
-- implemented statusbar popover with connections statuses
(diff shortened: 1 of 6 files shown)


krassowski:
To upgrade:

```
pip install --pre jupyter-lsp -U
jupyter labextension install @krassowski/jupyterlab-lsp@0.7.0-beta.0
```

krassowski:
Seems to be working ok, I managed to get it in my day-to-day environment. The bugfix-only phase for 0.7.0 begins (there are a lot of open issues which I will try to tackle before RC, I think we will have at least two betas more)

Issue #133, opened 2020-01-06
Diagnostics panel does not show up in 0.7.0 beta


krassowski:
It seems that minimization of the source causes the class (feature) names to be lost so we cannot access features by the original class name. Need to add the Feature.name: string field to the abstract class.

This is manifested by:
```
TypeError: Cannot read property 'switchDiagnosticsPanelSource' of undefined
```

bollwyvl:
Eek. Guess we need to ensure we're testing against prod assets in CI

https://github.com/krassowski/jupyterlab-lsp/blob/master/ci/job.test.yml

As well as in the demo:

https://github.com/krassowski/jupyterlab-lsp/blob/master/postBuild

Will PR.

krassowski:
Though about it, fortunately errors like this are not common (I knew that there was a risk of it when writing the code). However if it will not cause prolong the CI time, I all for it.

bollwyvl:
It definitely _will_ take more time. We _could_ just do this on _some_ of the runs (e.g. linux), but the assurance of the installability of the extension(s) without error, especially on peril-fraught windows, is worth its weight in future, under-described support issues.

PR #134, opened 2020-01-06
test against production assets (#133)

bollwyvl:
## Demo

[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/test-minified-assets?urlpath=lab)

## References

- informed by #133

## Code changes

This uses `--dev-build=False --minimize=True` for acceptance test activities. These flags were not available on some older versions of lab, so an `npm` flake might cause issues for the older excursions. It _will_ take more time. However, this is how users will actually have to deal with the extension(s) as installed, so _not_ testing it is somewhat reckless.

## User-facing changes

None

None

## Chores

- [x] linted
- [ ] tested
- [x] documented
- [ ] changelog entry


bollwyvl:
Linted, and in the process also clobbered some new warnings from `pytest` and `pytest_azurepipelines`).

bollwyvl:
Looks like it takes about 30 seconds longer to do the build on windows. Also didn't actually push my lint, which spiraled into more linting. And squashing the warnings created new warnings. Anyhow, if it comes up clean (with or without warnings, I guess) this is good to go, as I have to get going.

bollwyvl:
Given the number of things that conflicted between `black`, `flake8`, etc. I've gone ahead and formalized the linting i have been doing when i was doing the initial work on the python side.

atest/01_Editor.robot
```diff
@@ -21,17 +21,15 @@ Bash
 
 CSS
     ${def} =    Set Variable    xpath:(//span[contains(@class, 'cm-variable-2')][contains(text(), '--some-var')])[last()]
-    Editor Shows Features for Language    CSS    example.css    Diagnostics=Do not use empty rulesets    Jump to Definition=${def}
-    ...    Rename=${def}
+    Editor Shows Features for Language    CSS    example.css    Diagnostics=Do not use empty rulesets    Jump to Definition=${def}    Rename=${def}
```
krassowski: IMO The best thing would be to have the keyword arguments in separate lines, as right now it is difficult to read and see which ones are there. Like:
```
Editor Shows Features for Language    CSS    example.css
...    Diagnostics=Do not use empty rulesets
...    Jump to Definition=${def}
...    Rename=${def}
```
I do not suggest that you take any action (I guess this ain't as intelligent linter as black), just complaining about life.

bollwyvl:
Ooh, yeah, that's gross.

Maybe we move the whole shooting match to a python file. There comes a time
when robot is too cute.

I think there was a pr upstream for that... But my associate and on the
robots-from-jupyter stuff have considered doing a language server based on
pylgs, and would definitely like to have formatting that we'd actually want
to use, in addition to rflint and the dry run tools.

krassowski:
Thank you, looks great to me!

Obviously, my predilection for single quotes will remain unsatisfied ;)

environment-lint.yml
```diff
@@ -0,0 +1,11 @@
...
+
+channels:
+  - conda-forge
+  - defaults
+
+dependencies:
+  - black
+  - isort
+  - mypy
+  - robotframework-lint
```
krassowski: Also needed in `requirements-dev.txt`, right?

krassowski:
I wonder if I broke master, or if it's just something with the tests for merges:

Screenshot from 2020-01-07 16-17-33

:confused:

PR #139, opened 2020-01-07
correct mac path

yunair:
## Code changes

In `index.ts`, only checks `/home`,   For mac os, user_settings with starts with `/Users`,
So, Mac User will not get the correct path.


krassowski:
Thank you, this looks great.

PR #140, opened 2020-01-07
fix ws signature failed

yunair:
## Code changes

change `ws-connection.ts`, `getSignatureHelp` function.
In ipynb with python3, `location.ch`, will get wrong character.
So, use `location.ch-1` to fix the character right

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 ## `lsp-ws-connection 0.3.1`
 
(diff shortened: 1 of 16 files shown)


bollwyvl:
Great find! How might we test this to make sure 1) it's fixed and 2) it doesn't happen again?

yunair:
add a test,  mock user click, then trigger signature help function, and assert click character is same as typedCharacter

krassowski:
Build passed, merging. Thanks @yunair!

Issue #30, opened 2019-09-10
Code completion sometimes overwrites existing text 


krassowski:
Happens only when completing a code symbol which is in the middle of a line (not at the end). Needs unit tests to prevent this in the future.

Example:
`DataFrame(da<tab>)` leads to `DataFrame(dta=)`, should `DataFrame(data=)`

R examples:
`data.frame(row.nam<tab>` → `data.frame(row.nrow.names`
`data.f<tab>` → `data.fdata.frame`

krassowski:
Minimal Python example without the need to import anything:

```
CamelCase = 1
mean(cam<tab>
```

krassowski:
Another test case: `from jupyter_helpers.tabl<tab>`

krassowski:
```python
def display_table():
    pass

disp<tab>data
```
leads to
`ddisplay_table`

yunair:
I don't think this is a valid test case
``` python
def display_table():
    pass

disp<tab>data
```

Python doesn't support to  write code like this

krassowski:
Some explanations: I use `<tab>` to indicate invocation of the completer with <kbd>tab</kbd> key and `<cursor>` to indicate the expected position of cursor (thus the code is valid).

krassowski:
The `from jupyter_helpers.tabl<tab>` can be reproduced in bare Python with:

`from os.pat<tab>` → choose `path` or `os.path`; the result expected for both is `from os.path`, but the latter produces `os.os.path`. This is because the kernel returns prefixed tokens (with `os.`), while LSP (pyls) does not (and we can handle either prefixed responses or non-prefixed responses, but not both mixed together, due to how the completer in JupyterLab is written).

- we cannot detect the prefix from kernel response alone
- we have to take a look at the token preceding the requested token.

krassowski:
Even more nicer test case `from os.pathse<tab>` → `from os.pathsep`.

I now got fixes for every case reported so far, will PR soon.

PR #143, opened 2020-01-08
completion case sensitive will  keep a no need token value

yunair:
## Code changes

There has two changes.

All in `completion.ts`.

1.  change `merge_replies` function.
when `lsp.start > kernel.start`, correct the right prefix. 
`kernel.start` is the wrong index, because this index is merge all content in cell, and calculate the index, not the current line index.

2.  change `hint` function.

if check text startsWith token case sensitive,
It will keep a no need token value.

Input `from time import T`
before after completion, there will be `from time import Ttime`
now, there will be `from time import time`

## Chores

- [x] linted
- [x] tested
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@
   - completion now work properly when the kernel is shut down (
     [#146](https://github.com/krassowski/jupyterlab-lsp/pull/146)
     )
+  - a lowercase completion option selected from an uppercase token
+    will now correctly substitute the incomplete token (
+    [#143](https://github.com/krassowski/jupyterlab-lsp/pull/143)
+    )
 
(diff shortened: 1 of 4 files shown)


krassowski:
Thank you, your contributions are very welcome. I see that these changes might be a step in the right direction, but we will need tests for this before I merge it.

I will be preparing basic browser tests for features of interests soon and let you know here.

krassowski:
Many thanks again for the fix @yunair. Please have a look at the tests that I added here and feel encouraged to include such tests yourself for the future PRs!

krassowski:
This improves #30 (but does not solve it, especially not for R where tokens are more tricky with all the dots everywhere).

yunair:
Ok, I'll add such tests for the future PRs in this situation

PR #145, opened 2020-01-08
Clean diagnostics upon rename, more panel tests

### LINKED ISSUE #141: Diagnostics are not cleared upon file rename
## Description

This can be seen clearly in DiagnosticsPanel of a notebook after rename. Needs a browser test. 

## Reproduce

1. Open notebook and the DiagnosticsPanel
2. Rename notebook
3. See the diagnostics displayed twice, once for old filename and once for the new filename

## Context

- JupyterLab version: 1.2.4
- `@krassowski/jupyterlab-lsp` version: 0.7.0-beta.1

krassowski:
## References

Fixes #141. Also fixes a case when the diagnostic code separation out of the message text was too eager for undefined codes (dropping the first word of the message if it happened to be "undefined").

## Code changes

Created a new file for more detailed interface tests `04_Interface.robot`, though right now I think about making a directory `04_Interface` with `DiagnosticsPanel.robot` and `Statusbar.robot` files (thus allowing to split the test cases into individual... test cases.

## User-facing changes

None

None

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry


krassowski:
@bollwyvl I would appreciate your feedback on the wider picture here (organization of test suites).

I know that those (robot) were intended as simple acceptance tests, but (for example) the completion and signature features rely heavily on the integration with Jupyter (the former requires a kernel), so ideally I would have also `Features` directory with `Completion.robot`, `Signature.robot` etc. These would NOT be tested with ALL the language servers, just with one or two depending on the need, but would test particular corner case which are known/likely to produce bugs. What do you think?

bollwyvl:
Whatever structure makes the most sense! Indeed the purpose is to keep us
honest when shipping new features, and as the features get more
complicated, so will the tests. But I've really found no other way to test
this stuff, whatever shortcomings the approach may have.

We can/should break up Keywords.robot anyway... And or consider bringing in
robotframework-jupyterlibrary, which needs some love anyway.

As a suite gets more involved, the biggest thing is maintaining
per-test-file isolation. More features means more flake, and if we can keep
the tests isolated, we can start only rerunning failing tests, or running
multiple tests in parallel.

I think having a keyword file per _feature_, and test suite/file per
_kernel_ makes the most sense, such that each test is pretty
straightforward, just reusing the pieces: "trigger completions at token",
"completions should contain", that kind of thing. We can still add tags
person feature, so that the roll-up report can show "all the completions
broke".

Again, codemirror is kind of the bottleneck, as it does its own things vs
standard Dom events. So we may need to bring in some stuff in to handle
it... It's in rfjl, but has some warts.

If the behavior gets really gross, we can always drop to python, but it's
an exercise in clarity to be forced into the more limited typography. All
depends!

Of course happy to help here, since I've set us down this path.

krassowski:
@bollwyvl could you take a look at this combine step failure?

bollwyvl:
Sure... It'll be this evening/weekend...

krassowski:
Thank you! In the meantime, I will merge it (as everything passes ok, it must be a result of a caching issue). I retained the build but not sure If you will be able to see it from the link: https://dev.azure.com/krassowskimichal/jupyterlab-lsp/_build/results?buildId=287&view=results.

PR #146, opened 2020-01-11
Add initial tests for completion feature

krassowski:
Fixes a previously unreported bug which prevented completions from working when kernel was shut down, and removes a spurious console error log. 

## References

A pre-requisite for #143.

## Future work

I tried to test auto-invocation of completer (using a trigger character of dot) but while it works perfectly in my browsers, the selenium seems to have a problem with the amount / frequency of async and semaphore operations I used to synchronize CodeMirror and VirtualDocument changes.

My failed attempt was:

```robot
Triggers Completer On Dot
    Setup Notebook    Python    Completion.ipynb
    Enter Cell Editor    2    line=1
    Press Keys    None    .
    Wait Until Keyword Succeeds  10x   0.5s   Cell Editor Should Equal  2   list.
    # this takes longer as it has to be processed by CodeMirror
    Sleep   2s
    # Problems here, too much async? Elsewhere 6 sec is more than enough..
    Wait Until Page Contains Element    ${COMPLETER_BOX}    timeout=10s   
    Completer Should Suggest    append
    ${content} =    Get Cell Editor Content    2
    Should Be Equal    ${content}    list.append
    [Teardown]    Clean Up After Working With File    Completion.ipynb

Does Not Trigger When Dot Was Already There
    Setup Notebook    Python    Completion.ipynb
    Enter Cell Editor    3    line=1
    Press Keys    None    cle
    Sleep    6s
    Page Should Not Contain    ${COMPLETER_BOX}
    [Teardown]    Clean Up After Working With File    Completion.ipynb
```

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@
     [#129](https://github.com/krassowski/jupyterlab-lsp/pull/129)
     )
   - added a widget panel with diagnostics (inspections), allowing to
-    sort and explore diagnostics, and to go-to the respective location
-    in code (on click); accessible from the context menu (
+    sort and explore diagnostics, and to go to the respective location
+    in code (with a click); accessible from the context menu (
     [#127](https://github.com/krassowski/jupyterlab-lsp/pull/127)
     )
   - all commands are now accessible from the command palette (
@@ -36,6 +36,9 @@
     )
   - fixed LSP of R in Python (`%%R` magic cell from rpy2) (
     [#144](https://github.com/krassowski/jupyterlab-lsp/pull/144)
+    )
+  - completion now work properly when the kernel is shut down (
+    [#146](https://github.com/krassowski/jupyterlab-lsp/pull/146)
     )
 
 ## `lsp-ws-connection 0.3.0`
--- a/atest/05_Features/Completion.robot
+++ b/atest/05_Features/Completion.robot
@@ -27,7 +27,7 @@ Works With Kernel Running
     Should Contain    ${content}    TabError
     [Teardown]    Clean Up After Working With File    Completion.ipynb
 
-Works With Kernel Shut Down
+Works When Kernel Is Shut Down
     Setup Notebook    Python    Completion.ipynb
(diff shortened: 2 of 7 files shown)


krassowski:
Hmm... This is another build where Windows 3.6 is flaky (random things fail due to timeout), while Win 3.7 passes smoothly. Do we really need 3.6 on all the OSes?

krassowski:
The problem with the attempt above was that the script was not waiting for the LSP initialization to finish. This is now solved with the addition of new [`Wait Until Fully Initialized` keyword](https://github.com/krassowski/jupyterlab-lsp/pull/140/commits/02a397d897c8f48fd261991694f1151846d04c73#diff-356561514a74d123a52b9389bba88852R198), the test case was added [here](https://github.com/krassowski/jupyterlab-lsp/pull/140/commits/02a397d897c8f48fd261991694f1151846d04c73#diff-3da8b903581e6c78ee8467dfcda9a4b5R66) and [now passes](https://github.com/krassowski/jupyterlab-lsp/pull/140/commits/d76c8e577b853203c1cd279bfbc0217b0b95ae0f) after pinning the lsp-ws-connection version (Windows is still flaky, but it's unrelated).

bollwyvl:
> Do we really need 3.6 on all the OSes?

We can mark whatever we want [non-critical](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#setting-criticality), and they'll show up flaky but not fail the build. But yeah, I kinda feel like we should keep as broad of a testing suite as we can, because people _use_ all those version/osx pairs. Also, it keeps us honest: I'd love to start using f-strings, but since we're _not_ exactly in the scientific computing sphere, we can't drop 3.5 yet, implicitly.

> new Wait Until Fully Initialized keyword

Ah, that's great that we have that keyword now... I think that was (my) original desire for the status stuff!

bollwyvl:
> we can't drop 3.5 yet, implicitly.

But we should probably _add_ 3.8... sigh.

PR #150, opened 2020-01-12
More Robot Test Robustness

bollwyvl:
## References

- https://github.com/krassowski/jupyterlab-lsp/pull/146#issuecomment-573379944
  - adds 3.8 excursions
- https://github.com/krassowski/jupyterlab-lsp/pull/145#issuecomment-572315068
  - only retries failed tests

## Code changes

- [x] Moves the lab version detection up to the suite startup, which _should_ leave all the tests independent of each other. 
- [x] turns on test/suite order randomization, which should help shake out interdependencies between tests.
- [ ] marks `WHAT VERSIONS DO WE WANT` to be `--noncritical`, so that we don't get hard fails from stuff we know is flaky, but don't know how to unflake... yet

## User-facing changes

None

None

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/scripts/atest.py
+++ b/scripts/atest.py
@@ -16,11 +16,33 @@ OUT = ATEST / "output"
 OS = platform.system()
 PY = "".join(map(str, sys.version_info[:2]))
 
+OS_PY_ARGS = {
+    # notebook and ipykernel releases do not yet support python 3.8 on windows
+    ("Windows", "38"): ["--exclude", "*"]
+}
+
+
+def get_stem(attempt, extra_args):
+    stem = "_".join([OS, PY, str(attempt)]).replace(".", "_").lower()
+
+    if "--dryrun" in extra_args:
+        stem = f"dry_run_{stem}"
+
+    return stem
+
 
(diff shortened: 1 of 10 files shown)


bollwyvl:
Hooray green checkmark!

So this is now enforcing the isolation of tests: if we have to turn it down, we can go to just randomizing and rerunning per file, but we're in a pretty good spot now.

scripts/atest.py
```diff
@@ -16,11 +17,33 @@
...
 
 def atest(attempt, extra_args):
     """ perform a single attempt of the acceptance tests
     """
-    stem = "_".join([OS, PY, str(attempt)]).replace(".", "_").lower()
+    extra_args += OS_PY_ARGS.get((OS, PY), [])
+
+    stem = get_stem(attempt, extra_args)
+
+    if attempt != 1:
```
krassowski: Great idea, much needed!

krassowski:
Thank you!

PR #152, opened 2020-01-12
Add manual packages installation instructions

krassowski:
Yesterday I ran into a problem with packages and had to `rm -rf node_modules` (all of them). The frontend installation instruction did not work for me as it did not install deep dev dependencies.

I am a newbie to metapackages, lerna and other such things so I would just start by adding explicit instructions for manual install (but I guess that this might be solved by an improvement to the global-level `package.json`; there is also a small chance I am getting something wrong).

## Other changes

- Removed incorrect link to testing section (no such section, we have three of them - let's not confuse new contributors with a broken link)
- Fixed a typo

## Chores

- [x] linted
- [ ] tested
- [x] documented
- [ ] changelog entry

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ You can contribute to the project through:
 - proposing parts of the architecture that can be [extended](./EXTENDING.md)
 - improving [documentation](#Documentation)
 - tackling Big Issues from the [future roadmap](./ROADMAP.md)
-- improving [testing](#Testing)
+- improving testing
 - reviewing pull requests
 
 [jupyterlab-lsp]: https://github.com/krassowski/jupyterlab-lsp.git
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -74,6 +74,15 @@ jlpm build
 jlpm lab:link
 ```
 
+If `npm` dependencies did not get installed, you can trigger installation script for each package separately:
+
+```bash
+cd packages/jupyterlab-lsp
+jlpm install
+cd ../lsp-ws-connection
+npm install
```
bollwyvl: i really don't recommend `npm install` in a `yarn`-using project. What wasn't getting installed?

krassowski: things like `codemirror`, `vscode` protocol parts etc. I will do a clean test some time (no pressure to merge it, just wanted to share a solution which helped me so that we do not scare off prospective contributors)

bollwyvl: Looking at a local checkout, I see all of those in the root `node_modules`... yarn squashes everything all the way up into there to avoid the multiple solves, and manages them in the `yarn.lock`...

krassowski:
Closing for now; I added the typo and link fix in #212

PR #144, opened 2020-01-08
Fix R foreign extractor, add bash alias, add tests

krassowski:
Fix R foreign extractor, add bash alias, add notebook and extractors acceptance tests.

## Code changes

Added `03_Notebook.robot`, generalized some keywords relevant to both file editor and notebook and moved them to `Keywords.robot`.

## User-facing changes

Foreign code extractors for `%%bash` and `%%R` cells work now and again (respectively) :tada: 

## Chores

- [x] linted
- [X] tested
- [ ] documented
- [x] changelog entry

PR #153, opened 2020-01-12
Fix the (hopefully) last corner case of %R line magic

krassowski:
## References

Follows #144 with a fix for a case when `%R` line magic is used to import / export stuff but does not introduce any code to the R document, e.g. `%R -i data`.

## Code changes

In addition, the dummy assignments to introduce the imported varaibles are now all in one line with the line code; this might improve diagnostics until a better solution is in place.

## User-facing changes

None

None

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,8 @@
   - rpy2 magics received enhanced support for argument parsing
     in both parent Python document (re-written overrides) and
     exctracted R documents (improved foreign code extractor) (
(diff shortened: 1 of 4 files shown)


krassowski:
@bollwyvl If it passes (and there is no other PR in the meantime), this will be the last fix in the RC phase and I will then release 0.7.0!

PR #155, opened 2020-01-12
add some docs for customizing robot run

### LINKED ISSUE #151: How to run specific browser tests only
## What I am trying to do...

1. Run robot test for one test suite (`.robot` file) only, but not the others (since this is now possible thanks to #150)
2. Run a specific test case from this suite only

## How I would like to learn how to do it...

For (1) I tried `python scripts/atest.py -t 05_Features/Completion.robot` but it does not work as intended, I get:

```
attempt 1 of 1...
[ ERROR ] Suite 'Linux38' contains no tests matching name '05_Features/Completion.robot'.
```

I also tried a few answers from [SO](https://stackoverflow.com/q/25005277/6646912), but I guess that the ordering of arguments in our `atest.py` script may affect the ability to simply select one suite/case.

bollwyvl:
## References

- fixes #151

## Code changes

Will output what the arguments would have been for `robot`. Sorta.

## User-facing changes

None

None

## Chores

- [x] linted
- [x] tested
- [x] documented
- [ ] changelog entry

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,10 +121,13 @@ python scripts/utest.py
 ## Browser-based Acceptance Tests
 
 The browser tests will launch JupyterLab on a random port and exercise the
-Language Server features with [Robot Framework][] and [SeleniumLibrary][].
+Language Server features with [Robot Framework][] and [SeleniumLibrary][]. It
+is recommended to peruse the [Robot Framework User's Guide][rfug] (and the existing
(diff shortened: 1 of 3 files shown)


bollwyvl:
I'm mostly out for the rest of the day, hope this covers some bases!

krassowski:
That's perfect, thank you! Will merge just after 0.7.0 release and I am out too. This was very productive week, thank you very much for the help!

Issue #154, opened 2020-01-12
Allow to copy the diagnostics message from the Diagnostics Panel


krassowski:
## Elevator Pitch

Currently, it might be difficult to copy the inspection/diagnostic message from the panel.

## Motivation

Unexperienced users of any language tend to find themselves searching for the meaning of specific diagnostics online, e.g. on Stack Overflow. Having a way to copy the message easily can help them.

## Design Ideas

This might be difficult as programmatic copying in browsers might be restricted (and JupyterLab does not offer a copy option yet, though the recent web standards developments made other online editors reconsider and implement it already).

We could have a context menu action:
- "Search this diagnostic in XYZ" feature where XYZ is user-configurable search engine (e.g. Google or SO, probably default to SO).
- "Copy this message" which either:
   - opens a popup with an input field with allowing the user to easily copy the message
   - implements a novel programmatic copy action

Issue #128, opened 2020-01-05
Turn off warning diagnostics


austinmw:
## What I am trying to do...

How can I turn off the pycodestyle warning diagnostic underlines (or alternative, maybe disable only specific pycodestyle E#'s)? Some of these are a bit too verbose for me.

## How I would like to learn how to do it...

I don't see any documentation for this or how to modify any settings.

krassowski:
Thanks for writing. We plan to provide a configuration GUI at some time, but in the meantime, you can use the instructions on the [pyls](https://github.com/palantir/python-language-server) website (see [configuration](https://github.com/palantir/python-language-server#configuration)). For me on linux it simply meant creating a file called `~/.config/pycodestyle`; for my personal use I have the following settings:
```
[pycodestyle]
ignore = E402, E703
max-line-length = 120
```

E402 is to allow imports which are not on the very top of the file, E703 allows terminating semicolon (useful for matplotlib plots to prevent the object representation string).

For more check the [configuration section of pycodestyle documentation](https://pycodestyle.readthedocs.io/en/latest/intro.html#configuration). They describe where the configuration file should be created on different OS (e.g. Windows) and options and error codes are there.

Generally, it is a good idea to read-up on the language servers by visiting their GitHub pages as there may be useful information (such as configuration, tricks, and known limitations).

I will close this issue but feel free to post any additional questions or feedback here, or reopen this issue if any problems with the configuration appear.

bollwyvl:
Diagnostics follow some _pretty_ predictable patterns, and we've talked
about having an All The Diagnostics view. it wouldn't be _crazy_ to have
this be an UI entry point to "hide Diagnostics like this", with some basic
field matching (e.g. type, code, source, message regex, path glob), which
get stored in the lab config.

What this points to architecturally: it's worth looking at how tightly our
transport model is tied to our view model. We may need to think about ways
for such messages to be _hidden_ and not _destroyed_, as in the above page,
you'd want to see (at least) how many Diagnostics are hidden total, even if
you don't know which filter hides what. We wouldn't want to have this be
implemented anywhere near the actual codemirror stuff.

But generally... the documentation burden is pretty large on language
servers, especially ones with plugins, and it is not viable for this
project to document all of them.

What is probably a good idea in the near time is to have an optional
"doc_url" field (or family of fields) in the language server launcher spec
that would link to the upstream docs, and offer it in the UI: by making
them commands, in addition to wherever we might put it, the same url could
[...]

PR #159, opened 2020-01-18
Allow to hide chosen columns in diagnostics panel

### LINKED ISSUE #149: Allow to hide columns in diagnostics panel
Right-click on columns in the diagnostics panel should bring up a menu which would allow selecting which columns should be shown.

krassowski:
Allows choosing which columns to display/hide in the diagnostic panel, using a context menu action.

## References

Closes #149 

## Code changes

Greatly simplifies the `DiagnosticListing.render()` function body by moving the column-specific code to a newly created `Column` class (not exported).  A new interface `IListingContext` facilitates this transition.

## User-facing changes

Screenshot from 2020-01-18 11-59-47

None.

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
+++ b/packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
@@ -110,7 +110,10 @@ interface IDiagnosticsRow {
   data: IEditorDiagnostic;
   key: string;
   document: VirtualDocument;
-  cell_nr?: number;
+  /**
+   * Cell number is the ordinal, 1-based cell identifier displayed to the user.
+   */
+  cell_number?: number;
 }
 
 interface IListingContext {
@@ -223,8 +226,8 @@ export class DiagnosticsListing extends VDomRenderer<DiagnosticsListing.Model> {
     }),
     new Column({
       name: 'Cell',
-      render_cell: row => <td>{row.cell_nr}</td>,
-      sort: (a, b) => (a.cell_nr > b.cell_nr ? 1 : -1),
+      render_cell: row => <td>{row.cell_number}</td>,
+      sort: (a, b) => (a.cell_number > b.cell_number ? 1 : -1),
       is_available: context => context.editor.has_cells
     }),
     new Column({
@@ -265,19 +268,19 @@ export class DiagnosticsListing extends VDomRenderer<DiagnosticsListing.Model> {
     let by_document = Array.from(diagnostics_db).map(
       ([virtual_document, diagnostics]) => {
         return diagnostics.map((diagnostic_data, i) => {
-          let cell_nr: number = null;
+          let cell_number: number = null;
           if (editor.has_cells) {
             let notebook_editor = editor as VirtualEditorForNotebook;
             let { cell_id } = notebook_editor.find_cell_by_editor(
               diagnostic_data.editor
             );
-            cell_nr = cell_id + 1;
+            cell_number = cell_id + 1;
           }
           return {
             data: diagnostic_data,
             key: virtual_document.uri + ',' + i,
             document: virtual_document,
-            cell_nr: cell_nr,
+            cell_number: cell_number,
             editor: editor
           } as IDiagnosticsRow;
         });
--- a/packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
+++ b/packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
@@ -288,7 +288,7 @@ export class DiagnosticsListing extends VDomRenderer<DiagnosticsListing.Model> {
     let sorted_column = this.columns.filter(
       column => column.name === this.sort_key
     )[0];
-    let sorter = sorted_column.sort;
+    let sorter = sorted_column.sort.bind(sorted_column);
     let sorted = flattened.sort((a, b) => sorter(a, b) * this.sort_direction);
 
     let context: IListingContext = {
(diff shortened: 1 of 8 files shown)


bollwyvl:
Cool!
 
Maybe on this PR:

This is definitely something that we would start to want to save in the settings. I guess the schema could be like:

```
{
  "title": "Language Server",
  "description": "Language Server Protocol settings.",
  "type": "object",
  "properties": {
    "diagnostics": {
      "type": "object",
      "properties": {
        "panel_columns": {
          "type": "array",
          "items": {"type": "string"},
          "default": ["virtual_document", "message", "code", "severity", "source", "cell", "line", "ch"]
        }
      }
    }
  }
}
```
Hiding a column would remove it from the list, showing would put it back. In the future, they could even be reorderable. It's helpful to have a lenient schema... what if we discover new columns that need adding later?

Probably on this PR:

A nit on the UI: _multi-level_ menus _only_ triggerable by right-click are a discoverability/accessibility nightmare... you've already built something, so keep it there, for sure! But perhaps as a lead-in to stuff down the road, have a small `...` button right-most column that triggers the menu, and flatten everything into there. For now, each row would then be "wasted" space (but only 24x24). 

Not on this PR:

Each row could eventually have a `...` icon which had `Hide messages like this...` #128, `Copy to clipboard` #154, etc. that act at the row level. 

The header-level button might then eventually reveal something like:
```
|---------------------------------|>[...]
|     Copy All To Clipboard       |
|     5 Messages Hidden...        |
|---------------------------------|
|     Show All Columns            |
| [x] Virtual Document            |
  //... more columns           //
| [x] Ch                          |
|---------------------------------|

```

krassowski:
1. Settings and re-ordering: thought about it before, and I agree, but I wanted to keep the PRs small (so postponed for later).
2. I was thinking about two ways for actions:
    - using toolbar icons as the area widget has a toolbar (as does the notebook, or the recent log console, [see the log level filter screenshot here](https://github.com/jupyterlab/jupyterlab/pull/7406)); I would prefer the proposed `...` to got here (and maybe be more verbose than a hamburger?)
    - adding a dedicated column with buttons for [→] (jump to this diagnostic), [x] (hide this diagnostic" etc, but I think that this might be a waste of space and not the best UI. have not made up my mind yet, but definitely the next point (3):
3. Sounds like a good idea. I was thinking about putting it all to the context menu, but why not both? Would you mind opening an issue for this?

Also, it might be beneficial to make a review of existing practices in other editors so that we know what the users expect.

Another quick thought: if we were to use three dots, what about the vertical arrangement (⋮)?

krassowski:
@bollwyvl I am not sure about the yarn lock file, if the way the new dependency ((still) @phosphor/menu) got addedd looks strange to you feel free to correct me and teach how to do it better

packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
```diff
@@ -185,82 +277,58 @@ export class DiagnosticsListing extends VDomRenderer<DiagnosticsListing.Model> {
             data: diagnostic_data,
             key: virtual_document.uri + ',' + i,
             document: virtual_document,
-            cell_nr: cell_nr
+            cell_nr: cell_nr,
```
bollwyvl: what's an nr?

packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
```diff
@@ -185,82 +277,58 @@ export class DiagnosticsListing extends VDomRenderer<DiagnosticsListing.Model> {
...
-        </th>
-      );
-    };
+    let columns_headers = columns_to_display.map(column =>
+      column.render_header(this)
+    );
 
     return (
-      <table className={'lsp-diagnostics-listing'}>
+      <table className={DIAGNOSTICS_LISTING_CLASS}>
```
bollwyvl: +1. I'm a big fan of squooshing all of them into a single file someplace, e.g. `css.ts` rather than strewing them all over the place.

bollwyvl:
> Settings and re-ordering: 

Right. But trying to keep configurability in mind can save headaches down the line.

> area widget has a toolbar 

For the reasons mentioned elsewhere, more vertical height usually is a bummer. If it grows to a lot of things that need lots of explanation, perhaps. For example, if there was a filter box that let you add `path:*.py` to hide everything that wasn't currently python related, _that_ would be worth my space... sometimes. So perhaps toolbar features could be enabled by context/hamburger.

> I think that this might be a waste of space 

If we're talking one button that reveals one flat menu, I'll take 24px of horizontal off vs 24px off vertical any day. 

```
------------------------------------------------|
| document   | message   | line   | col   | ... |
|-----------------------------------------------|
| foo.py     | bad code  | 100    | 42    | ... |
|           -------------------------------/    | 
|           | Go to code                   |    |
| bar.py    | Copy to clipboard            |... |
|           | Filter messages like this... |    |
|           |------------------------------|    |
|
```

I think for complex sortable/filterable table, it's reasonable to have an escape hatch that lets you hide "table junk" until a user is ready for more complexity. 

> Would you mind opening an issue for this?

Incoming...

>  if we were to use three dots, what about the vertical arrangement (⋮) i

Ooh, look at that unicode. Good one. But yeah, that or something. I think people are pretty well trained at this point.

> yarn lock

If you add new deps, and also since a release recently went out, it's fine. Since a release recently went out, and you're changing something, it would be fine to rebuild it.

packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
krassowski: A leftover of my native language ;) **n**umbe**r**. Will expand the abbreviation.

krassowski:
> If we're talking one button that reveals one flat menu

Agreed, my "waste of space comment" was about my previous idea of having multiple buttons, your proposal is totally fine!

I will still advocate the use of toolbar for coherence with other widgets in the JupyterLab, but I see a very important use case for #160 - if we decide to allow to dock the diagnostics listing in a sidebar panel then we might want not to have the toolbar and here it is a big advantage to have it as proposed.

PS. I am going to work on improving the completion now (a few hours this week and even more next week), also maybe "quick-fix" feature as those seem more urgent right now.

PR #165, opened 2020-01-19
Refactor LSPConnection, ConnectionManager

bollwyvl:
## References

This has been mentioned a few places (e.g. #76), but doesn't have a concrete issue. Whoops!

## Code changes

Changes the design contract to support:
- exactly _one_ `ConnectionManager` per application (instead of one per Notebook/Editor)
- _one_ `LSPConnection` per language (instead of one per Editor/Cell(!))
  - therefore one WebSocket
- in at least one feature `highlight.ts`, instead of waiting for the `emit`, the results are consumed directly behind a promise (or `Thenable`, but who's counting)
  - we probably want to use this pattern all over for request/response style messages
  - otherwise, i was getting duplicate "same token" highlights in totally crazy places in unrelated editors
  - we need to _generally_ start testing with multiple documents open, as these might already be issues

In addition, this lazy-loads the `connection.ts` and all of `lsp-ws-connection` when the first language server is requested via `await import`.

## User-facing changes

Apart from the initial delay loading the 1.5mb slug of `lsp-ws-connection`, this generally makes things _much_ snappier, as each additional document/cell adds very little overhead.

No doubt, a lot of things that were expecting to "just work" will need to be more aware of the `documentUri`... I haven't merged up to `master`, so perhaps the recent completion fixes will uncover new issues.

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


bollwyvl:
Probably need to put this down for a bit.

- all of the acceptance tests are passing. I've `|| echo`d the jest tests for now on azure so i can get some eyes on other OSes.
- most of the unit tests working, and will be able to figure out the `editor is undefined`, but haven't dug into what rename is doing, so can't really tell why they might be failing.

bollwyvl:
Having a look on [Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-one-socket-per-lang?urlpath=lab)...

bollwyvl:
Curse you, codebeat. Anyhow, will remove some of the `[wip]`, as this is now passing on travis and AP.

packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics.ts
```diff
@@ -273,20 +268,21 @@ export class Diagnostics extends CodeMirrorLSPFeature {
               start_in_root
             );
           } catch (e) {
-            console.log(e, diagnostics);
+            DEBUG && console.log(e, diagnostics);
```
krassowski: I would prefer to move this to [console](https://github.com/krassowski/jupyterlab-lsp/blob/master/packages/jupyterlab-lsp/src/virtual/console.ts) - add `console.debug` and have a switch over there rather than scattered constants everywhere. Then one day, ideally the switch would be exposed in the console interface so we could have an official guide on how to debug from within the Lab without rebuilding.

krassowski:
> per language (instead of one per Editor/Cell(!)) 

Was it really one per cell? I thought that it was one per virtual editor (which, especially in notebook, aggregates multiple physical editors).

Either way, I agree on the direction, but I do not think that we should stop here (I feel uncomfortable about passing around the document info everywhere - if this is always passed,  then we probably should adapt the class hierarchy - which is overdue a refactor already). On the other hand, I do not want to start to break things yet (even though it accumulates some debt) as I would like to have higher browser tests coverage first. We may want to have a call with a whiteboard on this sometime in the future.

The remarks above are not an encouragement to expand this PR - its fine as it is, rather general comments on the plans for future.

packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics.ts
bollwyvl: Yes, sounds great. The console stuff scattered around everywhere made it very had to track down what was broken, and the event emitter stuff was coming from a deep, dark hole, hidden inside the webpack-of-neccessity.

If configurable, I'd see it as having a set of `Token`s per feature/component or other key that can be set to different log levels.

bollwyvl:
> Was it really one per cell? 

I don't know. So many websockets got created (and left open) when trying to start a connection (especially on slower connections/servers) that it was hard to tell.

> passing around the document info everywhere 

This was to maintain semi-compatibility with the existing `lsp-ws-connection` (replace `this.documentInfo` with `documentInfo) while working on it (see debuggability of promises vs events, etc). 

> probably should adapt the class hierarchy 

In now having done a more thorough read-through of that layer of the code (i tried to stay out of other layers, when possible) i'm pretty sure we don't need it. We still need the webpack for `MessageConnection` (making it work simply in the browser has been `wontfix`ed upstream), but then I never want to have to read that code more than I have, and can pretty implicitly trust that it works

> We may want to have a call with a whiteboard on this sometime in the future.

Always a challenge with timezones! Perhaps as an entree to the conversation i whip up the previously aforementioned `docs` so we can put some graphviz/drawio/whatever. Of course we have bigger docs to fry, but this might be a forcing function.

>  do not want to start to break things yet (even though it accumulates some debt) as I would like to have higher browser tests

Agreed. I'll get it back to green, and can take out the DEBUG stuff, but otherwise don't think it can be broken into smaller pieces. About the green: can we turn off codebeat? I don't even read it, and it makes my browser tabs less useful. There are other things out there that not only pick your nits, but also suggest changes to _fix_ them.

atest/Keywords.robot
```diff
@@ -207,7 +217,7 @@ Place Cursor In Cell Editor At
     Execute JavaScript    return document.querySelector('.jp-Cell:nth-child(${cell_nr}) .CodeMirror').CodeMirror.setCursor({line: ${line} - 1, ch: ${character}})
 
 Wait Until Fully Initialized
-    Wait Until Element Contains    ${STATUSBAR}    Fully initialized    timeout=35s
+    Wait Until Element Contains    ${STATUSBAR}    Fully initialized    timeout=60s
```
bollwyvl: Seemed to reduce flake a fair amount: on a fresh browser load, the lazy load does make the first connection slower to start, so this time isn't getting rolled up into the aggressive up-front animation waiting.

azure-pipelines.yml
```diff
@@ -22,6 +22,9 @@ variables:
   THIRD_PARTY_LABEXTENSIONS: >-
     @krassowski/jupyterlab_go_to_definition
 
+  LINKED_EXTENSIONS: >-
```
bollwyvl: should have done this _ages_ ago

bollwyvl: it's not testing the as-built tarball, but what can you do? that would be an advantage of killing the intermediate layer, as the entire module would be:

```ts
export import { ConsoleLogger, listen, MessageConnection } from 'vscode-ws-jsonrpc';
```
...plus 10 files of boilerplate, and would only need to be updated once an upstream release (if then).

krassowski:
> About the green: can we turn off codebeat? I don't even read it, and it makes my browser tabs less useful. There are other things out there that not only pick your nits, but also suggest changes to fix them.

Agreed. I mean, it is sometimes useful but no longer in this project for me too. It should be turned off by now.

bollwyvl:
Oh, sorry: I'll put WIP back on this. I want to:
- merge the docs in, and show what i'm actually doing
- track down some memory leaks
  - no doubt present in the existing stuff... dangling `eventListeners` on codemirror or something
- and maybe add some more tests

bollwyvl:
Sorry for the radio silence: I think I've finally clobbered all of the memory leaks. The primary remaining development task is figuring out what to do with the three remaining `connection_handlers`. Presently doing local testing with "open ten notebooks" to actually let me know when stuff starts failing.

bollwyvl:
Ok: i ended up just cranking the `maxListeners` to a ludicrous number so that I didn't have to do too much more stuff encroaching on #176 for diagnostics. I am not sure about `hover` and `highlight`: i _think_ i can remove those `connection_handlers`, but can quite tell. Anyhow, locally the tests pass.

A significant change is moving the `pathChanged` concern (#141) up to the root level and just destroying the adapter from the outside rather than trying to make it handle getting renamed.

Most of the memory leak work i did is defeated by the amount of debug output: basically any hot object not cast to a string becomes uncollectable. We'll need to do a full sweep of what the approach needs to be.

bollwyvl:
Also taking a look on binder to see if the slow connection issues are cleared up:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-one-socket-per-lang?urlpath=lab)

bollwyvl:
Ugh. It looks like binder installs r-studio from apt/random shell scripts now: no idea what that's going to do to the r language server install we have.

bollwyvl:
Hooray CI and binder both working the first time :P

bollwyvl:
Still haven't merged in docs branch, but am otherwise starting to feel like this is done...

bollwyvl:
Noticed some more stuff in looking into #190, doing some more work here...

bollwyvl:
Similar to https://github.com/krassowski/jupyterlab-lsp/pull/165#issuecomment-581177638, i've also moved the `kernelChanged` handling up to the plugin level.

bollwyvl:
Of note to the above: it's possible for a file editor to _also_ have a kernel attached to it and provide completion (and even execution). Not going to handle on this PR, but probably also needs some love.

bollwyvl:
Ah: that change won't work, because the `kernel_info` never comes back from the `No Kernel`. This is probably not desirable, will just revert.

bollwyvl:
Well, the docker test is just being a jerk on osx, but I really can't think of anything else I want to do on this, having done the docs piece directly over on #177. I feel like we should probably merge this and move forward.

krassowski:
Will do some local testing and merge today!

bollwyvl:
🛸

krassowski:
Thank you!

PR #170, opened 2020-01-20
fix windows server_root

yunair:
## Code changes

correct windows User server root path

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/index.ts
+++ b/packages/jupyterlab-lsp/src/index.ts
@@ -92,6 +92,15 @@ const plugin: JupyterFrontEndPlugin<void> = {
           'Guessing Mac the server root using user settings path',
           server_root
         );
+      } else if (user_settings.startsWith('C:/Users/')) {
+        server_root = server_root.replace(
+          '~',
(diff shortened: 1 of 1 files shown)


krassowski:
Hi @yunair, thanks for this contribution. With this fragment of code getting bigger and bigger, it would be worth moving the 74-109 to a separate function.

Let's call it `function deduceAbsolutePath`; it could accept two arguments: `(server_root: string, user_settings: string)`; once the code is moved out to a new function it could be easily tested.

I can help you with that, but basically we would move this new function to `utilis.ts`, create a new file `utilis.spec.ts` and write tests using the same tools as in [positioning.spec.ts](https://github.com/krassowski/jupyterlab-lsp/blob/master/packages/jupyterlab-lsp/src/positioning.spec.ts).

Do you think that you could follow up on this with the function extraction and tests?

bollwyvl:
Can we not just have the backend give an authoritative answer via
PageConfig?

yunair:
Sorry for my late apply, I think get from backend is the better way, I'll try late

PR #177, opened 2020-01-24
Start sphinx docs

bollwyvl:
## References

- #169 install docs started
- #165 as-is architecture started
  - will probably upstream this there so i can start documenting what the proposed changes might be

## Code changes

Goals for this PR:
- [ ] collect envs/reqs into a folder
  - [x] Adds new (nb)sphinx env/requirements
- [x] Adds doc script
- [x] Handle some initial documentation tasks
- [ ] icon
- [ ] review some best practices
  - [ ] https://www.divio.com/blog/documentation/
- [ ] Add docs to automation
  - [ ] integrity
  - [ ] linting
  - [ ] CI
- [ ] integrate existing content:
  - [ ] ROADMAP
  - [ ] README 
  - [ ] CONTRIBUTNG
  - [ ] LANGUAGESERVERS
  - [ ] EXTENDING

(Maybe) non-goals for this PR:
- [ ] add readthedocs
- [ ] add screenshots/gifs from acceptance tests

## User-facing changes

None, yet

## Backwards-incompatible changes

None, yet

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

Issue #163, opened 2020-01-19
Build/test some minimal, non-full-Lab examples


bollwyvl:
## Elevator Pitch

A la [jupyterlab/jupyterlab](https://github.com/jupyterlab/jupyterlab/tree/master/examples) it would be useful to show how the existing/to-be pieces could be used without the full lab environment.

## Motivation

To help move towards #76, this would help determine what parts of a full Lab are absolutely required, and potentially break up future packages according to their heavyweight dependencies. This would also motivate things that we can move into the editor itself, rather than building new fancy UI in new panels.

## Design Ideas

- just an editor (no dockpanel)
- just a notebook (no dockpanel)
- minimal multi-editor dockpanel (with file tree, diagnostic panel)
- console (once we support console)

I think any more than that, and you've basically got a full lab.

PR #180, opened 2020-01-25
reduce number of top-level files

bollwyvl:
## References

- blocks #177 https://github.com/krassowski/jupyterlab-lsp/pull/177#discussion_r370933583

## Code changes

Moves a lot of things around. There is a hard reference to `LANGSERVERS.md` in the source: we should probably just inline that or something... or point to `readthedocs` when we do that.

## User-facing changes

The repo is a tad cleaner on first blush:
- fits on a full screen, meh
  - Screenshot from 2020-01-25 15-19-16

_shouldn't_ create any issues on PRs...

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


bollwyvl:
To _really_ go after it, we need all the meta junk from python:
- `setup.{cfg,py}`
- `MANIFEST.in`
and js:
- `package.json`
- `lerna.json`
- `yarn.lock`
- `.prettierignore`
.. in separate dirs altogether. I'm happy to do that (once I get at least _this_ working), but that will start to impact the usability, i fear.

bollwyvl:
Another thought: if the DX is indeed getting onerous, we can fix it Once And For All and introduce a top-level automation tool, and indeed, that might even be worthwhile, especially as we increase the number of built artifacts (#177, #163). If we want to go that way, this PR is a _fine_ place to do it: i find it onerous to incrementally change automation if we can get consensus vision of where we want to go.

`make` would be great, but in 2020, it's still hard to reliably get a working windows toolchain, same goes for node-based stuff, so i'd favor [doit](https://pydoit.org/). Others i've tried include `pants` and `scons`: again, they are rather heavy opinions.

Positing `doit` (but likely very similar for other tools): this would replace `scripts` with a single, top-level `dodo.py`, and after a `pip install -e .` (and getting node somehow), you'd type `doit <thing>`. Running without any args would basically run the entire azure pipeline locally, which is actually _fantastic_ for reviewing PRs and CI. If you do a touch of extra work, it can do the make-like thing: e.g. if you change the ts, but don't touch the python, it won't even run the python unit tests, but _would_ run the acceptance tests (after first rebuilding your `lab` so you don't get the `thinger blocked by dialog` warning).

Anyhow, food for thought...

bollwyvl:
Well, I guess I'll take it passing the first time. No doubt i missed some stuff in the markdown links, but that will be tested aggressively on #177, so less inclined to worry about it at this point...

.prettierrc
```diff
@@ -1,3 +0,0 @@
-{
-  "singleQuote": true
```
krassowski: `.prettierrc` got lost in the transition. It was there for interoperability of the codebase with the core (https://github.com/jupyterlab/jupyterlab/blob/master/.prettierrc). Probably could go to `packages`.

package.json
```diff
@@ -28,6 +28,9 @@
       "pre-commit": "precise-commits"
     }
   },
+  "prettier": {
```
bollwyvl: This is where `.prettierrc` went!

https://github.com/krassowski/jupyterlab-lsp/pull/180#discussion_r370960292

packages/jupyterlab-lsp/src/adapters/jupyterlab/components/statusbar.tsx
```diff
@@ -211,7 +211,7 @@ class LSPPopup extends VDomRenderer<LSPStatus.Model> {
           Documentation:{' '}
           <a
             href={
-              'https://github.com/krassowski/jupyterlab-lsp/blob/master/LANGUAGESERVERS.md'
+              'https://github.com/krassowski/jupyterlab-lsp/blob/master/docs/LANGUAGESERVERS.md'
```
krassowski: One lazy option would be to point to a version-specific commit and have a link to master in the document itself (something like "there might  be an updated version of this document >click here<".)

PR #183, opened 2020-01-26
[wip] add doit

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-doit?urlpath=lab)

This is an exploration of refining the developer experience with a file-based build automation system.

In addition to the fine-grained tasks, there are a few common ones:

- `LAB` binder runs this. sets up the js/python for a working dev lab
- `LINT` format/lints everything
- `ALL` is heavy, but should be useful when switching branches
- `WATCH` is okay, but interleaving the `tsc`, `webpack` and `lab` output is pretty busy...
- `RELEASE` is useful for just getting to some files
  - could be extended to _actually_ do the release, but meh...

tasks

(if we want to keep `doit`, we can generate/customize the above graph in #177 with `doit-graph`)

## References

- #180 https://github.com/krassowski/jupyterlab-lsp/pull/180#issuecomment-578440914

## Code changes

- [x] adds a `make`-like system with `doit`
- [x] use in binder
- [ ] use in CI

## User-facing changes

None

- `lint.py` is removed in favor of the more granular approach
- some output paths were changed to make them more predictable

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


bollwyvl:
I'd still like to see a top-level automation tool that wasn't (entirely) hand-cobbled scripts, but this has bitrotted fairly badly at this point.

Issue #190, opened 2020-02-08
Use mime types from server spec for language detection


bollwyvl:
## Elevator Pitch

If available, use the mime types provided by specs to detect languages rather than (just) string parsing or lookup tables.

## Motivation

#189 points out a case where a language doesn't (yet) have a well-defined language type (https://github.com/robots-from-jupyter/robotkernel/issues/40).

## Design Ideas

In addition to the [lookup table](https://github.com/krassowski/jupyterlab-lsp/blob/master/packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts#L92) of languages, detection should also consult the [server spec](https://github.com/krassowski/jupyterlab-lsp/blob/master/py_src/jupyter_lsp/schema/schema.json#L137) mime types. 

It may be that we need to open this up even wider, and allow for server specs to offer the full rosetta stone of `mimetype`, `codemirror_mode` and `name`, depending on how much information might be available.

krassowski:
Agreed, this the right thing to do. However, we need to adjust the schema slightly as we need mimetype → language mapping, as one server can handle multiple languages.

krassowski:
I believe we already do that, so closing as completed. Any future enhancements to the logic would need to be done in `@jupyterlab/lsp` package.

PR #199, opened 2020-02-15
Use language server implementation instead of language for URLs

### LINKED ISSUE #192: Use implementation name rather than language for WebSocket URLs
## Elevator Pitch

Use a more precise definition of _which_ language server should be used for a connection in URLs.

## Motivation

At present, the WebSocket URL scheme is rather nebulous in that _all_ language servers that register for a language will be connected during a `Session`, which is probably not right and would no doubt explode if there were multiple providers for a given language. Further the "language" concept doesn't _really_ provide enough depth. Combined with #190, we should have enough information to do a much better job of detecting (or deferring to the user) which implementation to use in a particular case.

Additionally, we can save extra connections when multiple documents of different languages connect to the same polyglot server, rather than hiding it.

## Design Ideas

While we're at it, let's add another (short) prefix to the WebSocket URLs so that we are less constrained in potential future endpoints.

| Old | New |
|-|-|
| `/lsp` | `/lsp/status` |
| `/lsp/python`     | `/lsp/ws/pyls` |
| `/lsp/javascript`<br/>`/lsp/jsx`<br/>`/lsp/typescript`<br/>`/lsp/typescript-jsx`<br/>`/lsp/typescriptreact`<br/>`/lsp/javascriptreact` | `/lsp/ws/javascript-typescript-langserver` |

bollwyvl:
## References

- fixes #192

## Code changes

- backend
  - [x] uses implementation name for websocket key, e.g. `/lsp/ws/pyls`
  - [x] moves status to `/lsp/status`
  - [x] move reading/writing to stdio language servers to threads
  - [x] move reading/writing to shadow files to threads

- [x] fix frontend
  - [x] introduce `ILanguageServerManager`
  - [x] use `ILSM` in `ConnectionManager` and `StatusBarItem`

## User-facing changes

- None, other than launching a document with a language already supported by another language will be faster, as there's only one actual WebSocket/process behind them

- [x] bumps schema to version `2`
  - [x] changes session payload to be a dict, e.g. `{"session": {"pyls": {}}}`

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry


bollwyvl:
starting point for getting the frontend to compile:
```
../jupyterlab-lsp/src/adapters/jupyterlab/components/statusbar.tsx(341,7): error TS2740: Type '{ [k: string]: LanguageServerSession; }' is missing the following properties from type 'LanguageServerSession[]': length, pop, push, concat, and 26 more.
```
here we go...

py_src/jupyter_lsp/tests/conftest.py
```diff
@@ -12,45 +12,39 @@
...
-    "json",
-    "jsx",
-    "less",
-    "markdown",
-    "python",
-    "scss",
-    "typescript-jsx",
-    "typescript",
-    "yaml",
+KNOWN_SERVERS = [
```
bollwyvl: nice thing here is it is now slightly faster...

bollwyvl:
Turns out I didn't _have_ to do #190, so I didn't. Otherwise seems pretty good!

atest/01_Editor.robot
```diff
@@ -19,7 +19,7 @@ CSS
 
 Docker
     ${def} =    Set Variable    xpath://span[contains(@class, 'cm-string')][contains(text(), 'PLANET')]
-    Editor Shows Features for Language    Docker    Dockerfile    Diagnostics=Instruction has no arguments    Jump to Definition=${def}    Rename=${def}
+    Wait Until Keyword Succeeds  3x  100ms  Editor Shows Features for Language    Docker    Dockerfile    Diagnostics=Instruction has no arguments    Jump to Definition=${def}    Rename=${def}
```
bollwyvl: So this "fixes" the problem, but the fact remains that you don't reliably get diagnostics back from a dockerfile on the first file opened until you change the file (then it seems to work every time). I tried sending more change events, but it didn't have an effect. I'm willing to deal with it if it makes the tests pass for now.

py_src/jupyter_lsp/stdio.py
```diff
@@ -34,6 +40,7 @@ def __repr__(self):  # pragma: no cover
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.log.debug("%s initialized", self)
+        self.executor = ThreadPoolExecutor(max_workers=1)
```
bollwyvl: I thought all of the jiggery pokery i did with the streams was sufficient to handle this, but I've still hit some slowdown. This should Make Sure we don't block the main thread on IO below

py_src/jupyter_lsp/stdio.py
```diff
@@ -87,7 +94,7 @@ def wake(self):
                 else:
                     self.wake()
 
-                await self.queue.put(message)
+                IOLoop.current().add_callback(self.queue.put_nowait, message)
```
bollwyvl: again, I don't _think_ this was causing issues, but better be safe...

py_src/jupyter_lsp/stdio.py
```diff
@@ -98,12 +105,12 @@ def wake(self):
         message = ""
         headers = HTTPHeaders()
 
-        line = self._readline()
+        line = await convert_yielded(self._readline())
```
bollwyvl: there are other wrappers that wouldn't require converting, but this works...

py_src/jupyter_lsp/stdio.py
```diff
@@ -142,17 +150,21 @@ class LspStdIoWriter(LspStdIoBase):
...
-                self.stream.flush()
+                await convert_yielded(self._write_one(response.encode("utf-8")))
             except Exception:  # pragma: no cover
                 self.log.exception("s couldn't write message: %s", self, response)
             finally:
                 self.queue.task_done()
+
+    @run_on_executor
+    def _write_one(self, message) -> None:
+        self.stream.write(message)
```
bollwyvl: could have moved more string operations here, but keeps the critical region more clear

packages/jupyterlab-lsp/src/virtual/editor.spec.ts
```diff
@@ -59,6 +60,17 @@ describe('VirtualEditor', () => {
...
   );
 
+  const LANGSERVER_MANAGER = new MockLanguageServerManager({});
+  const CONNECTION_MANAGER = new DocumentConnectionManager({
+    language_server_manager: LANGSERVER_MANAGER
+  });
+
+  const DEBUG = false;
+
+  if (DEBUG) {
```
krassowski: a debug - left intentionally or just slipped in?

bollwyvl: yeah, accidental!

bollwyvl:
I'll have a look at fixing the merge conflicts presently...

krassowski:
Ok, I will wait!

packages/jupyterlab-lsp/src/virtual/editor.spec.ts
bollwyvl: Nope, needs it! Have changed to make it part of the actual test environment.

bollwyvl:
Ok, provided docker doesn't flake out on us, this is good to go if CI passes.

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -16,11 +23,19 @@ def extract_or_none(obj, path):
 
 
 class EditableFile:
+    executor = ThreadPoolExecutor(max_workers=MAX_WORKERS)
```
bollwyvl: following the same approach as stdio, but in this case we have one executor for all instances of `EditableFile`. 4 seems like enough for now.

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -16,11 +23,19 @@ def extract_or_none(obj, path):
...
     def __init__(self, path):
         # Python 3.5 relict:
         self.path = Path(path) if isinstance(path, str) else path
-        self.lines = self.read_lines()
 
+    async def read(self):
+        self.lines = await convert_yielded(self.read_lines())
+
+    async def write(self):
+        return await convert_yielded(self.write_lines())
```
bollwyvl: these just wrap the underlying tornado yield-style async stuff.

py_src/jupyter_lsp/virtual_documents_shadow.py
```diff
@@ -125,7 +141,9 @@ def setup_shadow_filesystem(virtual_documents_uri):
             return
 
         path = file_uri_to_path(uri)
-        file = EditableFile(path)
+        editable_file = EditableFile(path)
+
+        await editable_file.read()
```
bollwyvl: renaming is cosmetic, just don't need to go re-defining builtins without a good reason.

It's a little ugly to have to do the read: could add an `async` class method to `EditableFile` to do both in one go...

bollwyvl:
have definitely still been getting the dockerfile issue locally... very challenging to reproduce, apparently...

py_src/jupyter_lsp/virtual_documents_shadow.py
krassowski: It's a small hill, but [file is not a builtin in Python3](https://stackoverflow.com/a/24942363/6646912) and it is ok to use it. It was in Python2, and yes some syntax highlighters did not catch up, but we can use it now.

bollwyvl:
ha, fair.

bollwyvl:
I swear I'm done tinkering now.

krassowski:
Now let's hope we will have CI pass on master ;)

bollwyvl:
Hooray merge! Boo ci fail! We could mark it as non-critical, but those have
a tendency to get ignored. looks like there's a new version we could try. I
guess we need an issue.

Aside from fixing the docker issue, what's the next thing we want to get
to? With this PR in, per-server configuration is unblocked, and would help
a lot, as is the mimetype-based detection stuff.

I'm also interested in the multiple connection providers: being able to
install js-only servers as lazy loading labextensions would help the
install complexity around run-time node without negatively impacting
performance much: and json and markdown would be great initial candidates,
especially if more things like settings manager/cell metadata and markdown
cells could be extended.

>

PR #202, opened 2020-02-23
sendOpen after connection handlers are installed

### LINKED ISSUE #201: Dockerfile diagnostics don't always show up on first load
## Description

When opening a dockerfile with errors in an editor, the diagnostics don't appear until a change is made.

## Reproduce

1. Open the binder
2. Open `atest/examples/Dockerfile`
3. _Sometimes_ the third line, `RUN` will contain a diagnostic about an argument being required, but sometimes not

This frequently, and repeatedly, occurs during acceptance tests.

## Expected behavior

Diagnostics should always show (as soon as they are available) on first load.

## Context

- Operating System and version: debian in docker
- Browser and version: chromium 80.0.3987.116 , firefox 68.5.0 esr
- JupyterLab version: 1.2.6
- `jupyter-lsp` version: `master`
- `@krassowski/jupyterlab-lsp` version: `master`

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/fix-docker-diagnostics?urlpath=lab%2Ftree%2Fatest%2Fexamples%2FDockerfile)

## References

- fixes #201

## Code changes

- [x] upgrade to newest `"dockerfile-language-server-nodejs": "^0.0.22",`

## User-facing changes

- [x] dockerfile diagnostics show up every time

N/A

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
   - ignores malformed diagnostic ranges, enabling markdown support ([#165][])
   - passes tests on Python 3.8 on Windows ([#165][])
 
+- bug fixes
+
+  - reports files are open only after installing all handlers to avoid missing messages ([#201][])
+
+[#201]: https://github.com/krassowski/jupyterlab-lsp/issues/201
+
 ### `lsp-ws-connection 0.4.0` (unreleased)
 
 - breaking changes
--- a/ci/job.docs.yml
+++ b/ci/job.docs.yml
@@ -31,7 +31,7 @@ jobs:
               - script: ${{ parameters.env_docs }} || ${{ parameters.env_docs }} || ${{ parameters.env_docs }}
                 displayName: docs dependencies
 
-              - script: ${{ platform.activate }} && conda uninstall -y pytest-cov
+              - script: ${{ platform.activate }} jupyterlab-lsp && conda uninstall -y pytest-cov
                 displayName: remove coverage to prevent warning
 
               - script: ${{ platform.activate }} jupyterlab-lsp &&  jlpm || jlpm || jlpm
--- a/ci/job.docs.yml
+++ b/ci/job.docs.yml
@@ -31,7 +31,7 @@ jobs:
               - script: ${{ parameters.env_docs }} || ${{ parameters.env_docs }} || ${{ parameters.env_docs }}
                 displayName: docs dependencies
 
-              - script: ${{ platform.activate }} jupyterlab-lsp && conda uninstall -yf pytest-cov
+              - script: ${{ platform.activate }} jupyterlab-lsp && conda uninstall -y --force-remove pytest-cov
                 displayName: remove coverage to prevent warning
 
               - script: ${{ platform.activate }} jupyterlab-lsp &&  jlpm || jlpm || jlpm
--- a/ci/job.docs.yml
+++ b/ci/job.docs.yml
@@ -31,7 +31,7 @@ jobs:
               - script: ${{ parameters.env_docs }} || ${{ parameters.env_docs }} || ${{ parameters.env_docs }}
                 displayName: docs dependencies
 
-              - script: ${{ platform.activate }} jupyterlab-lsp && conda uninstall -y pytest-cov
+              - script: ${{ platform.activate }} jupyterlab-lsp && conda uninstall -yf pytest-cov
                 displayName: remove coverage to prevent warning
 
               - script: ${{ platform.activate }} jupyterlab-lsp &&  jlpm || jlpm || jlpm
--- a/ci/job.docs.yml
+++ b/ci/job.docs.yml
@@ -31,6 +31,9 @@ jobs:
               - script: ${{ parameters.env_docs }} || ${{ parameters.env_docs }} || ${{ parameters.env_docs }}
                 displayName: docs dependencies
 
+              - script: ${{ platform.activate }} && conda uninstall -y pytest-cov
+                displayName: remove coverage to prevent warning
+
               - script: ${{ platform.activate }} jupyterlab-lsp &&  jlpm || jlpm || jlpm
                 displayName: install npm dependencies (for language servers)
 
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   },
   "devDependencies": {
     "bash-language-server": "^1.6.1",
-    "dockerfile-language-server-nodejs": "^0.0.21",
+    "dockerfile-language-server-nodejs": "^0.0.22",
     "eslint": "^5.16.0",
     "eslint-config-prettier": "^4.1.0",
     "eslint-plugin-prettier": "^3.0.1",
--- a/packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
+++ b/packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
@@ -294,8 +294,7 @@ export abstract class JupyterLabWidgetAdapter
       this.on_foreign_document_opened,
       this
     );
-
-    await this.connect(virtual_document).catch(console.warn);
+    return await this.connect(virtual_document).catch(console.warn);
   }
 
   protected async on_foreign_document_opened(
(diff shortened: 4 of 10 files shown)


bollwyvl:
Some notes:

- showing diagnostics
Screenshot from 2020-02-22 21-35-32

- not showing diagnostics
Screenshot from 2020-02-22 21-34-12

It looks like when diagnostics are not showing, perhaps the document hasn't been fully initialized.

bollwyvl:
Looks like the event is being fired before the diagnostic listener is actually installed:
Screenshot from 2020-02-22 21-48-30

So apparently, the docker server is just _really fast_ to initialize and report diagnostics.

My first attempt will be to have adapter take over signaling that it wants to open a document by enqueuing this with the connection, rather than have the connection manager handle it...

package.json
```diff
@@ -3,7 +3,7 @@
   },
   "devDependencies": {
     "bash-language-server": "^1.6.1",
-    "dockerfile-language-server-nodejs": "^0.0.21",
+    "dockerfile-language-server-nodejs": "^0.0.22",
```
bollwyvl: unnecessary, but not hurting anything, either...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -294,21 +294,29 @@ export abstract class JupyterLabWidgetAdapter
       this.on_foreign_document_opened,
       this
     );
-
-    await this.connect(virtual_document).catch(console.warn);
+    return await this.connect(virtual_document).catch(console.warn);
```
bollwyvl: rather than changing the output signature of this, we could change the input signature and add an `deferOpen = false`: foreign documents would be opened immediately, presumably, as all the features would already be connected.

packages/jupyterlab-lsp/src/connection.ts
```diff
@@ -14,8 +14,26 @@ import { until_ready } from './utils';
...
 
 export class LSPConnection extends LspWsConnection {
+  protected documentsToOpen: IDocumentInfo[];
+
   constructor(options: ILSPOptions) {
     super(options);
+    this.documentsToOpen = [];
+  }
+
+  sendOpenWhenReady(documentInfo: IDocumentInfo) {
```
bollwyvl: not sure if this is the best way, but seemed the most straightforward. i guess it could notify that it did open each document, but nobody would be listening right now.

bollwyvl:
Checking it out on binder (link added to top), but pretty bullish on this fixing #201.

bollwyvl:
Yep, looks good on binder.

ci/job.docs.yml
```diff
@@ -31,6 +31,9 @@ jobs:
               - script: ${{ parameters.env_docs }} || ${{ parameters.env_docs }} || ${{ parameters.env_docs }}
                 displayName: docs dependencies
 
+              - script: ${{ platform.activate }} && conda uninstall -y pytest-cov
+                displayName: remove coverage to prevent warning
```
bollwyvl: adding this to remove warning, but also to trigger another run (haven't found anything else worth trying/changing)

packages/jupyterlab-lsp/src/connection.ts
krassowski: Yes, we are not allowed to send anything before receiving `InitializeResult`:

> Until the server has responded to the initialize request with an InitializeResult, the client must not send any additional requests or notifications to the server

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
krassowski: yes, please - this would make it much easier to understand in the future!

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -294,21 +294,29 @@ export abstract class JupyterLabWidgetAdapter
...
       this.on_foreign_document_closed,
       this
     );
+
+    if (connection_context) {
+      connection_context.connection.sendOpenWhenReady(
+        foreign_document.document_info
+      );
+    } else {
+      console.warn(`Connection for ${foreign_document.path} was not be opened`);
```
krassowski: typo/grammar in `was not be opened` fragment

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
bollwyvl: sure, pushed this around. `on_foreign` and `reload_connection` will both try to `sendOpen` as the adapter will have been fully initialized by that point, i'm pretty sure.

packages/jupyterlab-lsp/src/connection.ts
bollwyvl: we should change ws-connection to complain much more loudly about that... but maybe not on this PR.

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
bollwyvl: fixed

bollwyvl:
Thanks for the quick reviews. Provided this passes (doing a full run locally, too), and you don't see any other glaring issues, I don't plan on doing anything else on it at this time! :rocket: :guitar:

packages/jupyterlab-lsp/src/connection.ts
bollwyvl: another approach: once this is in, instead of failing (or dropping messages on the floor, as it currently does), add `Connection.ready: Promise<void>`, and have everything except `initialize` do an `await` on that. Would have to do some digging around to see if that's viable.

bollwyvl:
Local test looks good, trying binder.

ci/job.docs.yml
bollwyvl: hopefully right/faster now...

bollwyvl: now takes 6s rather than 1min, though I did forget to add pip to the env... we'll see how the current run goes: not strictly required for merge.

packages/jupyterlab-lsp/src/connection.ts
bollwyvl: tried this out on a branch... breaks a bunch of things. doesn't mean it's not worth doing, just needs more than some simple checks...

bollwyvl:
last fail was pure-and-simple azure flake..

bollwyvl:
Back to happy...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -287,36 +287,62 @@ export abstract class JupyterLabWidgetAdapter
     });
   }
 
-  protected async connect_document(virtual_document: VirtualDocument) {
+  /**
+   * Opens a connection for the document. The connection may or may
+   * not be initialized, yet, and depending on when this is called, the client
+   * may not be fully connected. In
```
krassowski: In?

bollwyvl: fixed. also made calls to `connect_document` more explicit, providing `false` (though left the default value).

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@
   - ignores malformed diagnostic ranges, enabling markdown support ([#165][])
   - passes tests on Python 3.8 on Windows ([#165][])
 
+- bug fixes
+
+  - reports files are open only after installing all handlers to avoid missing messages ([#201][])
```
krassowski: `reports files` possibly another typo. I will merge regardless.

krassowski:
Thank you.

CHANGELOG.md
bollwyvl: i guess reading it my head, it's "<version whatever> reports files are open"

krassowski: ok, makes sense!

Issue #214, opened 2020-03-12
Statusbar stuck in "Connecting..." state


krassowski:
## Description

This might be why our Linux/Mac tests are failing - the status bar which we rely on for tests to go ahead seems to have an issue...

![Screenshot from 2020-03-12 15-50-41](https://user-images.githubusercontent.com/5832902/76539940-5ea65180-6479-11ea-979c-04657320d230.png)

## Reproduce

1. Open a Python notebook
2. LSP works great, but the status is stuck at "Connecting..."

## Expected behavior

Should change to "Fully initialized"

## Context

This might be a pyls issue?

The recently failing tests (on CI) are:
![Screenshot from 2020-03-12 15-55-33](https://user-images.githubusercontent.com/5832902/76540369-f99f2b80-6479-11ea-9190-c8235f107bfd.png)

- Operating System and version: Ubuntu 19.10
- JupyterLab version: 1.2.5
- `jupyter-lsp` version: 0.8.0
- `@krassowski/jupyterlab-lsp` version: 0.8.0

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>
Paste the output from your command line running `jupyter lab` here, use `--debug` if possible.
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
Paste the output from your browser Javascript console here.
</pre>
</details>

krassowski:
They had a release in Feb, maybe we should try pinning to the previous version?

Side note: there is now a spyder fork which might be of our interest: https://github.com/spyder-ide/spyder-pyls

krassowski:
Warnings to be investigated:

![Screenshot from 2020-03-12 21-24-17](https://user-images.githubusercontent.com/5832902/76567994-e1460580-64a7-11ea-9fa8-8cc4996e137e.png)

![Screenshot from 2020-03-12 21-24-47](https://user-images.githubusercontent.com/5832902/76568033-f458d580-64a7-11ea-9221-a5b23d6e5c9e.png)

![Screenshot from 2020-03-12 21-24-47](https://user-images.githubusercontent.com/5832902/76568046-f884f300-64a7-11ea-99f5-78d65d9d92f9.png)

bollwyvl:
[...]

PR #217, opened 2020-03-13
Add readthedocs configuration, docs updates

### LINKED ISSUE #216: Publish docs on readthedocs
## Elevator Pitch

Publish the built docs on Read the Docs.

## Motivation

The current state of the docs is _meant_ to be built and published, but it just isn't yet. Even if we can't get everything they offer working, it's better than where we are: #211.

## Design Ideas

I'll do a test setup in a temporary project, get it working, and then we can make a new one that won't have all the "fix this, fix that" commits/builds, and hopefully it will work the first tiem.

Of late I've been running into more limitations when using conda there, which might limit the approach of "only document what works"... and their apt-provided nodejs (8) has been EOLed. Basically I've been using the (also old, but not compatible with our deps) nodejs from anaconda, and then doing everything with pip, but all of the r stuff is _right_ out.

So we might have to cheat, a bit, to get the data out without the servers actually _working_: definitely better to have the docs generated from the code that we test than to maintain it in both places, the whole motivation behind only checking in sources.

bollwyvl:
# [Preview](https://jlsptest.readthedocs.io)

> i'll delete that project once we're up and running

## References

- fixes #216
- might fix some of the other docs issues, but then again, maybe not

## Code changes

- [x] add `readthedocs.yml`
- [x] add `docs/rtd.yml`, a special env just for some limitations there (can't use conda-forge, basically)
- [x] add version checking of `Installation.ipynb`
- [x] add `JUPYTER_LSP_BUILDING_DOCS` environment variable to disable all of the checks for servers we won't be able to install in rtd

## Not in the PR
- [ ] set up real readthedocs project
  - don't forget the env var!
- [ ] add link to readme (once it's up)
- [ ] change in-lab link to point to readthedocs 
- [ ] decide how to manage `latest`, `stable`, etc.

## User-facing changes

- [x] updates some docs to consistently use `jupyter-lsp` (vs `jupyter_lsp`)
- [x] fix some versions
- [x] fix some package names

- none

## Chores

- [x] linted
- [x] tested (still haven't locally reproduced #214)
- [x] documented (ha!)
- [ ] changelog entry (don't think it needs it, yet)


bollwyvl:
Probably done for the day, but will take a look at the CI fails this weekend.

krassowski:
Thank you. Merging as is.

krassowski:
@bollwyvl will you be setting up the prod version, or should I take it from here? For domain, I would use jupyterlab-lsp to match the repo name.

bollwyvl:
Your call either way. It'll be a bit before I can get to it!

bollwyvl:
I see you got the docs up, hooray! I'll take a dive into our pyls issues...

PR #218, opened 2020-03-13
Remove --pre from pip install instructons

andaag:
Its no longer needed, and severely annoying when combining with pipenv (as this enables pre mode for the whole project , which can and will break other things.)

I brutally scrapped the whole template for how to contribute a patch and I think I'll let the diff speak for itself 😂. Loving the project and I'm eagerly watching the changes happening, great job so far 👍.


krassowski:
That's right. Thank you for your contribution!

Issue #200, opened 2020-02-20
Partial freeze of jupyterlab when typing "pd.read_pickle"


stephenkraemer:
First of all thanks for your great work! This is really cool stuff!

## Description

In notebooks beyond moderate size (a few hundred lines of code and minimal output), when I type "pd.read_pickle('path')", the function call is underlined in red, and jupyterlab partially freezes: i) the kernel does not respond anymore ii) the document cannot be saved anymore iii) also other kernels do not respond anymore.

## Reproduce

Unfortunately, this happens so randomly that I have no useful reproducible example. If this makes this issue less than helpful, please close it right away! I just wanted to report my current troubles in case it's interesting.

However, I have found that the error can be prevented if one types slowly enough, so that the autocompletion kicks in. Ie: if I type "pd." and then wait until the autocompletion pop-up appears, the error will not happen.

## Expected behavior

"pd.read_pickle" is correctly recognized as legitimate code and jupyterlab does not freeze

## Context

- Operating System and version: CentOS
- I am connecting to a worker node on a computing cluster via ssh tunneling - maybe that leads to a lag causing the problem?
- Browser and version: Firefox 72 and Chrome 79
- JupyterLab version: 1.1.4
- `jupyter-lsp` version: 0.7.0b0
- `@krassowski/jupyterlab-lsp` version: 0.7.1

<details><summary>Troubleshoot Output</summary>
<pre>
Paste the output from running `jupyter troubleshoot` from the command line here.
You may want to sanitize the paths in the output.
</pre>
</details>

If you think this is worth following up upon, I'd be happy to try and copy-paste log messages when the error appears next!

krassowski:
Thank you for the report and very useful insights. It sounds like a race condition...

I experienced similar hiccups with the `pyls` server in the past. My workaround is:

- `killall pyls` from console, which makes the JL work again
- save documents that need saving
- continue work without lsp features, or restart JupyterLab

I hope the workaround helps - until a better solution is available. Please let me know if it works for you.

stephenkraemer:
Thank you very much for the quick response! A workaround that allows me to continue my ongoing work when things get stuck is perfect for now, great! I'll report back if this workaround should ever fail. Thanks again!

bollwyvl:
[...]

PR #220, opened 2020-03-14
Add notes on pyls dependencies issues

krassowski:
## References

May help solve #200 and #203 for some users.

## Code changes

None

## User-facing changes

None

None

## Chores

- [ ] linted
- [ ] tested
- [x] documented
- [ ] changelog entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,13 +155,9 @@ Use of a python `virtualenv` or a conda env is also recommended.
 
    Note: it is worth visiting the repository of each server you install as
    many provide additional configuration options.
-   
+
    Note on pyls (python-language-server) issues: pyls is known to require specific
-   versions of some dependencies such as [ujson <= 1.35](
-   https://github.com/krassowski/jupyterlab-lsp/issues/203#issuecomment-599039556
-   ), [jedi == 0.15.2 and parso == 0.5.2](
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,9 +153,13 @@ Use of a python `virtualenv` or a conda env is also recommended.
    [Microsoft list](https://microsoft.github.io/language-server-protocol/implementors/servers/)
    should work after [some additional configuration](./CONTRIBUTING.md#specs).
 
-   Note: it may be worth visiting the repository of each server you install as
+   Note: it is worth visiting the repository of each server you install as
    many provide additional configuration options.
 
+   Note on pyls (python-language-server) issues: pyls is known to require specific
```
bollwyvl: :+1: would add to docs notebook as well. ideally, if there are "gotchas" in the installation, we'd add those to the schema, so they end up in the docs/ui.

krassowski: Yeah, was thinking about how to get the best of having self-contained docs AND the most frequent issues in README. Could we include README in docs as "home page" of sorts? I guess this may cause issues with links..

bollwyvl:
Yeah, we can embed it as displayed markdown and rewrite links.

PR #222, opened 2020-03-14
Update CONTRIBUTING.md - postBuild moved

krassowski:
@bollwyvl - is this ok?


bollwyvl:
Yeah, looks right.

We could make some more of these parts of docs into notebooks and actually
run them under test (and just leave them when building html) so we don't
miss these kinds of things, which are definitely important.

Since we have the binder folder, we should figure out how to ship a
workspace that would open, say, the example notebook, a "real" file or two
(of our code) and the diagnostic panel... Not in this PR, of course!

Would also love to revisit the doit pr: we're up to three js packages, and
docs, and tests, and... I do like the simplicity of (barely) documenting
"doit" and let it figure itself out.

PR #225, opened 2020-03-14
[very wip] Add Running sidebar integration

### LINKED ISSUE #205: Integrate with Running sidebar
## Elevator Pitch

Add language server status to the JupyterLab _Running_ sidebar.

## Motivation

_Running_ is the place where users will expect to be able to get information about processes running on their behalf.

## Design Ideas

- make `ILanguageServerManager`, `IConnectionManager` (or some other wrapper around it, but whatever) extend `IRunningSessions.IManager`

```ts
    // Name that is shown to the user
    name: string;
    // called when the shutdown all button is pressed
    shutdownAll(): void;
    // list the running models.
    running(): IRunningItem[];
    // Force a refresh of the running models.
    refreshRunning(): void;
    // A signal that should be emitted when the item list has changed.
    runningChanged: ISignal<any, any>;
```
- make something (probably an `ILanguageServer`) that extends `IRunningSessions.IRunningItem`

```
    // called when the running item is clicked
    open: () => void;
    // called when the shutdown button is pressed on a particular item
    shutdown: () => void;
    // LabIcon to use as the icon
    icon: () => LabIcon;
    // called to determine the label for each item
    label: () => string;
    // called to determine the `title` attribute for each item, which is revealed on hover
    labelTitle?: () => string;
```
- wire up the server status report from the backend (and refresh it occasionally, like when some new websocket occurs)
- add a `DELETE` handler for each server to actually trigger the shutdown
- handle killing the websocket and the server in the correct order
- decide on an icon approach
  - we could muddy things a bit, and drive this off the status of the websocket, a la the status bar
  - or pull the file icon for its first language/extension
  - or add an svg icon to the server spec to make it prettier... but a lot of the servers don't have branding, and the language is 
- decide on a label approach, probably `display_name`, but we could get fancy
- decide what to do when you click
  - initially, it won't do much of anything, and shouldn't block this
  - once #196 lands, redirect to whatever that is, as it's something about the actual server (rather than documents, etc) and could have links to docs, issues, etc. for that server, show the response time information, etc.

bollwyvl:
## References

- fixes #205

## Code changes
- `jupyter-lsp`
  - [x] add `DELETE` to websocket handler
  - [x] stop reader/writer _before_ stopping process
  - [x] make stop (and therefore some other things) async
- `jupyterlab-lsp`
  - [x] remove all `ignored_languages` stuff from connection manager
  - [x] remove connections from the connection cache when they close
- `lsp-ws-connection`
  - [x] make `close` more aggressive

## User-facing changes

TBD

TBD

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/connection_manager.ts
+++ b/packages/jupyterlab-lsp/src/connection_manager.ts
@@ -3,7 +3,7 @@ import { LSPConnection } from './connection';
 
 import { Signal } from '@lumino/signaling';
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
-import { sleep, until_ready } from './utils';
+import { until_ready } from './utils';
 
 // Name-only import so as to not trigger inclusion in main bundle
 import * as ConnectionModuleType from './connection';
@@ -51,12 +51,10 @@ export class DocumentConnectionManager {
     Map<VirtualDocument.id_path, VirtualDocument>
   >;
   language_server_manager: ILanguageServerManager;
-  private ignored_languages: Set<string>;
 
   constructor(options: DocumentConnectionManager.IOptions) {
     this.connections = new Map();
     this.documents = new Map();
-    this.ignored_languages = new Set();
     this.connected = new Signal(this);
     this.initialized = new Signal(this);
     this.disconnected = new Signal(this);
@@ -164,11 +162,6 @@ export class DocumentConnectionManager {
         this.forEachDocumentOfConnection(connection, virtual_document => {
           console.warn('LSP: disconnecting ' + virtual_document.id_path);
           this.closed.emit({ connection, virtual_document });
-          this.ignored_languages.add(virtual_document.language);
-
-          console.warn(
-            `Cancelling further attempts to connect ${virtual_document.id_path} and other documents for this language (no support from the server)`
-          );
         });
       } else if (error.message.indexOf('code = 1006') !== -1) {
         console.warn('LSP: Connection closed by the server ');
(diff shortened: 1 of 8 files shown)


bollwyvl:
I think i see a way to get adapters to reconnect when a connection comes back... but it's going to take a fair amount of plumbing. My current thought is to do a bit of #76: basically, the `ILanguageServerManager` will take over a lot more of the lifecycle, rather than anonymous functions defined in the extension, and the widget-specific stuff will become a consumer of those. 

```ts
// plugins/core.ts
const corePlugin: JupyterFrontEndPlugin<ILanguageServerManager> = {
  id: `${NS}:ILanguageServerManager`,
  provides: ILanguageServerManager
};

// plugins/fileeditor.ts
const fileeditorPlugin: JupyterFrontEndPlugin<IFileEditorLSP> = {
  id: `${NS}:IFileEditorLSP`,
  requires: [IEditorTracker, ILanguageServerManager],
  provides: IFileEditorLSP
};

// plugins/notebook.ts
const notebookPlugin: JupyterFrontEndPlugin<INotebookLSP> = {
  id: `${NS}:INotebookLSP`,
  requires: [INotebookTracker, ILanguageServerManager],
  provides: INotebookLSP
};

// plugins/index.ts
export default [corePlugin, fileeditorPlugin, notebookPlugin];
```

However it happens, it's going to take a bit...

bollwyvl:
This would be best applied after #278, and deciding when (or how frequently) to update sessions... closing for now.

PR #230, opened 2020-03-22
Don't use hardcoded color for fuzzy highlight, show outline when in active line

krassowski:
## References

Fixes #195. I gave up on having a different style for fuzzy matches - it does not seem to be something that others editors do, and there is no reasonable support from the servers for the distinction between fuzzy and exact highlight.

As this still uses the same colour as is used for line highlights, I added a small dotted outline when in the active line.

## User-facing changes

When using the dark theme, the fuzzy highlights (as returned by R language server) are now readable and indistinguishable from normal highlights:
Screenshot from 2020-03-22 10-33-52
Screenshot from 2020-03-22 10-33-42

There is a very subtle outline added for active line:
Screenshot from 2020-03-22 10-33-21
Screenshot from 2020-03-22 10-33-01

Same with the light theme:
Screenshot from 2020-03-22 10-32-41
Screenshot from 2020-03-22 10-32-19
Screenshot from 2020-03-22 10-32-08
Screenshot from 2020-03-22 10-32-01

## Backwards-incompatible changes

None

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

Issue #196, opened 2020-02-13
Language Server Configuration UI


bollwyvl:
## Elevator Pitch

Add a per-language-server configuration system. 

## Motivation

Many useful properties of the language servers (and their plugins) cannot be controlled with checked-in files. This has been mentioned for `pyls` in specific, but other servers (like `markdown`) are almost unusable without deeper configuration (e.g. ignored words in spell check).

## Design Ideas

Once #192 allows for fine-grained control of _which_ server you are talking to, we can store each server's config in a Big Untyped Blob in the lab settings, keyed by implementation name. You'd just edit it by hand in Advanced Settings. It probably makes sense to separate this from our (hopefully) well-typed config. So the advanced settings might look like:
```
[ ] Language Server Client    | // consult your language server for configuration options
[x] Language Servers          | {
...                           |    "pyls": {}
                              | }
```
I don't think we can have (or would even want) one per server, as I don't think there's a way to dynamically change schema at the labextension level.

Once it's available, it would "just" be a new message for the client to send to the proxy, and on to the server.

### Maybe on the initial PR?
Once getting it working the first time, we could look at using the per-server [schema](https://github.com/krassowski/jupyterlab-lsp/tree/master/py_src/jupyter_lsp/specs/config) to generate a UI: 

We may _also_ want to let such configuration (or humanely, collection of files, keyed by implementation name) to be checked into a repo (or put in a home dir) and found at runtime. This suggests a:
- a set of well-known locations, e.g. for `pyls`:
```
${PREFIX}/etc/jupyter/lsp/pyls.json
~/.jupyter/lsp/pyls.json
$(cwd)/.jupyter-lsp/pyls.json # could hide .virtual in here
```
- a `lsp/config/<language-server>` handler
  - merges configs from the search path?
- a new part of `ILanguageServerManager` that knows how to fetch it
  - merge with any stuff from jupyterlab-managed config?

krassowski:
> Once getting it working the first time, we could look at using the per-server schema to generate a UI

Yes!

> We may also want to let such configuration (or humanely, collection of files, keyed by implementation name) to be checked into a repo (or put in a home dir) and found at runtime. This suggests a:  [...]

I would be very careful about this, especially using the well-known locations... Sounds high-maintenance.

bollwyvl:
Re: config on server: for context, my motivating use case around all of
this is about aligning a team in hub+lab+lsp to get some work done:
[...]

PR #231, opened 2020-03-22
use more theme variables, start per-theme work

### LINKED ISSUE #195: Highlights use hardcoded rather than theme-responsive colors
## Description

Screenshot from 2020-02-12 03-38-07

## Reproduce

1. Open an R document (notebook or editor)
2. Click on a variable
3. Switch theme to dark
4. See unreadable text

## Expected behavior

Should see the adaptive colour of highlights.

bollwyvl:
## References

- alternative to #230
- fixes #195

## Code changes

Concretely implements some of the ideas from https://github.com/krassowski/jupyterlab-lsp/issues/195#issuecomment-586286852

## User-facing changes

> going to add a screenshot test to cycle through the full permutations of lab theme vs build-in codemirror theme

- none

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


bollwyvl:
Anecdotal screenshot:

Screenshot from 2020-03-22 14-01-50

atest/03_Notebook.robot
```diff
@@ -15,7 +15,7 @@ Python
 Foregin Extractors
     Setup Notebook    Python    Foreign extractors.ipynb
     # if mypy and pyflakes will fight over `(N|n)ame 'valid'`, just hope for the best
-    @{diagnostics} =    Create List    Failed to parse expression    ame 'valid'    Trailing whitespace is superfluous. (lintr)
+    @{diagnostics} =    Create List    Failed to parse expression    ame 'valid'    Trailing whitespace is superfluous.
```
bollwyvl: we should consider just hot-patching this to master...

atest/examples/example.py
```diff
@@ -1,6 +1,6 @@
 '''Fibonacci accumulation'''
 
-from itertools import  (accumulate, chain)
+from itertools import  (accumulate, chain); accumulate
```
bollwyvl: in retrospect, should have added a new stress test, as i don't even think we're getting any `hint`s

bollwyvl:
This is mostly working, still needs some tweaks... i made a little notebook for reviewing them, here's the meat of it:

```py
from pathlib import Path
import IPython

atest = Path.cwd() / "atest"
style = atest / "output" / "linux_37_1" / "style"
dark = style / "jupyterlab-dark"
light = style / "jupyterlab-light"

for i in sorted([*dark.glob("*.png")]):
    if i.name[0] == "0":
        continue
    print(i.name)
    IPython.display.display(IPython.display.Image(i))
    IPython.display.display(IPython.display.Image(light / i.name))
```

packages/jupyterlab-lsp/style/variables/base.css
```diff
@@ -0,0 +1,21 @@
...
+  --jp-editor-mirror-lsp-highlight-background-color: var(--jp-layout-color2);
+  /* diagnostics */
+  --jp-editor-mirror-lsp-diagnostic-decoration-style: dashed;
+  --jp-editor-mirror-lsp-diagnostic-error-decoration-color: var(
+    --jp-error-color1
+  );
+  --jp-editor-mirror-lsp-diagnostic-warning-decoration-color: var(
+    --jp-warn-color1
+  );
+  --jp-editor-mirror-lsp-diagnostic-information-decoration-color: var(
```
krassowski: Looks good! The names are getting a bit long though...

bollwyvl: thanks for taking a look. 

yeah, the length of variable names doesn't bother me too much, in this case. the `--jp-editor-mirror-` prefix isn't going anywhere... dropping `decoration` might be a thing, but frankly, although they are "canonical" i don't really like the underlines (multiple spans make look "bumpy," no real way around it), so maybe at some point there would also be `-background-` colors. or `-border-`, or `-outline-`, each of which might have other color needs: as we're seeing, one size usually doesn't fit all, when it comes to color.

atest/examples/example.py
bollwyvl: you wouldn't happen to know of anything that generates `hint`s? it's kinda hard to find in all the various things...

krassowski: Looking at the pyls code, the "refactor" messages of pylint, e.g. too few public members in a class. But these require pylint to be installed AND enabled. It used to be, but they decided to disable it by default: https://github.com/palantir/python-language-server/pull/698.

krassowski: 4 months ago

bollwyvl: i found an `info` (some of the `lintr` stuff)...

bollwyvl:
Updated screenshots all together:

https://gist.github.com/bollwyvl/2065f7f9b107eb879f293a2cdd0d1bce

base16 is just kinda brutal.

bollwyvl:
This is ready for review: of course open to some more color insights...

atest/examples/Diagnostic.ipynb
```diff
@@ -0,0 +1,67 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from itertools import  (accumulate, chain); accumulate\n",
```
krassowski: What do you think about adding comments above the lines where we expect a diagnostic to show up? Something simple, like 
```python
# expecting a warning on "; accumulate"
```
it might help to remember what is where especially if the default linters will change in the future. 

Also, I am slightly colourblind and not sure if we check the error diagnostics (I expected to see a red underline somewhere, but I do not).

bollwyvl: > adding comments above the lines

good idea!

>  slightly colourblind

this gets into the whole perceptual analysis stuff, for which there are some libraries:
- https://github.com/joergdietrich/daltonize
- https://github.com/jkulesza/peacock

...plus heavies like opencv and skimage. It's a pretty deep hole, and I don't have any ready solutions for detecting whether something is "good." from screenshots, but all the better to have your eyes on it!

> not sure if we check the error diagnostics

will verify

bollwyvl:
Did some more work to ensure different things were large and inspectable. Here's an updated screenshot notebook: 

https://gist.github.com/bollwyvl/7f6939de227c17c6ff24421668ffeeb8

i found some `information` and `hint` from various servers to embed in the notebook, but couldn't find a server that generates all of them... without extra config, a la #196. It's notably difficult to determine a consecutive `error` and `warning` on the light themes.

Issue #244, opened 2020-04-14
`yaml-language-server` Schema Support


Trajamsmith:
At present, it seems that the [`yaml-language-server`](https://github.com/redhat-developer/yaml-language-server) can only be configured using a `didChangeConfigurationRequest`. Our team would like to be able to validate YAML files against a schema using `jupyterlab-lsp`, but it doesn't seem that there's any way to send the change configuration request from the JupyterLab client.

Is that correct? Is there any plan to support configuration changes using that connection method? 

Thanks!

Trajamsmith:
Would this enhancement allow for the above: https://github.com/krassowski/jupyterlab-lsp/issues/196

krassowski:
Thank you for opening this issue.

While I do not have the bandwidth for it now (probably until June), we are [very welcoming of new contributions](https://github.com/krassowski/jupyterlab-lsp/blob/master/CONTRIBUTING.md) - if you or anyone at the team would like to try to tackle the topic of configuration. 

I think we could start with some simplest working configuration to support your use case rather than try to implement all of the intended configuration features. @bollwyvl might have some more thoughts.

bollwyvl:
Indeed, there are a number of servers that are all but useless without some level of config! Handling config should certainly be the next non-bugfix, non-chore, non-architecture, non-robustness features... though we might have to do some of those. It's a shame I haven't been able to put a bow on #205 yet, as no doubt being able to restart servers easily would be nice for testing it, but here we are.

> Would this enhancement allow for the above: #196

That is correct: we need to add the frontend plumbing to emit the correct LSP messages. I'll do a little updated vignette over there...

I, too, am spread a bit thin right now. As @krassowski says, PRs, _or contribution of any kind_, are very welcome! In this case, if you have (or can fabricate) really quite specific open source use examples that we could try to turn into tests, those would be very welcome on #196, or as links to gists.

bollwyvl:
@trajamsmith how's it been going? do you reckon 0.10.1 works now? I want to do a refresh of all our js dependencies, including language servers, soon, and was hoping we could finally put this one to bed!

trajamsmith:
Hey! Yes, the newest release of the `yaml-language-server` package should work nicely with the extension now. I haven't tested it recently, but I'll make some time to try it out. Hopefully it's as easy as a dependency bump.

PR #239, opened 2020-04-04
Integrate ICompletionItems

edzkite:
##  References

LSP extension companion changes to https://github.com/jupyterlab/jupyterlab/issues/7044

Dependent on https://github.com/jupyterlab/jupyterlab/pull/8080

## Code changes
- Update LSP fetcher to use `ICompletionItems`
- Transform Kernel reply into `ICompletionItems`
- Update `merge_replies` to combine `ICompletionItems`

## User-facing changes
- JupyterLab will now render display text vs. insert text.

- Previously:

## Backwards-incompatible changes
None that I know of.

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

PR #245, opened 2020-04-17
Language server configuration using the Advanced Settings Editor

trajamsmith:
## References

* [#196 ](https://github.com/krassowski/jupyterlab-lsp/issues/196) 
* [#244 ](https://github.com/krassowski/jupyterlab-lsp/issues/244)

## Code changes

We've added the capacity for the extension to emit `didChangeConfiguration` messages to the running language servers whenever the extension settings in the Advanced Settings Editor are changed.

Most of the updates are in the `jupyterlab-lsp` package's `connection_manager` and in the `lsp-ws-connection` package's `ws-connection`. We've added handlers to both the connection manager and the WebSocket layer for `workspace/didChangeConfiguration` messages, which are built in to the language server protocol.

## User-facing changes

Users can now pass settings from JupyterLab to the language servers using the Advanced Settings Editor.

image

After saving these settings, any open (or future) editor widgets will show updated diagnostics.

None that I'm aware of.

## Chores

- [x] linted
- [x] tested
- [x] documented
- [ ] changelog entry


bollwyvl:
Excellent stuff: will have a look for early gotchas...

bollwyvl:
Ah, i see we're _very_ wip. Good Luck!

Initial thoughts:

Under the hood, we've moved away from using the language name as a key for anything, and instead use the implementation key, mostly to support _exactly_ this case. There is no reason to expect consistency between any two servers, even for the same language. Some of them _do_ have JSON schema, but I wouldn't worry about that right now. Anyhow, you can pull the keys off the REST endpoint for server status

From the screenshot:
We'll also need to reserve somewhere to put properties for global things, specific to jupyterlab-lsp, and similarly per-language-server, we might need places to put bespoke custom stuff. We _could_ have multiple config entries, e.g. `JupyterLab LSP` and `Language Servers`, but we almost certainly don't want one _per language server_. Anyhow, first time out, we might just want to keep basically what we have, but add some nesting levels:

```yaml
{
  "language_servers": {
    "yaml-language-server": {
      "config": {
        "yaml.schemas": {"http://json.schemastore.org/composer": "*"}
      }
    }
  }
}
```

bollwyvl:
@krassowski any thoughts on the [nodejs OSX CI](https://travis-ci.org/github/krassowski/jupyterlab-lsp/jobs/676288448#L931) fail on travis? I'd be inclined to just drop the osx excursion (on a separate PR) as it's not offering us anything that isn't over on azure.

krassowski:
@bollwyvl yeah, saw this fail on #239 too. It comes down to:

> lsp-ws-connection: Assertion failed: (0), function uv_close, file ../deps/uv/src/unix/core.c, line 187.
> lsp-ws-connection: error Command failed with signal "SIGABRT".
> lsp-ws-connection: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

we can mark it as an allowed failure for now

Trajamsmith:
Yes, this was just an initial commit, just now have time to give it all a look!

I agree that having configurations per _server type_ makes more sense than generically by language, thanks for clarifying that. I'll see if I can make any headway on this today, but I might not start pushing anything up until next week.

Additionally, I'm totally unfamiliar with the acceptance testing framework you all are using. I'll try adding some tests myself at some point but would be happy for anyone to contribute one they think is appropriate 🙏

bollwyvl:
Yeah, robot is a bit unique, but it's serving our needs fairly well... and you can always drop down to selenium in python, if needs must.

I'll try to get a test up: basically, to have it meet your use case, we just need to know:
- what file we are interested in, e.g. `examples/config.yaml`
- what you should see when it's not configured, e.g. no diagnostics about schema errors
- what the new settings should be, specific to the language server
- what you should see when it _is_ configured, e.g. an error because a type is wrong

We have most of the other plumbing parts in place, so it will likely be a single file to add... I'll try to get a skeleton up, but it will likely be this weekend..

bollwyvl:
I have some (failing!) tests started against `master` to get some of the CI fixes on #247... you can upstream them already, wait until it fails and demonstrated the condition, or wait until you feel like you've got something to start testing! Happy to help further.

Trajamsmith:
Thanks Nick, that's very helpful! I'll upstream your branch and work on making the atests pass.

Trajamsmith:
So, this is still very WIP, but I _believe_ settings changes should now be making their way to the individual language servers (keyed by implementation). I have yet to verify that the servers are successfully updating though. I've gotten errors from `pyls` when the configuration keys are invalid, and silence when they're correct, but it doesn't seem like the diagnostic messages are updating to reflect the new configurations.

@bollwyvl Do you know if there're any changes that need to be made to the server extension? I feel like the running language servers should be good to go in terms of accepting those `didChangeConfiguration` messages.

bollwyvl:
Luckily the server extension is very dumb, and just trying to not trip over its own feet... though we had to take drastic action with the shadow file stuff. But nothing should need changing on this PR... as we haven't had this functionality before, I don't really know what to expect. I'll give it a whirl on binder!

[Binder](https://mybinder.org/v2/gh/Trajamsmith/jupyterlab-lsp/server-config?urlpath=lab)

`pyls` is a pretty complicated beast... does your yaml test do anything? Maybe we also need to do something per document after a reconfiguration? Would be odd...

bollwyvl:
And sterling work, by the by, @Trajamsmith: aside from a missing run of the linter (`jlpm lint`), the only broken thing is the expected-to-fail test I wrote! Good sign!

Trajamsmith:
I spent some time today investigating — I was focusing on the `yaml-language-server`, but really couldn't tell what was going on. It seemed like the server was silently crashing when the settings object wasn't correctly shaped, so debugging was very difficult. Ostensibly the settings shape _was_ correct though.

I'll keep investigating later this week. I might change focus to `pyls` since that implementation seems more robust (i.e. we get good error messages when the settings are ill-configured, and the server doesn't fail).

packages/lsp-ws-connection/src/ws-connection.ts
```diff
@@ -268,6 +268,21 @@ export class LspWsConnection extends events.EventEmitter
...
+    // TODO: define types for server configurations
+    settings: any
+  ) {
+    console.log(`Updated server configuration: ${JSON.stringify(settings)}`);
+    if (!this.isReady) {
+      return;
+    }
+
+    this.connection.sendNotification(
+      'workspace/didChangeConfiguration',
```
krassowski, line 281: Looking great!

Two quick comments:
- should we check if the server supports it with `serverCapabilities?.didChangeConfiguration`?
- some servers may be trying to follow up and query us with `workspace/configuration` (and since it is not implemented, they may fail badly)

krassowski:
Had a quick look at the Spyder implementation and found this which may be relevant to us too:

https://github.com/spyder-ide/spyder/blob/9a5f9a72fb329b931154535e70f02bf8b299cd90/spyder/plugins/completion/languageserver/client.py#L100-L105

Basically, sending the `didChangeConfiguration` notification may cause a restart of some LSP servers (and we may need to support it at the backend). I guess it will be highly server-specific (and maybe some node servers do not do that).

packages/lsp-ws-connection/src/ws-connection.ts
Trajamsmith: Thanks Michal, very good to know! Maybe that has something to do with the strange server behavior I'm running into.

I know that at least `pyls` and `yaml-language-server` support the change configuration methods, but we should definitely check for safety.

Trajamsmith:
Had some more time to plug away at this today. Everything is, I think, working for `pyls` now. I added an acceptance test to reflect that.

Next steps are to get the YAML server config working (still having issues), then maybe work on one more server implementation to make sure things are working more generally. I also still need to look into the potential failure cases @krassowski mentioned above 👆

krassowski:
Excellent work! I guess we need to add `flake8` to CI dependencies to make the tests pass.

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -24,12 +24,12 @@
...
       "type": "object",
       "default": {},
-      "properties": {
-        "settings": {
-          "title": "Language Server Settings",
-          "description": "Configuration to be sent to language server over LSP when initialized: see the specific language server's documentation for more",
-          "type": "object",
-          "default": {}
+      "patternProperties": {
+        "^.*$": {
```
bollwyvl, line 28: Good catch on the vscode weirdness.

However, the idea was to reserve the top-level namespace for things _other_ than the what's sent over the wire. For example, we might just want a client-side `disabled` flag, or something else. Perhaps move what you have here back into a `settings` field, or make it even more specific, e.g. `serverSettings`:

```json
"properties": {
  "serverSettings": {
    "title": ...,
    "description": ...
    "patternProperties": ...
  }
}
```

packages/jupyterlab-lsp/src/utils.ts
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash';
```
bollwyvl, line 1: `lodash` is really good, but I guess I'd prefer doing the piece-wise import so that we're not loading up the entire API when we just need `merge`.

Trajamsmith: Totally agree with that 👍Importing the whole library for one function is silly.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "lodash": "^4.17.15"
```
bollwyvl, line 3: If we're making this a runtime dependency, it would go down in `packages/jupyterlab-lsp`. And, per the point where imported, if we just need `merge`, there's `lodash.merge`.

packages/jupyterlab-lsp/schema/plugin.json
Trajamsmith: Gotcha! Thanks for clarifying, so rather than this:
```json
{
    "language_servers": {
        "pyls": {
            "pyls.plugins.pydocstyle.enabled": true,
            "pyls.plugins.pyflakes.enabled": false,
            "pyls.plugins.flake8.enabled": true
        }
    }
}
```

We'd prefer something like this:

```json
{
    "language_servers": {
        "pyls": {
            "settings": {
                "pyls.plugins.pydocstyle.enabled": true,
                "pyls.plugins.pyflakes.enabled": false,
                "pyls.plugins.flake8.enabled": true
            },
            "otherProps": {
                "enabled": false
            }
        }
    }
}
```

Trajamsmith: Unfortunately the settings shapes that different servers accept are somewhat idiosyncratic. Figuring out the best way to organize these JSON objects is a good upfront investment. Appreciate the help.

bollwyvl: Right... eventually. It's non-breaking if we _add_ things later (especially optional things with defaults) but moving or renaming things will definitely break peoples' configuration. For now, just claim that one key... and explicit being better than implicit, `serverSettings` or such (with those docs, which get promoted all the way to the client, would be helpful.

packages/jupyterlab-lsp/src/utils.ts
krassowski: This may not be helpful in this particular case, but in general Jupyterlab has some of the lodash-like helpers implemented in `lumino/algorithms` (formerly phosphor) see e.g. https://github.com/jupyterlab/jupyterlab/commit/41334403bb4bad7571526ce535fe56d0bb15ee5b

package.json
Trajamsmith: Apparently the preferred approach is just to add the whole library and only import the subdirectory of the method you need: https://lodash.com/per-method-packages

It really doesn't matter too much, I can write something manually if you all would prefer not to add a dependency.

bollwyvl: yeah, i had a look through the lumino/lab codebases and didn't see a merging function... the settings feature itself merges defaults and user values, but does it rather naively at the top-level key value. so i'd say go ahead and add it, then rely on tree-shaking as suggested... these things are tricky to get right, won't benefit from typescript much, and lodash is a pretty known value. If anything, it would go into JSONExt, and it would take _a while_ for that to get into lumino and then into lab.

bollwyvl: actually, `lodash.mergewith` _is_ a dependency already in lab core (at least, it appears in `jupyterlab/staging/yarn.lock`) version-matching that wouldn't add any new dependencies. 

https://lodash.com/docs/4.17.15#mergeWith

It might simplify the overall behavior...

.prettierignore
```diff
@@ -9,6 +9,7 @@
 **/lib
 **/node_modules
 **/package.json
+**/plugin.json
```
bollwyvl, line 12: motivation here? i for one like keeping as much as possibly makes sense behind a formatter...

Trajamsmith: Totally agree. I was running into errors with something, maybe the pre-commit hooks? Not sure, but I'll remove that before anything's merged.

Trajamsmith: Oh yeah, this was the issue I was having: https://github.com/nrwl/precise-commits/issues/2

Husky won't let me commit because prettier doesn't seem to want to format that JSON file.

Trajamsmith:
Alright, small update — apparently the YAML language server is a bit buggy, and that's why I've been having issues with it. I opened an [issue](https://github.com/redhat-developer/yaml-language-server/issues/256) with them and hope I can submit a PR to fix it.

_Other than that_, everything seems to be working now 🎉 I think maybe some documentation updates and another acceptance test or two would be good. 

Is there anything else you all think should happen with this? I was wondering if there's any way to dynamically get schemas for the individual servers, to validate them in the settings editor. The level of feedback for bad configurations varies a lot by server implementation 😕

packages/lsp-ws-connection/src/ws-connection.ts
Trajamsmith: So after some digging, I don't believe accepting config changes is considered a server capability in the protocol. I think we're safe to assume that the servers accept them.

I still need to look into potential `workspace/configuration` messages, though. I haven't run into any issues with Python/R/YAML, but I'll look through the protocol spec for how to safely handling those.

krassowski:
I would say merging the master in and fixing the CI issues; it currently halts on:

```
@krassowski/jupyterlab-lsp-metapackage: $ tsc -b
@krassowski/jupyterlab-lsp-metapackage: ../jupyterlab-lsp/src/utils.ts(2,23): error TS2307: Cannot find module 'lodash/mergewith'.
@krassowski/jupyterlab-lsp-metapackage: error Command failed with exit code 2.
```

Ideally also a changelog entry and an update to the README (but I can take care of that) and we would be ok to merge.

Trajamsmith:
Another update — I fixed the CI error and updated the README. I'm not really sure how you all would like the changelog entry to look, so I might leave that up to you if that's alright.

On another note, I realized that the saved settings weren't being applied to the servers on initialization. I made some changes that address the issue, but would appreciate some feedback. I'm not sure if I chose the best, most idiomatic way of doing it.

bollwyvl:
Has the pyls config test ever passed locally?


Pyls Configuration | FAIL |
Element 'css:.cm-lsp-diagnostic-Error[title="undefined name 'foo'
(flake8)"]' did not appear in 20 seconds.
------------------------------------------------------------------------------

Linux36.05

Trajamsmith:
Yes, it currently passes locally. Does it pass on your local if you pull everything down? 

I'm not sure why the CI doesn't like it. Maybe a missing package? Flake8?

bollwyvl:
> Does it pass on your local if you pull everything down?

I usually wait until there's at least some green on the board, but will take a look. It's also good to do a smoke test on binder... fired off a build there.

bollwyvl:
> I would say merging the master in

Also please do this: it merges cleanly, so I'll be running locally against that...

.prettierignore
bollwyvl: Hm... could not reproduce locally:
```bash
# after removing the line from prettierignore
$ jlpm lint
$ git add .prettierignore packages/
$ git commit -m "restore plugin.json to prettier"
husky > pre-commit (node v13.13.0)
✔  precise-commits: 1 modified file(s) found
✔  [1/1] Processing file: packages/jupyterlab-lsp/schema/plugin.json
ℹ        --> No formatting changes required in: packages/jupyterlab-lsp/schema/plugin.json
✔  Formatting complete 🎉
[server-config afa02aa] restore plugin.json to prettier
 2 files changed, 6 insertions(+), 15 deletions(-)
```

bollwyvl:
Indeed did pass locally (with `master` merged in). Go ahead and try pushing with that and see if we get anything different, I guess! 

`flake8` is indeed installed as part of the base environment in ci, and used during the unit tests... i'm having some network issues at the moment, but pulling down one of the robot output archives from azure, it might be possible to determine more from `lab.log`, which should hopefully include any crazy output from pyls. 

If that doesn't work, we can also explore:
- getting more verbose geckodriver logs in... 
  - it's quite the firehose... as in, every screenshot is duplicated again as base64. we may only want to enable it on the final run
- just increase some timeouts?
  - it's worked before, but is _extra_ brutal when it's actually broken, and goes the full distance every time

README.md
```diff
@@ -194,11 +194,64 @@ jupyter labextension install @krassowski/jupyterlab-lsp@0.7.0-rc.0
...
+        "pyls.plugins.flake8.enabled": true
+      }
+    },
+    "r-languageserver": {
+      "serverSettings": {
+        "r.lsp.debug": false,
+        "r.lsp.diagnostics": false
+      }
+    },
+    "yaml-language-server": {
```
bollwyvl, line 215: if this isn't supported yet, i'd drop it for now.

packages/jupyterlab-lsp/src/tokens.ts
```diff
@@ -8,6 +8,27 @@ export type TLanguageId = string;
 
 export type TSessionMap = Map<TLanguageServerId, SCHEMA.LanguageServerSession>;
 
+/**
+ * TODO: Should this support custom server keys?
```
bollwyvl, line 12: yeah, this is pretty much wild west stuff: we won't be able to type it, and these strings should not appear in the frontend source code. A more robust setup would likely parse the `plugin.json` schema and generate typings for it, but that's unnecessary for now.

packages/jupyterlab-lsp/src/tokens.ts
```diff
@@ -18,6 +39,15 @@ export interface ILanguageServerManager {
   statusUrl: string;
 }
 
+export interface ILanguageServerConfiguration {
+  /**
+   * The config params must be nested inside the settings keyword
+   */
+  settings: {
```
bollwyvl, line 46: is `serverSettings` now? Our python-focused backend provides everything as `snake_case`... but I wonder if we shouldn't adopt all `camelCase` for user settings, now that I'm looking at it, e.g. `languageServers`.  It looks like lab core uses all `camelCase`, which is probably reason enough to do so, for consistency.

packages/jupyterlab-lsp/src/utils.ts
```diff
@@ -130,3 +131,37 @@ export function uri_to_contents_path(child: string, parent?: string) {
...
+  return obj;
+};
+
+export const vscodeStyleSettingsParser = (settingsObject: any) => {
+  const settings: any = [];
+  for (let setting in settingsObject) {
+    const parsed = vscodeStyleSettingParser(setting, settingsObject[setting]);
+    settings.push(parsed);
+  }
+  return mergeWith({}, ...settings);
```
bollwyvl, line 166: Perhaps:
- avoid some intermediate values
- crank up the typing
with:
```ts
import { ReadonlyJSONObject, ReadonlyJSONValue } from '@lumino/coreutils';
...
/**
 * The docs for many language servers show settings in the
 * VSCode format, e.g.: "pyls.plugins.pyflakes.enabled"
 *
 * VSCode converts that dot notation to JSON behind the scenes,
 * as the language servers themselves don't accept that syntax.
 */
export function expandPath(
  path: string[],
  value: ReadonlyJSONValue
): ReadonlyJSONObject {
  const obj: any = {};

  let curr = obj;
  path.forEach((prop: string, i: any) => {
    curr[prop] = {};

    if (i === path.length - 1) {
      curr[prop] = value;
    } else {
      curr = curr[prop];
    }
  });

  return obj;
}

export function expandDottedPaths(obj: ReadonlyJSONObject): ReadonlyJSONObject {
  const settings: any = [];
  for (let key in obj) {
    const parsed = expandPath(key.split('.'), obj[key]);
    settings.push(parsed);
  }
  return mergeWith({}, ...settings);
}
```

packages/jupyterlab-lsp/src/tokens.ts
trajamsmith: This should be better named — this interface is for *after* the settings have been converted for language-server consumption (`serverSettings` -> `settings`, the required key). 

I'm all for uniform camel case if we want to switch to that. The only key here passing through Python is that "settings" key, which the user on the front-end won't see.

packages/jupyterlab-lsp/src/tokens.ts
trajamsmith: Yeah, I was thinking through how we might do that with the schema, but I think it's overkill for this PR.

Should I leave the server-name union type in just for IntelliSense hinting? It doesn't actually do any typechecking 🤷‍♂️

README.md
trajamsmith: Good catch. The Red Hat team said they're planning a release sometime in the next week, which would include my PR. Will drop it for now though.

packages/jupyterlab-lsp/src/utils.ts
trajamsmith: Those Lumino types have been throwing me off a little bit, will try this!

.prettierignore
trajamsmith: For whatever reason, I just tried it again and it worked 🤷‍♂️

trajamsmith:
I'm going to try increasing the config test timeouts, just to see if that's the problem. Will revert if not.

trajamsmith:
I don't have access to the published browser test results on Azure pipeline. If either of you all (@krassowski , @bollwyvl ) could let me know what those say, I'll keep plugging away on the CI fix!

krassowski:
Strange... I see everything in the private browsing mode... I clicked on the red cross between `Linting...` and `7a7a187 `, then Details, then `View more details on Azure Pipelines` → 11 published (artifacts). The interactive interface also seems to work without logging in. I may be missing something though...

krassowski:
The most recent failures from `7a7a187` commit was still `Element 'css:.cm-lsp-diagnostic-Error[title="undefined name 'foo' (flake8)"]' did not appear in 20 seconds.`

krassowski:
Might this discussion be relevant: https://github.com/palantir/python-language-server/issues/554#issuecomment-495563139?

krassowski:
image

Maybe we could try to add the diagnostics panel to the screenshot?

trajamsmith:
Ahh, I found it and do have access, thanks Michal! I was trying to follow the link in the publish stage log:
image
Apologies, not super familiar with Azure's CI.

trajamsmith:
So after adding the diagnostics panel to the Robot test, it's clear that the server settings _are_ getting updated, Selenium's just not seeing the tooltip element. I'm going to update the test to look for something else.

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -2,7 +2,38 @@
   "title": "Language Server",
   "description": "Language Server Protocol settings.",
   "type": "object",
-  "properties": {},
+  "properties": {
+    "language_servers": {
+      "title": "Language Server",
+      "description": "Language-server specific configuration, keyed by implementation, e.g: \n\npyls: {\n  serverSettings: {\n    pyls: {\n      plugins: {\n        pydocstyle: {\n          enabled: true\n        },\n        pyflakes: {\n          enabled: false\n        },\n        flake8: {\n          enabled: true\n        }\n      }\n    }\n  }\n}\n\nAlternatively, using VSCode's naming convention:\n\npyls: {\n  serverSettings: {\n    pyls.plugins.pydocstyle.enabled: true,\n    pyls.plugins.pyflakes.enabled: false,\n    pyls.plugins.flake8.enabled: true\n  }\n}",
```
krassowski, line 8: the VSCode's naming convention keys need to be quoted (otherwise not a valid JSON), e.g.:

`pyls.plugins.pydocstyle.enabled` → `"pyls.plugins.pydocstyle.enabled"`

It is OK in README.

atest/05_Features/Configuration.robot
```diff
@@ -0,0 +1,63 @@
...
+Documentation     Configuration of language servers
+Suite Setup       Setup Suite For Screenshots    config
+Force Tags        feature:config
+Resource          ../Keywords.robot
+
+*** Variables ***
+${CONFIG PYLS}    {"language_servers": {"pyls": {"serverSettings": {"pyls": {"plugins": {"flake8": {"enabled": true},"pyflakes": {"enabled": false}}}}}}}
+# pyflakes is enabled by default, but flake8 is not
+${PYFLAKES DIAGNOSTIC}    ${CSS DIAGNOSTIC}-Error[title="undefined name 'foo' (pyflakes)"]
+${FLAKE8 DIAGNOSTIC}    ${CSS DIAGNOSTIC}-Error[title="undefined name 'foo' (flake8)"]
```
krassowski, line 11: This is not an error by flake8 standards, should be -Warning

krassowski:
I think I found a culprit (at least potential). I got the code to run locally and noted yellow underline when only flake8 is enabled. It seems that flake8 in pyls is configured to flag everything as a "Warning":

Screenshot from 2020-05-05 21-57-39

atest/05_Features/Configuration.robot
krassowski:
```suggestion
${FLAKE8 DIAGNOSTIC}    ${CSS DIAGNOSTIC}-Warning[title="undefined name 'foo' (flake8)"]
```

krassowski: Without the suggested change the test fails locally for me, with it, it passes. It might be that you had a different version of flake8 or python-language-server installed (and I maynot have the latest, but so could be the situation with the CI).

krassowski: I tested at:
```
python-language-server==0.31.8
flake8==3.7.9
```

packages/jupyterlab-lsp/schema/plugin.json
trajamsmith: Good catch!

atest/05_Features/Configuration.robot
trajamsmith: Just updated it, we'll see what the pipeline says 🤞

trajamsmith: I have `python-language-server==0.31.10` so that might very well be it.

krassowski:
Awesome work @trajamsmith. Once the loose ends get resolved I will add changelog entry and merge.
I would probably rename the test file `07_Configuration.robot` rather than having it in `05_Features/Configuration.robot` (by the features here I meant descendants of `CodeMirrorLSPFeature`), but it is not blocking nor very important.

@bollwyvl do you have an opinion on releasing 1.0.1 and calling this PR 1.1, vs skipping over 1.0.1 and releasing just 1.1? Would you have a chance to look over my recent PRs (those which are all-green, so excluding #264 for now)

atest/05_Features/Configuration.robot
krassowski: Seems that it is not ni the pyls version (`python-language-server==0.31.10` is what Azure installs and I also upgraded locally). Let's see if it will pass this time!

trajamsmith:
Hmm, the only thing failing at this point is the Windows environment, which doesn't register any configuration changes at all. Is there anything unique about the Windows environment in the pipeline (or the way that the servers are spun up on Windows) that might cause the discrepancy?

I'll investigate, just wanted to know if you all had any ideas off the tops of your heads.

krassowski:
I do not see anything obvious, but Windows used to be ridiculously slow so you may want to try to increase the timout even more - just to be sure...

bollwyvl:
Awesome progress! Windows is always the long pole... Will investigate some
later today.

>

bollwyvl:
I'm trying this out locally in a windows VM (painful!), and saw that a number of upstreams have changed (e.g. pyls 0.32). The close/open is just to restart CI, still would love to see this get in!

trajamsmith:
I'm still very committed to getting this merged, I've just been absolutely swamped with other priorities this past week! I have a Windows machine that I've been testing locally against, and the CI issues have been hard to reproduce. 

I'm hoping to have some time this weekend to really dig into it.

bollwyvl:
You've done a lot, and we really appreciate it, but obviously "will investigate later today" turned into "y'know, some time next week" for, so I certainly understand the challenges!

And yeah, we do what we can to make local development as similar as possible to CI, but some things are hard. For example, i'm pretty sure azure has some stuff on `c:` and some stuff on `d:`... little things like that can freak python/node/conda/pip/r/etc out rather badly.

Anyhow: speaking of CI, things look pretty bad on windows on that rebuild. win38 failed the same way it had been, 36/37 are going to time out at an hour, and we won't even get logs... we probably need to do something about that. there are a few places we can put timeouts at various levels of aggression. when jobs start running that long, it gets hard to do anything until we fix them, as we "only" have 10 containers. :crying_cat_face:

bollwyvl:
Yeah, I can repro the issue that is consistently failing of not seeing the flake8 warning. 

However, if i then do _anything_ to the file editor (like hover anything), then it updates. I think we can be okay with that. From a test perspective, it might just need something like 
```robotframework
Pyls Configuration
  # ...
  Click Element    css:.lm-FileEditor .CodeMirror .some-token
  Wait Until Page Contains Element    ${FLAKE8 DIAGNOSTIC}    timeout=60s
  # ...
```

Even if that's not robust, as i'd mentioned before, I'd happily accept closing the file and re-opening it (either in the ui, or via messages), if that's what it takes.

Unfortunately burned a lot of time getting to this point, so I'll probably  have to also have to hope to be able to pick up further investigation this weekend...

bollwyvl:
I got a thing to work semi-reliably in windows... will hot patch...

atest/07_Configuration.robot
```diff
@@ -0,0 +1,71 @@
...
+    Drag and Drop By Offset    ${JLAB XP DOCK TAB}\[contains(., 'Diagnostics Panel')]    600    -200
+    Ensure Sidebar Is Closed
+    Capture Page Screenshot    01-diagnostics-and-settings.png
+    # Diagnostic panel should show pyflakes diagnostics, but no flake8
+    Wait Until Page Contains Element    ${PYFLAKES DIAGNOSTIC}    timeout=60s
+    Page Should Not Contain    ${FLAKE8 DIAGNOSTIC}
+    Set Editor Content    ${CONFIG PYLS}    ${CSS USER SETTINGS}
+    Click Element    css:button[title\='Save User Settings']
+    Drag and Drop By Offset    ${tab}    0    100
+    Lab Command    Save Python File
```
bollwyvl, line 36: This apparently creates enough traffic for the language server to "wake up"...

atest/07_Configuration.robot
```diff
@@ -0,0 +1,71 @@
...
+    Open Diagnostics Panel
+    Drag and Drop By Offset    ${JLAB XP DOCK TAB}\[contains(., 'Diagnostics Panel')]    600    -200
+    Ensure Sidebar Is Closed
+    Capture Page Screenshot    01-diagnostics-and-settings.png
+    # Diagnostic panel should show pyflakes diagnostics, but no flake8
+    Wait Until Page Contains Element    ${PYFLAKES DIAGNOSTIC}    timeout=60s
+    Page Should Not Contain    ${FLAKE8 DIAGNOSTIC}
+    Set Editor Content    ${CONFIG PYLS}    ${CSS USER SETTINGS}
+    Click Element    css:button[title\='Save User Settings']
+    Drag and Drop By Offset    ${tab}    0    100
```
bollwyvl, line 35: This is not really doing anything, but was having issues getting the document to become active (it doesn't have any buttons, etc.)

atest/07_Configuration.robot
trajamsmith: Huh, very interesting that it needs this. I don't think I was experiencing this on my Windows machine at all 🤔

bollwyvl:
Holy smokes:
Screenshot from 2020-05-16 15-16-10

Don't know why travis didn't run...

atest/07_Configuration.robot
bollwyvl: my windows machine, like azure's windows machine, is very slow.

trajamsmith:
Big thanks @bollwyvl for getting to the bottom of that!

What more needs to be done for the merge on this, do you think? I'm happy to keep plugging away if need be. It looks like RedHat released a new YAML language server with my PR, so we could test for the YAML schema change now too.

bollwyvl:
I think getting your yaml use case squared would be good. It should just be
a matter of:

jlpm add --dev yaml-language-server@x.y
z

To update the yarn.lock

And then off you go...

Anyhow, take the WIP off, and we'll do some thorough review

package.json
```diff
@@ -12,16 +11,16 @@
...
     "tslint-plugin-prettier": "^2.0.1",
     "tslint-react": "^4.0.0",
-    "tslint": "^5.15.0",
     "typescript": "~3.7.2",
     "unified-language-server": "^0.3.0",
     "vscode-css-languageserver-bin": "^1.4.0",
     "vscode-html-languageserver-bin": "^1.4.0",
     "vscode-json-languageserver-bin": "^1.0.1",
-    "yaml-language-server": "^0.4.0"
+    "yaml-language-server": "^0.8.0"
```
trajamsmith, line 23: This update seems to have broken the YAML editor acceptance test 😓 Looking into it.

bollwyvl:
Sorry this is such a bear, but hooray regression testing!

Our endgame is to keep all of the servers we know about happy, which I
guess would include tests for configuring all of the other servers, but I
think "two is enough" to demonstrate that the feature is robust. It will
also help reveal the pattern we'll want to use when we scale up, as the
file should increase in length by at most three or four lines per server,
not 20.

Another thing we could be testing (and would be appropriate on this pr) is
using a mix of the dotted syntax along with the protocol-spec tree...
Presumably they should work together?

trajamsmith:
The dot and nested objects already work together, I'll add a test to demonstrate it.

I started to look into tests for all of the different language servers, but some of them actually don't have very obvious settings to test (though I admittedly didn't spend a whole lot of time on this). I think adding more config tests would be a great future target, and I'd love to keep contributing to this project down the line!

krassowski:
> I started to look into tests for all of the different language servers, but some of them actually don't have very obvious settings to test (though I admittedly didn't spend a whole lot of time on this). I think adding more config tests would be a great future target, and I'd love to keep contributing to this project down the line!

I would be happy with just testing settings two or three representative servers for now (e.g. R/Python/something typescript-based) - and the work here is already a great with that respect, but it would be great to have tests for more in future PRs!

bollwyvl:
Right, don't spin your wheels too hard on servers you aren't using!

trajamsmith:
*Update:* it looks like there's a really pernicious parsing issue with the YAML language server. After updating to the new version, `0.4.0 -> 0.8.0`, the diagnostic positions are completely off. Because the positions are nonsensical, the diagnostics never appear in our text editor and we can't look for them in the acceptance test.

I'm very sure this is on their end, since nothing's changed on ours. I've been trying to track it down in their source, logging a thousand things from their server, but it's been a slow process. I'll probably open an issue/PR with them today or tomorrow. Just wanted to keep you all updated.

bollwyvl:
You're a super trooper to keep fighting there... i had to get a patch or two in, myself, for some config thing. The community there is at least responsive. If you feel confident it will solve itself once it's fixed on that end, we could move forward with getting this in, but I'm not in a particular hurry, and we'd probably still want one other server tested... there is precedent for "Works For Pyls" features getting merge that ended up not being robust to the quirks of other servers... it's quirks all the way down. Your call!

trajamsmith:
No hurry on my end either — I'd really like the YAML config working robustly for my use case, so I'm happy to dig in. I think an additional test for the R language server would probably be ideal! 

Have some more time set aside tomorrow to look at this stuff. I'll try and get an R acceptance test in.

trajamsmith:
_Just an update (in light of the discussions around moving this repository):_
All of my limited OSS time lately has been going towards the yaml-language-server stuff. The issue I've been working on is pretty deep in the language parser, and it isn't limited to our implementation either (diagnostic positions are also broken in VSCode, for all 3.6 million installs).

Please let me know if priorities change around this PR! Getting the YAML server working better is a direct dependency for this, IMO, but if you all think it's a lower priority use-case, that's probably fair too.

bollwyvl:
@trajamsmith vicious yak shave: cheers to you for sticking with it. I'll personally use the heck out of it (conda junk, azure, github actions, travis), so I'm rooting for you.

The config feature is definitely important, and would be good to land: I was hoping we had two language servers tested, but as you point out, few of them are as configured as much as `pyls`. A _really_ clutch usecase for me is markdown editing support. The current one we test with, [unified-language-server](https://github.com/unifiedjs/unified-language-server) _does_ have [some options](https://github.com/unifiedjs/unified-language-server/blob/master/CONFIGURATION.md), so maybe I can whip up a test that fiddles with those settings...

package.json
bollwyvl: Do you have a version you can suggest? Should I just roll back to whatever we had on `master`?

bollwyvl: that'd be [`0.4.1`](https://github.com/krassowski/jupyterlab-lsp/blob/master/yarn.lock#L12448)

bollwyvl:
@trajamsmith at your leisure, see what you think of: https://github.com/trajamsmith/jupyterlab-lsp/pull/1

trajamsmith:
I had to pin `robotframework-lint <1.1` in the CI lint stage because I was running into compatibility issues with `robotframework < 3.2`. Everything passed with that change, but let me know if it's inadvisable for any reason!

trajamsmith:
That said, I'd be happy for us to wrap up this PR if you all would like to do any final review. I'll keep working with the `yaml-language-server` and let you all know when it's ready to be upgraded (likely in a small future PR).

Issue #250, opened 2020-04-20
404 GET /lsp/status?


robna:
## Description

I have installed jupyterlab-lsp extension, following the guideline here in the 
README.md. Although no problems with installation, I can't use it in jupyter lab. I think this might be similar to #203, however, I don't have `pyls` installed, and I can't remove `build_config.json` as it was recommended from the discourse channel linked there, because `/usr/share/jupyter/lab/settings/` is empty (not sure if this is related to it). The other jupyter lab extensions that I have installed work.
![Screenshot-20200420233922-243x423](https://user-images.githubusercontent.com/26489593/79802412-706bf600-834f-11ea-883b-5d82682b0bfc.png)

## Reproduce

1. pip install jupyter-lsp
2. jupyter labextension install @krassowski/jupyterlab-lsp
3. pip install python-language-server[all]
4. jupyter lab

Diagnose:
  I don't have any functionality of jupyterlab-lsp available in my notebooks. It 
is however listed as enabled in the jupyter lab extensions.
In the terminal running the server I get:
`[W 19:41:44.712 LabApp] 404 GET /lsp/status?1587404503700 (127.0.0.1) 57.23ms 
referer=http://localhost:8888/`

## Expected behavior

Run jupyter lab with working lsp functions

## Context

- Operating System and version: Linux-5.4.33-3-MANJARO-x86_64
- Browser and version: Firefox 75.0
- JupyterLab version: 2.1.0-1
- `jupyter-lsp` version: 0.8.0
- `@krassowski/jupyterlab-lsp` version: 1.0.0

<details><summary>Troubleshoot Output</summary>
<pre>
jupyter troubleshoot
$PATH:
     /home/me/.local/bin
     /usr/local/bin
     /usr/bin
     /bin
     /usr/local/sbin
     /usr/lib/jvm/default/bin
     /usr/bin/site_perl
     /usr/bin/vendor_perl
     /usr/bin/core_perl
     /var/lib/snapd/snap/bin
     /home/me/bin

sys.path:
     /usr/bin
     /usr/lib/python38.zip
     /usr/lib/python3.8
     /usr/lib/python3.8/lib-dynload
     /home/me/.local/lib/python3.8/site-packages
     /usr/lib/python3.8/site-packages

sys.executable:
     /usr/bin/python3

sys.version:
     3.8.2 (default, Apr  8 2020, 14:31:25)
     [GCC 9.3.0]

platform.platform():
[...]

PR #257, opened 2020-04-27
Add jupyter serverextension enable --py jupyter_lsp to install instructions

krassowski:
## References

See final comments in #250. I think that we talked about it earlier, but it turns out to be needed for quite a few users, so let's have it here. Not sure if this should be commented by default and marked as optional, or just kept uncommented - I do not see any harm in running this command again if extension is already enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Use of a python `virtualenv` or a conda env is also recommended.
 
    ```bash
    pip install jupyter-lsp
+   jupyter serverextension enable --py jupyter_lsp
    ```
 
 1. install `nodejs`


README.md
```diff
@@ -108,6 +108,7 @@ Use of a python `virtualenv` or a conda env is also recommended.
 
    ```bash
    pip install jupyter-lsp
+   jupyter serverextension enable --py jupyter_lsp
```
bollwyvl, line 111: So this gets into the whole configuration ball of wax, which the `jupyter_notebook_config.d` approach is trying to avoid. The bothersome thing is in both of our CI excursions, we install from the `whl` distributions, which is what most people will pull, and yet this still persists. But, yeah, i guess if it ain't workin'... 

As written, this is going to add or change a config file in `~/.jupyter` (or wherever the heck it goes in windows), and make it try to load for every jupyter notebook derived thing you load, at which point it will create errors. So especially if you have folk that `sudo` everything if it doesn't work the first time, you're going to get stuff owned by `root` in your `~`, which rapidly leads to more madness.

If anything, i guess we'd want to suggest:
```bash
jupyter serverextension enable --py --sys-prefix jupyter_lsp
```

and/or a pre/post-flight of:

```bash
jupyter serverextension list
```

krassowski:
Cross-referencing:
- https://github.com/krassowski/jupyterlab-lsp/issues/203#issuecomment-601320933
- https://github.com/krassowski/jupyterlab-lsp/issues/203#issuecomment-623103653

thakkarparth007:
@krassowski , please merge this. I would've saved two day's efforts if I'd known this earlier :)

krassowski:
It hopefully should not be a problem anymore with new `jupyter_server` required for 3.0.

PR #261, opened 2020-04-28
Fix caret icon in diagnostics panel (port to JupyterLab 2.0) #258

### LINKED ISSUE #258: Sort icon is missing in the diagnostic panel table
An artefact from porting to JL 2.0 

- JupyterLab version: 2.1

krassowski:
## References

Sort caret icons were not ported in JupyterLab 1.2 → 2.0 transition, this PR makes them work again. Fixes #258. I used upstream changes from https://github.com/jupyterlab/jupyterlab/commit/b30f505f5f701e3b61b76555d31a3c8e47bc0817 as a reference.

Some of the CSS was re-used from core `listingHeaderItem` stylesheet.

## Code changes

Use `caretDownIcon` and `caretUpIcon` icons from JupyterLab.

## User-facing changes

Icons are now right-aligned.

fixed-header

None

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [x] changelog entry

## Future work

Test it somehow? Probably we will not forget next time as we explicitly import the icons now.


bollwyvl:
Sorry to have missed a) this initially and b) the request for review. Will look over the code, but the gif raises a gut feel I had earlier, where we should be reserving enough space for an icon so that changing the sort doesn't change the width of the columns. I don't know if there is a "blank" icon, but we could just pick one and set it's visibility to hidden or opacity or something...

packages/jupyterlab-lsp/src/adapters/codemirror/features/diagnostics_listing.tsx
```diff
@@ -161,19 +162,18 @@ class Column {
...
-            (props.listing.sort_direction === 1 ? 'lsp-descending' : '')
-          : ''
-      }
+      className={is_sort_key ? 'lsp-sorted-header' : null}
     >
       {props.name}
-      {is_sort_key ? <span className={'lsp-caret'} /> : null}
+      {is_sort_key ? (
+        <sortIcon.react tag="span" className="lsp-sort-icon" />
+      ) : null}
```
bollwyvl, line 176: re: other comment, instead of `null`, return sorticon, but with a `className` that hides it... or, have something like:
```css
.lsp-the-table-class th:not(.lsp-sorted-header) .lsp-sort-icon {
  opacity: 0;
}

.lsp-the-table-class th:not(.lsp-sorted-header):hover .lsp-sort-icon {
  opacity: 0.25;
}
```

I guess making it work with the keyboard would be a task for another day...

bollwyvl: kinda like #269 i guess :laughing:

krassowski:
Note to self: maybe worth to add an info about the new icon from #269 to the changelog

bollwyvl:
Some thoughts I had while digging through this: 

`Diagnostic Panel` is a bit verbose, as well, it's always a panel (like everything else is, i guess). So we could probably drop "panel" altogether. Also, as it currently only shows one document at a time, and now has an icon, it could update with _which_ document you're looking at... this could be helpful if you keep it open and jump between different documents. So either `[ico] Diagnostics <document>` or `[ico] <document> Diagnostics` or the even more terse `[ico] <document>`, which would actually follow the pattern throughout the rest of the documentation.

Also, since the icon is actually _in_ the dom, we can mess about with it with css. For example, it could be colored to the highest level present in that document. I haven't seen anybody else doing that kind of stuff, and `error` isn't exposed as an [icon color](https://github.com/jupyterlab/jupyterlab/blob/master/packages/ui-components/style/icons.css), but we could figure something out.

krassowski:
I see `[ico] <document> [dD]iagnostics` as a worthy UX improvement (in a subsequent PR?). Then it is a singleton, so maybe `[ico] Diagnostics: <document>`? No strong opinions here, but I would say that it is distinct from the use-cases of notebook/editor/git diff (which all can have multiple tabs for specific documents) to warrant the inclusion of "Diagnostics" word and a bit different title pattern.

I would be happy with accepting the icon colour feature, but I am not sure if this is something many users would see as an important feature (and what would be the maintenance burden). We could have an issue with a proposal for that and let it hang there for some time to see if it gets any active support from the users (unless you want to have a go at it right now).

bollwyvl:
We can push all of it to The Future... just noting stuff as i thought of
it.

PR #264, opened 2020-04-29
Fix unwanted path auto-completion, allow user-invoked completion in strings/comments

### LINKED ISSUE #241: Auto-complete issue related to file paths
## Description

When using jupyterlab-lsp, the file path autocomplete in str stops working. Also, sometimes it will automatically add useless .ipynb_checkpoints.

## Reproduce

Issue 1:
1. type `'/'` (including quote)  in any notebook's code cell 
2. then press TAB after `/`.
3. Nothing happens.

Issue 2:
1. Go to any cell, type in the following code:
```
from scipy.io import loadmat
data = loadmat('/path/to/some/.mat/file')[0]
```
2. type `.` after `[0]`
3. `ipynb_checkpoints` will be appended after `.` symbol.

## Expected behavior

Issue 1:
A list of files/folders should pop out for selection.

Issue 2:
Nothing should happen or a valid completion.

## Context

- Operating System and version: Windows 10 Pro x64 Version 1909
- Browser and version: Chrome 80.0.3987.149
- JupyterLab version: 2.0.1
- `jupyter-lsp` version: 0.8.0
- `@krassowski/jupyterlab-lsp` version: 1.0.0

krassowski:
## References

Fixes #241 

## Code changes

`with_trigger_kind` replaced with `set_trigger_kind_for_next_invocation` as the with the idea did not play well here (it seems that even after converting the callback to await it was only waiting for the invocation, but not for completion due to how commands work); either way it is to be removed as setting `trigger_kind` on an instance is silly and error-prone and should not be done in the first place (but I did not want to re-implement `completer:invoke-notebook` and `completer:invoke-file`); now it seems that a thin wrapper/re-implementation/PR upstream which would allow passing the `trigger_kind` when executing the command is the best way out (also described with a TODO comment), but this may be out of the scope of this PR.

Maybe, just maybe, a custom implementation of a command, let's name it `completer:trigger` would allow getting rid of the current workaround (but then the value from this PR is in adding a test-case).

## TODO

- [ ] play around to check if we can implement `completer:trigger` easily, and in the process get rid of two workarounds:
    - `trigger_kind` on an instance,
   - adding the dummy completer item described below
- [ ] or - if the above cannot be accomplished within two weeks - open an issue for that.

## User-facing changes

- Auto-completion is not performed if LSP does not return results and Kernel returns only one result (which is usually `.ipynb-checkpoints` for Python); to workaround how `completer:invoke-notebook/file` work (auto-completing if there is one suggestion), I add a second dummy (but harmless) suggestion. This is not the final solution, just a workaround.
- Completion in strings and comments can now be manually invoked

None, but will likely need manual merge with #239

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/atest/05_Features/Completion.robot
+++ b/atest/05_Features/Completion.robot
@@ -1,7 +1,7 @@
 *** Settings ***
 Suite Setup       Setup Suite For Screenshots    completion
-Test Setup        Setup Notebook    Python    Completion.ipynb
-Test Teardown     Clean Up After Working With File    Completion.ipynb
+Test Setup        Setup Notebook    Python    Completion.ipynb    subdirectory=isolated_subdirectory
+Test Teardown     Clean Up After Working With File    Completion.ipynb    subdirectory=isolated_subdirectory
 Force Tags        feature:completion
 Resource          ../Keywords.robot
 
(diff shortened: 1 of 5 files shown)


krassowski:
On hold until I get a feedback on https://github.com/jupyterlab/jupyterlab/issues/8406

krassowski:
Superseded by https://github.com/jupyterlab/jupyterlab/pull/8375. Maybe some tests could be rescued from there, but not a priority ATM.

PR #269, opened 2020-05-06
some tweaks to css, diagnostic icon for #258

bollwyvl:
## References

- Builds on #258

## Code changes

- picks up some more stuff from lab styling (padding, etc)

## User-facing changes

- consistent padding for headers, contents
- icon is always present, even if not shown (ghosts in on `:hover` indicating which sort direction you _would_ get) so columns don't change width (so rows don't re-flow, changing height)
- adds icon for diagnostic panel (we can play with the colors, if desired)

awful hacked up gif:

ezgif-7-cc2e1c439214

n/a

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


krassowski:
Hot takes:

- +1 for the panel icon - looks great! Where does it come from? Does it have a licence?
- I was thinking about adding the margin to prevent jittering but was refraining from it as it increases the width of 4-5 columns and the panel is short on space anyways...
    - but I guess it is a UX improvement and we have an option to hide columns and most of us won't do any serious IDE stuff on a smartphone (and most phones now come with HD), so let's not design around a niche use-case
- I like the hover idea too (though it is absent upstream)
   - was previously considering adding title attribute on th stating "descending/ascending" (not sure if needed though...)

I will check it out locally (would like to test in bot FF and Chrome) and come back later (by Friday evening).

bollwyvl:
>  width of 4-5 columns

Maybe collapse cell/column/row into a single column? What would sorting on column even buy a user?

krassowski:
Works fine for me! What about the icon licence?

krassowski:
> Maybe collapse cell/column/row into a single column? What would sorting on column even buy a user?

I see a use-case of sorting by cell and row (line), maybe both at the same time (multi-key sort).

I agree that sorting by column (character) is mostly useless (maybe unless it is multi-key, i.e. row and column). But let's start merging and maybe address it in a later PR!

krassowski:
Ok, it seems the stethoscope icon is [SIL OFL licensed](https://iconify.design/icon-sets/mdi/stethoscope.html) so we can easily include it.

We just need to include the copyright notice and license info. I would put it in the icons directory as `icons/OFL_LICENSE.md` (license) and a copyright notice in `icons/README.md` (copyright notice) - I think it should suffice as we can put it [basically anywhere](https://opensource.stackexchange.com/questions/7035/where-include-sil-open-font-license-in-a-website-if-i-use-font-face).

bollwyvl:
Oops, sorry i dropped the ball on this. Hopefully the license stuff looks ok: I've tried to leave it open for easily adding later. Going to take a look at visually squashing row/col...

bollwyvl:
Looks ok.

Screenshot from 2020-05-23 15-33-16

That's the narrowest that particular one can be... Unfortunately, I don't know how to make it cleanly line break on underscores... any smaller, and....

Screenshot from 2020-05-23 15-34-27

... i'm okay with that for now, but we could add a `overflow: auto` or something...

bollwyvl:
This is targeting your branch, not master... we can either merge it into yours, and go from there, or retarget this... should work, though i haven't done any thorough testing of it...

krassowski:
I'm going to merge into my branch which should trigger CI (and then will look into merging to master on a later date, maybe tomorrow)

Issue #254, opened 2020-04-26
/docs/LANGUAGESERVERS.md missing


jhgoebbert:
Hi, 
thank you for this great extension.

When you click on the status bar where one can see the language servers, this includes a Link called"Documentation: Language Server".  

This link is dead, as it points to  
https://github.com/krassowski/jupyterlab-lsp/blob/master/docs/LANGUAGESERVERS.md  
which does not exist.

Best regards,
Jens Henrik

krassowski:
Hi Jens,

Thanks for highlighting this. I think that we should point to https://jupyterlab-lsp.readthedocs.io/en/latest/Language%20Servers.html

Issue #275, opened 2020-05-31
Jump to definition is not working


kiranik:
## Jump to definition is not working

Jump to definition option is not working for modules and packages.
When we use 'Jump to definition' option for an imported module or a package, nothing happens. It shows the following error in JS console.
```No jump targets found
Uncaught (in promise) TypeError: Cannot destructure property 'uri' of 'n' as it is undefined.
```
## Reproduce

1. Create a file named abcd.py with a class named sample
2. open a notebook
3. Import class sample from abcd
4. try to jump to the definition of sample class by selecting from menu or using Command+B

## Expected behavior

Open definition in a separate editor window.
## Context

- JupyterLab version:2.0.0
- python version - 3.7.4
- jupyterlab-lsp version - 1.0.0
- python-language-server - 0.30.0

<details><summary>Required: installed server extensions</summary>
<pre>
 jupyter_lsp  enabled 
    - Validating...
      jupyter_lsp 0.8.0 OK
 jupyterlab  enabled 
    - Validating...
      jupyterlab 2.0.0 OK
</pre>
</details>

<details><summary>Required: installed lab extensions</summary>
<pre>
JupyterLab v2.0.0
Known labextensions:
   app dir: <path>
        @krassowski/jupyterlab-lsp v1.0.0  enabled  OK
</pre>
</details>

<details><summary>Browser Output</summary>
<pre>
```No jump targets found
Uncaught (in promise) TypeError: Cannot destructure property 'uri' of 'n' as it is undefined.```
</pre>
</details>

krassowski:
Thank you for reporting this. I tried to reproduce this in two separate JupyterLab installations but it worked well in both cases for me: 

![it-is-working-jump-to-def-in-lsp](https://user-images.githubusercontent.com/5832902/83970702-01326d00-a8cf-11ea-9904-4e2d9630b8eb.gif)

Confirmed that it is working using
- JupyterLab v2.1.4,
- @krassowski/jupyterlab-lsp v1.0.0,
- jupyter-lsp==0.8.0
- python-language-server==0.33.1

as well as:
- JupyterLab v2.1.3,
- @krassowski/jupyterlab-lsp v1.0.0,
- jupyter-lsp==0.8.0
- python-language-server==0.31.9

I note that you use quite older versions of JupterLab and of the python-language-server. I would recommend you to update, as there were multiple changes in the JupyterLab since. Please feel free to reopen if you can reproduce the problem after updating.

Issue #221, opened 2020-03-14
Title tag for docs includes "0" instead of a release number


krassowski:
The docs title tag says "Installation - Jupyter[Lab] Language Server 0 documentation". This is because the release is set to `"0"`. There are two easy solutions:

- set it to `""` and ignore the issue (maybe sub-optimal for the UX)
- set it to the current `jupyter-lsp` version, like this: https://github.com/krassowski/jupyterlab-lsp/commit/cffd5ce582faeb80ecc8b4ea3a8534d2525c9eef

Any preferences/better ideas?

krassowski:
By the way, should we decide that the major version of `jupyterlab-lsp` and `jupyter-lsp` should always match?

krassowski:
Outdated.

PR #276, opened 2020-06-07
Update the recommended JupyterLab version

### LINKED ISSUE #289: Installation docs show two diff versions of jupyterlab
The README shows this for prerequisites:

image

and this for the installation commands:

image

Could you advise which it is? And then I am happy to make a PR that changes it

krassowski:
## References

Follow-up after #275. Closes #289.

## Code changes

- source of truth for JupyterLab version is now taken from packages, not from binder environment
- `versions.py` added to the root to ensure versions consistency across the repo
- Installation notebook now uses code cells with modified markdown magic to display the variables

## User-facing changes

None

None

## Chores

- [x] linted
- [x] documented


bollwyvl:
sorry the version appears in so many places, looks like just readme and the `lab.txt` are 
out of sync.

:+1: to requiring lab 2.1, though!

krassowski:
Would you think that converting the markdown cell in `docs/Installation.ipynb` to a code cell returning a markdown object, then hiding the code and only displaying the result would be possible/advisable? This would allow to having the versions from variables...

bollwyvl:
Yeah, we could parse `job.test.yml` and pull out the "broadest" spec. Though looks like right now they are all set to `>2,<3.0.0a0`.

I think we have some other examples of using `nbsphinx: source_hidden` in some of the other notebooks... it's buried down the the _metadata -> advanced_ sidebar in cell metadata.

krassowski:
I think I found an elegant solution: https://github.com/spatialaudio/nbsphinx/issues/81#issuecomment-660561769 :)

krassowski:
So this still WIP, but 43a08ce is POC that it works. Just need to bump the JupyterLab version in the npm package and decide where to put the version and markdown cell code (which currently is in the notebook, but it certainly should go _somewhere_).

krassowski:
I added `versions.py` in the root of the repo to track versions more reliably, with a single source of truth across the repo. CI now passes - apart from Windows being Windows.

bollwyvl:
Now that all this (almost) works: I thought there were things (related to completion #239?) we definitely wanted from Lab 2.2.0... how might we validate that the features in question landed?

krassowski:
This is a good question. I asked in the https://github.com/jupyterlab/jupyterlab/issues/8080. There was also https://github.com/jupyterlab/jupyterlab/pull/8375, none of which is in the final [changelog](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#v2-2-0). But then, the 2.2 release was a bit unusual...

bollwyvl:
Thanks for the refs! Looks like both of them shipped:
- https://github.com/jupyterlab/jupyterlab/blob/v2.2.0/packages/completer/src/widget.ts#L422-L428
- https://github.com/jupyterlab/jupyterlab/blob/v2.2.0/packages/completer-extension/src/index.ts#L96-L99

bollwyvl:
Hooray green CI! It seems like merging now, and doing the release dance, would make sense as 2.2.0 is not widely distributed. Then do the next release with the hard 2.2.0 dependency.

krassowski:
Looks like it - thanks for checking! So my current plan is to:
- merge this PR
- fix #254 
- maybe fix #281
- #221: close or use versions.py to set version in the title
- release a final major version targetting JupyterLab 2.1 **on Monday**
- branch off
- star working on JupyterLab 2.2 completion (merge #239)
- merge #278

Not sure where #288 would fit.

bollwyvl:
> merge #278

I'm super excited that you are interested in #278: we've been demoing with it on https://github.com/deathbeds/_fam/tree/07-11-2020 and it's been working pretty well. I've also been tinkering with a JEP to formalize the schema it uses.

> Not sure where #288 would fit.

Let's not hold up anything for #288: it is, however, pretty important to some other stuff I'm working on (and getting paid for) so it _will_ get finished, at some point. Presently, I'm still having issues getting it built on OSX, which is very hard to debug from a distance, but it won't actually have that big of an impact once it lands (would just be a small dot release of the server piece). _Really_ getting it to work, though, will require a dedicated extension:
- with a full bundled pdf.js, a la [jupyterlab-latex](https://github.com/jupyterlab/jupyterlab-latex)
- that can handle its [custom LSP extensions](https://texlab.netlify.app/docs/reference/custom-messages) :crying_cat_face:

Issue #184, opened 2020-01-28
Multiple sources of LSP messages on frontend and backend


bollwyvl:
## Elevator Pitch

Enable things other chains than...
```
Feature ⇄ LSPConnection ⇄ WebSocket ⇄ Handler ⇄ Session ⇄ Stdio ⇄ Language Server 
```
...to complete LSP requests. 

## Motivation

We've done a lot of work to make the first roundtrip work. However, there are a number of sources of messages that would provide more flexibility and robustness, and in some cases end-user simplicity.

### the frontend language server
```
Feature ⇄ FrontendLSPConnection
```
- what: many of the node-based language servers can actually be webpacked (with appropriate shims)
- why: would let us minimize the install pain for a large swath of servers/users
- how: create a `FrontendLSPConnection extends ILSPConnection` 
  - each server would then be a `packages/jupyterlab-language-server-xyz`, with enough boilerplate to wrap/patch its guts (e.g. `fs` over contents REST API)

### the tcp language server
```
Feature ⇄ LSPConnection ⇄ WebSocket ⇄ Handler ⇄ Session ⇄ Tcp ⇄ Language Server 
```

- what: in playing with live development of a language server in-loop with `pygls`, i really wanted to just pop open a tcp port and and add it to my language servers... but we only support `stdio`
- why: some things are better over tcp. will probably need this eventually anyway for sourcegraph, docker-hosted things, etc.
- how: create a `TcpReader` and `Writer`, add `mode` to the spec, allow a `{{ port }}` pass-through in the args

### the kernel language server
```
Feature ⇄ KernelLSPConnection ⇄ WebSocket ⇄ Kernel 
```

- what: similar to the above, and as demonstrated over on `lintotype`, we can define a comm and let kernels stuff LSP into it
- why: the kernels know a lot about... things. like crazy `%` syntax, and where _real_ files are (except for the notebook you're working on :rofl: )
- how: add a `KernelLSPConnection implements ILSPConnection`

### the static language server
```
Feature ⇄ LSIFConnection ← DocRegistry ← REST ← ContentsManager
```

- what: [LSIF](https://microsoft.github.io/language-server-protocol/specifications/lsif/0.4.0/specification/) defines a spec for a JSON-based dump of a language server in a workspace
- why: something like [jyve](https://deathbeds.github.io/jyve/lab/) or a lab-forward nbconvert would let you keep all the LSP features (hover, symbols, etc) in a static published setting
[...]

PR #278, opened 2020-06-07
[wip] Add Language Server Kernel

### LINKED ISSUE #268: Language Server Kernel
## Elevator Pitch

Write a wrapper kernel that can act as the proxy between a(ll) language server(s), replacing custom websocket connection wrangling.

## Motivation
Extending on #184, (and thinking about #237 which i have been able to reproduce when binder is getting thrashed), and some broader comments in the community... Lab 2 reworked sessions stuff so it is/tries to be even _more_ robust in the face of degraded connections. Trying to get (and maintain) that level of functionality will be hard for us.

Maximizing reuse of the kernel infrastructure would:
- get users all the buffering/reconnection/starting/stopping basically for free
- give us access to [better tooling](https://github.com/vidartf/jupyterlab-kernelspy) for kernel messages than we'll be able to build for our custom websocket transport
- obviate the need for some of the backend complexity (at least in the manager)
  - though we'd need _new_ complexity to handle dealing with "plain" kernels for the listeners, etc... unless all of that functionality could be moved into the kernel itself
- obviate the need for the clunky (and huge) webpack build of `lsp-ws-connection`

## Design Ideas

The `LanguageServerKernel` would be an `ipykernel.kernelbase.Kernel` subclass. It would (awkwardly) show up in the kernel listing (though we could we hide it with css... but it might be cool to interact with it, don't know).

Since we'd _still_ need the serverextension enabled (somehow #257), it could enable the kernelspec pretty directly, rather than needing another file, as something like `ilsp`.

The frontend would start a `KernelSession` for `ilsp`. When it came back with its `kernel_info`, it might be able to give us all of the server listing stuff, and we'd be able to start talking to it. Presumably the starting/stopping of language servers could be some pithy JSON-schema-constrained language, rather than being "full" python.

Much of the existing stdio machinery would be reused, but would have basically zero chance of hanging the `notebook` server process, which is A Good Thing. The listener API would be... more complicated to maintain, but it can't all be roses.

As for actually integrating with the kernel messaging protocol: [custom comms](https://jupyter-notebook.readthedocs.io/en/stable/comms.html), used by widgets and vdom, already give us a way to add this kind of feature in a way that could be used by other kernels in the future, provided they had some way to "opt in" to receiving LSP messages.

bollwyvl:
[Binder Screenshot from 2020-06-07 18-28-01
](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-language-server-kernel?urlpath=lab%2Ftree%2Fexamples%2FPython.ipynb)

## References

- fixes #268

## Code changes

A lot. Too much. Enough that this might be unmergeable, as-is and will bitrot _really fast_. Here's the various yak shaves i had to go down:

- [x] add a language server kernel
  - [x] now owns the `LanguageServerManager`
  - [x] `jupyter.lsp.control` comm target for status, etc. (formerly REST API)
    - [x] also handles PageConfig
  - [x] `jupyter.lsp.language_server` comm target for JSON-RPC (formerly WebSockets)
- [x] remove serverextension
- [x] remove `lsp-ws-connection` dependency and code
  - [x] removes `event-emitter` dependency
- [x] remove runtime `vscode-*` dependencies
  - [x] add type-only import of `vscode-language-server-protocol` (duplicate the one `enum`)
    - [x] upgrade typescript
      - [x] upgrade prettier
- [ ] add a mostly-faithful reimplementation of `connection` and `lsp-ws-connection` 
  - [ ] i might move this down one layer, so that we can #184 better
  - [ ] add a LSP-over-Comm
    - [x] add a JSON-RPC-over-Comm
    - [x] add a single LSP type entrypoint, use throughout (e.g. `lsProtocol` -> `LSP`)
  - [ ] some of the levels still aren't right (e.g. features shouldn't know about `comm`)

## User-facing changes

- [x] "finishing" installation would require:
  - [x] `jupyter lsp kernelspec install` instead of
  - [x] `jupyter serverextension enable`
- [x] A `Language Server` kernel will show up in the `Notebooks` and `Console` section
 - [ ] either we need to
   - [ ] hide it
   - [ ] brand it 
- [x] a single Language Server Kernel is reconnected/started when lab starts (visible in `Running`)
  - [x] everything waits until it has received the initial status and URIs

At once, total, as well as curiously backwards compatible. I really didn't touch the UI at all, aside from changing imports, and making a few more things asynchronous. But yeah, if anybody was working on a non-open fork and had been doing fancy stuff with our non-according-to-anybody's-spec websockets or REST API, they're hosed.

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry

## Thoughts

I'm actually pretty happy with how this is coming together. In the long run, the Kernel Message Spec will be more stable over time than our custom stuff. 

While there's a few singletons right now, I can imagine a kernel being able to register the comm targets, and our manager would be able to ask them if they'd like to contribute.

I'll be punching on this a bit more, but relatively soon, we'll have to decide a path forward: while there are a lot of commits, I don't know if i have the patience to rip this back apart into easily-reviewed pieces. 

- just take it into `master`, try a release against `2.1`
  - kinda scary, and I don't want the burden to be on open PRs to get it working
- start a `next` branch and target that
  - roll it out at some point in the future (e.g. post Lab 2.2)
  - handle incumbent PRs more piecewise 

Stay tuned!


bollwyvl:
Ha HA, green CI, I knew you where in there someplace.

Looking into some needed housekeeping that might drive the variability down.

bollwyvl:
Language Kernel restart is handled now... i've been manually testing with `kill` (humorously, but satifyingly through the lab terminal) but need to add it to the acceptance tests.

I'm considerably more convinced that this is the right way to go, now!

bollwyvl:
It wasn't _too_ bad merging in `master` with config, which is heartening, so I guess I can keep working on this... looking forward to #279 to make the diff more manageable. 

I've also unified the two comm targets (any lsp target can now return status). This will be helpful when describing it as a new JEP (also in progress, but too wip to show: the yak shave is endless)

bollwyvl:
Nice. merging `master` after #279 knocked 2k off the total `+` and `-` for this PR.

bollwyvl:
Nice. Removing the docs from azure removed the remaining pytest warning.

packages/jupyterlab-lsp/src/comm/connection.ts
```diff
@@ -0,0 +1,556 @@
...
+  sendOpen(documentInfo: ILSPConnection.IDocumentInfo) {
+    this.notify(LSP.Method.DID_OPEN, {
+      textDocument: {
+        uri: documentInfo.uri,
+        languageId: documentInfo.languageId,
+        text: documentInfo.text,
+        version: documentInfo.version
+      }
+    }).catch(err => console.warn(err));
+    this._openDocuments.set(documentInfo.uri, true);
```
bollwyvl, line 302: @krassowski adding this (and checking it in the various `change` spots squashes the pyls error about something missing in `self._docs`... apparently something was sending change events before the document was fully open.

krassowski:
I'm thinking about cherry-picking the improvements from `packages/jupyterlab-lsp/src/lsp.ts` and promoting it to protocol `protocol` package; I would then rework the current ws-connection to implement `ILSPConnection` as defined here and use the new protocol bits.

bollwyvl:
Take whatever you feel will help... I was hoping to get to a
fully-specified autogenerated definition of the protocol for our purposes,
but that's still a ways off, and the typing is already more concise in this
approach.

Follow-on would be some way to do dynamic message registration for e.g the
latex build/search, but having the full spec first is priority! This will
make adding new features _much_ easier, as the new feature package(s) would
already have the "boring" parts ready, and can get straight to the UI.

bollwyvl:
I **do** intend to get back to this, but it will be some time before other stuff settles down, I'll wager. I can also mentor anybody that might interested in picking this back up.

Some thoughts I had captured elsewhere, but is relevant here:
- This is still worth doing
  - As it is, it would free use from `nbclassic` vs `notebook` vs `jupyter_server` issues
    - we would still have to deal with them in test, but from an API perspective, not really (just the labextension file structure)
    - being a kernel has other nice benefits
      - you get a top-row launcher icon
        - at some point we need to support LSP Workspaces, maybe that's what an `Untitled.lsp-workspace.ipynb` is? 
  - Having an interactive kernel-based proxy would mean one could use the visual debugger 
    -  i've had good luck demoing the current Lab-LSP implementation along with the debugger, as they are running on two entirely separate channels/loops
    - being able to pause/step an LSP connection :kissing_heart: :ok_hand:      

Some I _think_ we need for this work to be successful/maintainable, long-term:
  - have a single source of type/data validation truth for (versioned) LSP messages in kernel-level comms, kernel-side listeners, etc. 
    - i still think JSON schema is the right lowest-common denominator
      - it's fine if it's _generated_ from typescript source... getting the enums right is probably the trickiest part

krassowski:
I am concerned about two things:
- performance
- changes in the kernel specs

The first one we can measure the other we don't have much influence on.

krassowski:
But I absolutely did not mean to close this one, there is a lot of valuable work in here :)

PR #279, opened 2020-06-10
Upgrade dev tools [squash]

bollwyvl:
## References

- things noticed while working on #245 and #278

## Code changes

- brings mostly more in line with lab core
  - a few rules we should look into fixing, but are fine for now: some might be false positives, just wanted to get it working
```
    'jest/valid-expect': 'off',
    'jest/no-test-callback': 'off',
    '@typescript-eslint/ban-ts-ignore': 'off',
    '@typescript-eslint/interface-name-prefix': 'off',
    'react/display-name': 'off',
    'prefer-spread': 'off',
    '@typescript-eslint/triple-slash-reference': 'off',
    'no-async-promise-executor': 'off',
``` 
- upgrade to robotframework 3.2, robotframework-lint 1.1
  - big deal here is the `robot.tidy` will respect `...`, so it's possible to control lines better
- typescript 3.9
  - we'll be wanting this for explicit type-only imports
- prettier 2.0 (needed to parse ts 3.9)

## User-facing changes

n/a

n/a

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


packages/.eslintrc.js
```diff
@@ -0,0 +1,69 @@
...
+    'no-useless-escape': 'off',
+    'prefer-const': 'off',
+    'jest/no-jest-import': 'off',
+    'jest/no-export': 'warn',
+    'jest/no-try-expect': 'warn',
+    'jest/expect-expect': 'off',
+    // deviations from jupyterlab, should probably be fixed
+    'jest/valid-expect': 'off',
+    'jest/no-test-callback': 'off',
+    '@typescript-eslint/ban-ts-ignore': 'off',
```
krassowski, line 57: this is listed twice (above with warn), could we just have it to warn? Not sure how much of ignores we can remove easily atm, but these include upstream typing errors which need to be reported

packages/.eslintrc.js
```diff
@@ -0,0 +1,69 @@
...
+    '**/_*.ts',
+    '**/_*.d.ts',
+    '**/typings/**/*.d.ts',
+  ],
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: 'packages/tsconfig.eslint.json',
+  },
+  plugins: ['@typescript-eslint', 'jest'],
+  rules: {
```
krassowski, line 30: `@typescript-eslint/interface-name-prefix` could be added (there was some controversy to whether it should be in defaults, I think it is gonna stas but lab explicitly lists it: https://github.com/jupyterlab/jupyterlab/blob/9ce3026ddf7653ebd2b0c4703853743d18fffe62/.eslintrc.js#L25-L28)

packages/jupyterlab-go-to-definition/jest.config.js
```diff
@@ -6,7 +6,7 @@ const reuseFromUpstream = [
   'moduleNameMapper',
   'reporters',
   'setupFiles',
-  'setupFilesAfterEnv'
+  'setupFilesAfterEnv',
```
krassowski, line 9: @bollwyvl wait a minute! JupyterLab does not enforce comma-dangle... Where is that coming from? Like literally, which eslint rule enables it?

I will not oppose it vehemently, but certainly need to understand the rules better...

krassowski: Another thought: if we decide to have the comma-dangle, could you potentially split it into two PRs, one with comma-dangle and one with everything else? It is quite long diff and I am afraid I would miss something otherwise.

bollwyvl: yeah, the commas are annoying. i think that's a prettier 2 thing, which core has _not_ adopted, but is definitely needed to support some ts 3.8+ syntax. We can turn them off, see if that quiets the diff down.

bollwyvl:
With unconfigured `trailingComma`: +1,169 −875

With `trailingComma: none`: +712 −417 

Seems much more reasonable.

packages/.eslintrc.js
bollwyvl: :+1: will add and fix

bollwyvl: actually, this breaks

jupyterlab-lsp/packages/jupyterlab-lsp/src/virtual/editor.ts
```
export interface VirtualEditor extends CodeMirror.Editor {}
```

renaming that breaks things all over, so just ignoring that instance.

packages/.eslintrc.js
bollwyvl: meh... always having 15 warnings fills almost a whole terminal:
```
/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/adapters/codemirror/feature.ts
  166:3  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/adapters/jupyterlab/components/free_tooltip.ts
  25:1  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore
  36:3  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore
  39:5  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
  509:5  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/extractors/regexp.ts
  45:9   warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore
  52:11  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/lsp.ts
  45:5  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/magics/maps.ts
  13:9  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

/home/weg/projects/lsp/jupyterlab-lsp/packages/jupyterlab-lsp/src/virtual/editors/notebook.ts
   14:1  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore
  188:5  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore
  190:5  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore
  272:5  warning  Do not use "// @ts-ignore" comments because they suppress compilation errors  @typescript-eslint/ban-ts-ignore

✖ 13 problems (0 errors, 13 warnings)


```

bollwyvl: ignore the ignores in these cases

bollwyvl:
I'll put back the `arrowParens` default: as with the trailing commas, i prefer `always`, but agree it can be a separate PR to evaluate the merits of this in favor of getting it in.

bollwyvl:
`+615 −304`, in a much-more-reasonable range to evaluate off the initial

packages/jupyterlab-lsp/src/adapters/jupyterlab/components/statusbar.tsx
```diff
@@ -215,7 +215,8 @@ class LSPPopup extends VDomRenderer<LSPStatus.Model> {
             href={
               'https://github.com/krassowski/jupyterlab-lsp/blob/master/docs/LANGUAGESERVERS.md'
             }
-            target={'_blank'}
+            target="_blank"
+            rel="noreferrer"
```
bollwyvl, line 219: this was an actual security fix, which seemed reasonable.

packages/jupyterlab-lsp/src/virtual/editor.ts
```diff
@@ -253,5 +253,4 @@ export abstract class VirtualEditor implements CodeMirror.Editor {
   }
 }
 
-// tslint:disable-next-line:interface-name
-export interface VirtualEditor extends CodeMirror.Editor {}
+export interface VirtualEditor extends CodeMirror.Editor {} // eslint-disable-line @typescript-eslint/interface-name-prefix
```
bollwyvl, line 256: will change this to be `next-line`

bollwyvl: bda0f67

bollwyvl:
alright, putting it down so CI can do its thing... also will updated title to reflect that it should be squashed so that the actual delta shows up in history, not all the intermediate cruft.

bollwyvl:
Hooray green.

>

krassowski:
Merging the settings editor produced merge conflicts here. I am thinking about a release as we are to ship the settings editor and then mrging this, let me know what you think.

bollwyvl:
oops, i'll take a look at fixing it up right now... i think the one
security fix might make it worth it.

bollwyvl:
It looks like our `yarn.lock` is getting pretty outdated, and is reflecting some fairly old resolutions... I think we can release with it as-is (as it doesn't affect behavior as tested at runtime), but we'll probably want to blow it away as an order of business after the release...

bollwyvl:
ok, linter's back to happy. i'm bullish on the rest going through. :crossed_fingers:

bollwyvl:
Although I did find one warning (max number of robot tests in one suite)... have bumped it up, as the `Editor` suite isn't going to be getting any smaller, and splitting it up wouldn't make a lot of sense.

bollwyvl:
Woof. we have to turn down the 3 minute timeout waiting for the splash screen. I'm rejiggering the timeouts: we're probably better off failing faster, and hitting stuff on the retry than just spinning for that long. of course this is mostly a problem on windows...

bollwyvl:
ah, much better. I actually don't see _any_ splash fails, which is a good sign.

krassowski:
Yet I managed to forget to squash. My apologies.

FYI I just enabled building of RTD on PRs (see https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html), elt's see if it is useful.

bollwyvl:
The RTD thing sounds great... over on #278 I've started reworking some docs to reflect the proposed changes, so should be easier to review, at least on the theoretical/user-facing level. The code changes remain rather... extensive. Merging in this PR and re-linting will certainly help on that front!

bollwyvl:
also, in light of the RTD thing, we can probably get rid of the azure docs job.

Issue #130, opened 2020-01-05
Enable strictNullChecks


krassowski:
To keep up with JupyterLab typescript settings we probably want to utilize `strictNullChecks`.

PR #300, opened 2020-07-27
stop using throw

bollwyvl:
## References

- attempted fix for part of #299
- ~~may investigate adding #130~~

## Code changes

- [x] removes all custom `throws`, instead returning `null` (most of these were already sometimes returning `null`)

### not on this PR
- [ ] ~~take this opportunity to enable `strictNullChecks` and clean up some more return types~~

## User-facing changes

- should not result in failed mousemove handler calls

- still looking into this... behavior was probably just unspecified

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [x] changelog entry


bollwyvl:
I had a look at #130 I think it's too much to do here. Indeed, we will likely want to go package-by-package rather than all in one go.

bollwyvl:
I'm not really sure how to test this properly... it seems the plumbing underneath the other extension is rather heavyweight, and doesn't have an existing binder.

docs/Installation.ipynb
```diff
@@ -8,15 +8,27 @@
    },
    "outputs": [],
    "source": [
-    "from markdown import markdown\n",
     "import sys\n",
-    "sys.path.insert(0, '..')\n",
+    "\n",
+    "from markdown import markdown\n",
+    "\n",
+    "sys.path.insert(0, \"..\")"
```
krassowski, line 15: Good one!

bollwyvl:
Hooray!

PR #301, opened 2020-07-29
Support JupyterLab 2.2

### LINKED ISSUE #241: Auto-complete issue related to file paths

krassowski:
## References

Supersedes #239, fixes #241.

## Code changes

- adopted to `ICompletionItem` interface
- implements `ICompletionItemsConnector` over simple `DataConnector`
- uses kernel icon for suggestions from kernel if type is not available
- passes documentation and deprecation note to the completer

## User-facing changes

In addition to the new shiny completer widget (see all three PRs for https://github.com/jupyterlab/jupyterlab/issues/7983):
- the autocomplete is no longer overaggressive (https://github.com/jupyterlab/jupyterlab/pull/8375) and does not cause path-completion issues - this opens way to implementing hitlander mode
   - this means that pressing dot `.` once does not cause autocompletion to insert text if only one item was found. Pressing tab twice does the trick.
- we get `label` rather than `insertText` displayed

competer-opt

`ICompletionItemsConnector` is a breaking change and not backwards compatible with JupyterLab 2.1

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
Note: disable pre-commit hook for now as it was destroying code and could not work around it.

@bollwyvl I am prettier illiterate; it is breaking `completions.ts` (literally does mish-mash) when run and I have no idea how to fix it, nor which piece it does not like. How do I lint/fix this?

bollwyvl:
You can disable it explicitly for a single file in `.prettierignore`. Might be sufficient for this case. There are also some pragmas we can enable, which would let you put it magic comments. But it'd be best if we could fix through normal means, no doubt.

Anyhow, can you post an example of before/after what it's doing as a gist? Maybe we need to upgrade? Please, let's not downgrade, as that would ruin my world on #278, which I'm already scared of fixing the bitrot on.

But indeed, who lints the linters! I disable the hooks, too, because I don't like node doing anything i don't type out with `node`, `yarn` or `jlpm`... `git` is for me.

packages/jupyterlab-lsp/src/adapters/jupyterlab/components/completion.ts
```diff
@@ -271,111 +254,87 @@ export class LSPConnector extends DataConnector<
...
+  private transform_reply(
+    reply: CompletionHandler.IReply
+  ): CompletionHandler.ICompletionItemsReply {
+    console.log('[LSP][Completer] Transforming kernel reply:', reply);
+    const items = new Array<CompletionHandler.ICompletionItem>();
+    const metadata = reply.metadata || {};
+    const types = metadata._jupyter_types_experimental as JSONArray;
+
+    if (types) {
+      types.forEach((item: JSONObject) => {
```
bollwyvl, line 270: maybe hoist some of these lambdas to the class, or standalone functions, up so they don't get created every time, and just use `map` rather than `forEach`?

yarn.lock
```diff
@@ -2,708 +2,855 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
-  version "7.8.3"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
-  integrity sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4":
```
bollwyvl, line 5: it may well be time to do a full re-solve of all our deps. We should probably do some better sanity pins on, for example, the node-based language servers as well.

krassowski:
Also, why didn't Azure CI fire?

krassowski:
Well, seems that running `jlpm prettier` from command-line worked better than from a hook...

yarn.lock
krassowski: surprisingly, seems to be largely working :) the most important trick was in https://github.com/krassowski/jupyterlab-lsp/pull/301/commits/bf52857f3f64bf74a8022111cf0a4856db3d54b5

krassowski:
All completer tests now pass locally! CI (https://dev.azure.com/krassowskimichal/jupyterlab-lsp/_build) has issues with backend lint:

```
source activate jupyterlab-lsp && python scripts/lint.py
========================== Starting Command Output ===========================
/bin/bash --noprofile --norc /home/vsts/work/_temp/3d27c3e5-264b-4ba5-8197-7ce2012605f5.sh
/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.7/site-packages/setuptools/distutils_patch.py:26: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  "Distutils was imported before Setuptools. This usage is discouraged "
All done! ✨ 🍰 ✨
52 files left unchanged.
py_src/jupyter_lsp/tests/test_paths.py:35: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"
py_src/jupyter_lsp/tests/test_paths.py:61: error: Argument 1 to "__call__" of "_SkipifMarkDecorator" has incompatible type "int"; expected "Union[str, bool]"
Found 2 errors in 1 file (checked 39 source files)
```

which seems like a good mypy catch because `~WIN` would mean:

```
>>> ~True
-2
>>> ~False
-1
```

so we should go for `not WIN`. Is this correct @bollwyvl?

bollwyvl:
Yeah, `not WIN` looks correct.

krassowski:
Hooray, azure kicked in after merging master this time!
Not sure if I have yarn lock right, there were conflicts and I just resolved it anew after merging... May be missing something - let's see how it plays out on CI

bollwyvl:
I've done some digging on that error code before... Sometimes it's enough
to  ensure the thing being selected by this keyword is visible, _and_
unique... Will dig into it this weekend, might have to rebuild my windows
VM.

krassowski:
Now the question is why
- MacOSXThreeSeven and
- WindowsThreeSeven, WindowsThreeEight

fail while MacOSXThreeSix, WindowsThreeSix, MacOSXThreeEight and all Linux tests pass. I see no pattern...

krassowski:
Well synced, your message appeared 0.5 second before I pressed enter ;)

bollwyvl:
Also, the win 36 is kind of "best effort" because of ignoring some previous
pyls issues... So it's not really an indicator for a lot of features. Will
revisit...

krassowski:
Now archspec (whatever this is) also fails to conda install on Windows...

bollwyvl:
On windows: cannot reproduce the archspec thing (not familiar with it), but can consistently reproduce the `rect` error, so probably will be able to fix it... or at least determine why we should allow it ignoring it.

atest/05_Features/Completion.robot
```diff
@@ -108,6 +130,7 @@ Cell Editor Should Equal
 Select Completer Suggestion
     [Arguments]    ${text}
     ${suggestion} =    Set Variable    css:.jp-Completer-item[data-value="${text}"]
+    Wait Until Element Is Visible   ${suggestion}  timeout=10s
```
bollwyvl, line 133: Hopefully this works! :crossed_fingers:

ci/steps.conda.yml
```diff
@@ -1,7 +1,7 @@
 parameters:
   name: Linux
   packages: ''
-  install_cmd: conda install -yn base -c conda-forge conda python-libarchive-c
+  install_cmd: conda install -yn base conda
```
krassowski, line 4: I guess this is an attempt to tackle the `archspec` issue?

bollwyvl: yes, the thinking is get something as _vanilla_ as possible, especially for windows, while we get this sorted. who knows what state it's in when they cook the docker images? We could also pin harder, e.g. `4.8` or something.

if we want to start getting far off vanilla, we could investigate installing `mamba` in this step instead of upgrading `conda`... no doubt it could shave off some minutes, at the price of sometimes giving different solves. definitely a separate PR, though!

bollwyvl: i also need to investigate caching (or packaging) the tectonic stuff.

krassowski: Was just thinking about mamba [while reading this](https://twitter.com/tangming2005/status/1292814278467223553)

bollwyvl: I'll open an issue...

bollwyvl: https://github.com/krassowski/jupyterlab-lsp/issues/311

PR #304, opened 2020-07-29
Delete LANGUAGESERVERS.md

krassowski:
This file has been there to redirect users hitting it from an old version of the JupyterLab extension (from the link from a statusbar), which is no longer supported and the file has been there for over 5 months. It's time to remove it.

Also, this is a test of Azure Pipelines (will the webhook trigger a new build?).


bollwyvl:
He's on fire!

PR #305, opened 2020-07-29
Fix tests/mypy error for "not WIN" condition

krassowski:
Attempts to address https://github.com/krassowski/jupyterlab-lsp/pull/301#issuecomment-665735786, which can be also seen in #304

### DIFF
--- a/py_src/jupyter_lsp/tests/test_paths.py
+++ b/py_src/jupyter_lsp/tests/test_paths.py
@@ -32,7 +32,7 @@ def test_normalize_posix_path_home_subdir(
     assert normalized_uri(root_dir) == expected_root_uri
 
 
-@pytest.mark.skipif(~WIN, reason="can't test Windows paths on POSIX")
+@pytest.mark.skipif(not WIN, reason="can't test Windows paths on POSIX")
 @pytest.mark.parametrize(
     "root_dir, expected_root_uri",
     [
@@ -58,7 +58,7 @@ def test_file_uri_to_path_posix(file_uri, expected_posix_path):  # pragma: no co
     assert file_uri_to_path(file_uri) == expected_posix_path
 
 
-@pytest.mark.skipif(~WIN, reason="can't test Windows paths on POSIX")
+@pytest.mark.skipif(not WIN, reason="can't test Windows paths on POSIX")
 @pytest.mark.parametrize(
     "file_uri, expected_windows_path",
     [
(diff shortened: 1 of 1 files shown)


krassowski:
Well, the implications are that we were NOT unit testing paths on Windows before...

krassowski:
We have the following errors:
```python
>       assert normalized_uri(root_dir) == expected_root_uri
E       AssertionError: assert 'file://vboxsvr/shared-folder/' == 'file://vboxsvr/shared-folder'
E         - file://vboxsvr/shared-folder
E         + file://vboxsvr/shared-folder/
E         ?                             +
```


```python
>       assert file_uri_to_path(file_uri) == expected_windows_path
E       AssertionError: assert 'C:/Windows/System32/Drivers/etc' == 'C:\\Windows\\System32\\Drivers\\etc'
E         - C:\Windows\System32\Drivers\etc
E         ?   ^       ^        ^       ^
E         + C:/Windows/System32/Drivers/etc
E         ?   ^       ^        ^       ^
```

```python
>       assert file_uri_to_path(file_uri) == expected_windows_path
E       AssertionError: assert 'C:/some dir/some file.txt' == 'C:\\some dir\\some file.txt'
E         - C:\some dir\some file.txt
E         ?   ^        ^
E         + C:/some dir/some file.txt
E         ?   ^        ^
```

Not quite sure if it's our expectations that are incorrect, or if the implementation is wrong.

bollwyvl:
Indeed... i need to take a look back through what prompted us to have to take these measures. I look forward to next month when we can drop 3.5, and use pathlib for many of these things.

bollwyvl:
Sorry I didn't get to responding to this until now!

I think if the implementation was wrong, we'd be seeing more failures downstream (e.g. `Foreign extractors.ipynb`). I _guess_ we could also force CI to do stuff on a different drive (there's a `D:` and `C:` on azure), and muck up the path with more strings.

Given the _intent_ of `normalized_uri`, it's safe to just add the missing `/` for `test_normalize_windows_path_case`. Everywhere we use URIs, now, I'm pretty sure we're using a proper URI manipulation function, so whatever it's doing is right.

Turns out the slashes in `test_file_uri_to_path_windows` are _also_ correct, as we're not doing anything path-related with them, just URL stuff, and so `/` should be expected. So those slashes can just be reversed in the expected, and the `r` taken off the front.

So yeah: just bad expectations, and boo on me for trying to be cute with `~`.

bollwyvl:
Hooray!

Issue #311, opened 2020-08-11
Continuous Integration Performance and Robustness Brainstorming


bollwyvl:
## Elevator Pitch

Let's gather some ideas for how we might improve the value we're getting from CI.

## Motivation

CI is not _really_ that broken, but it's pretty slow. There might be some other tools and services we can look into which wouldn't require altering the _code_ that much, so can be done in parallel, and incrementally, while feature work is on-going.

## Design Ideas

Post/emote ideas below.

bollwyvl:
### Migrate to GitHub Actions

It seems to be where things are going, and 2x simultaneous jobs than Azure Pipelines.

bollwyvl:
### Investigate mamba

Using `mamba` instead of upgrading `conda` will potentially save us some time when solving environments and downloading files. It can sometimes return different solutions (but so can different versions of `conda`).

bollwyvl:
### Investigate conda-lock

Instead of even _solving_ the environments in CI, we could generate them quickly, offline, and check them into the ci subfolder. Stacks with `mamba` (which it will use preferentially).

bollwyvl:
### Revisit caching

Being able to cache:
- conda lock solutions
- the built lab
- tectonic cache
- conda package cache
- yarn package cache

...in roughly that order, would knock some minutes off all of the runs, but particularly on windows which seems to have really bad IO. Github Actions and Azure both have more advanced (but of course, incompatible) caching actions. They all work best if we have things _checked in_ that get hashed. The entropy on the built lab is pretty challenging to describe.

bollwyvl:
### Investigate dodo

`dodo` has been _very_ useful for tying together some large builds, and normalizing ci vs local development. #183 has a good amount of work in it, but can be improved substantially. A challenge: `pathlib` is pretty important, so supporting 3.5 might be a challenge (`pathlib2` isn't _quite_ the same).

krassowski:
Some test cases fail on timeouts (especially on Windows):

```
Windows38.04 Interface.Statusbar                                              
==============================================================================
Statusbar Popup Opens                                                 | FAIL |
Setup failed:
Element 'css:.jp-mod-accept.jp-mod-warn' did not appear in 5 seconds.
```

or:

```
Windows38.04 Interface.Statusbar                                              
==============================================================================
Statusbar Popup Opens                                                 | FAIL |
Element 'css:div.lsp-statusbar-item' did not get text 'Fully initialized' in 1 minute.
```

```
[...]

PR #312, opened 2020-08-11
try mamba, remove docs job in azure

bollwyvl:
## References

- attempt at https://github.com/krassowski/jupyterlab-lsp/issues/311#issuecomment-671661109

## Code changes

- updates and uses `mamba` instead of `conda`
- removes docs jobs (because RTD)

## User-facing changes

- n/a

- n/a

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
  - not sure if we'll want to also _recommend_ this, since we _use_ it? 
- [ ] changelog entry


bollwyvl:
| before | after |
|----|----|
| Screenshot_2020-08-10 Pipelines - Run 20200811 2 logs  | Screenshot_2020-08-10 Pipelines - Run 20200811 3 logs

bollwyvl:
Hah, windows still failing, will try pinning conda (looks line conda-forge maybe jumped the gun with 4.8.4)

bollwyvl:
| before | after |
|------|-------|
| Screenshot_2020-08-10 Pipelines - Run 20200811 2 logs(2) | Screenshot_2020-08-10 Pipelines - Run 20200811 3 logs(1) |

yeah, so looks like a minute across the board (still-running job flaked).  Not bad! I've tried pinning conda, and telling it not to update incidental packages, will see if that offsets the additional expense of mamba and friends (or makes it worse).

bollwyvl:
Still fail on windows, but got further. interesting that our huge solve actually went slower, but there's so many factors...

| before | after |
|-|-|
| Screenshot_2020-08-11 Pipelines - Run 20200811 2 logs | Screenshot_2020-08-11 Pipelines - Run 20200811 4 logs |

bollwyvl:
The tea laves (win/36) suggest we're saving a couple minutes on windows, which makes sense. So we can either:
- merge this now, just rely on those gains for future stuff
- see how some other approaches (like conda-lock) work, in isolation (as it would become less reliant on runtime solves)

| before | after |
|-|-|
| Screenshot_2020-08-11 Pipelines - Run 20200811 2 logs(1) | Screenshot_2020-08-11 Pipelines - Run 20200811 5 logs

krassowski:
Yes, this seems like a step in the right direction. Feel free to merge or explore other options! I do have high hopes for GitHub Actions.

bollwyvl:
Welp, it's easy to back out, and making subsequent experimentation _faster_ is just, well, _faster_!

Issue #25, opened 2019-09-07
System settings


krassowski:
An issue to write down ideas or requests for what options should be included.

As a starting point I imagine that the JSON structure will hold 'general' settings in one key and language-specific overrides (empty by default) in all the following keys.

The settings will be further subdivided by the LSP function. Possibly like that:
```js
{
  "general": {
    "inspections": {},
    "autocompletion": {
        "auto-invoke": true,
        "invoke-after-any": false 
    },
    "highlight-on-cursor": {}
  },
  "python": {
     "inspections": {
        "ignore": ["E222"]
     },
    // specific functions can be disabled by the user by setting false
    // (or maybe 'disable'?)
    "highlight-on-cursor": false
  }
}
```

Issue #7, opened 2019-08-21
An error occurred when installing


simon-ritchie:
A UnicodeDecodeError occurred when installing with the following command:

```
$ jupyter labextension install "@krassowski/jupyterlab-lsp"
```

```
Building jupyterlab assets
/Exception in thread Thread-2:

Traceback (most recent call last):
  File "c:\users\****\anaconda3\lib\threading.py", line 917, in _bootstrap_inner
    self.run()
  File "c:\users\****\anaconda3\lib\threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\****\anaconda3\lib\subprocess.py", line 1238, in _readerthread
    buffer.append(fh.read())
UnicodeDecodeError: 'cp932' codec can't decode byte 0xef in position 7095: illegal multibyte sequence
```

# Environment

- OS: Windows 7（Japanese）
- jupyterlab==1.0.2
- conda-forge/win-64::conda-package-handling-1.4.1-py37_0
- conda-forge/win-64::nodejs-12.8.1-0
- conda-4.7.11-py37_0
- Python 3.7.3

# Notes

- `cp932` is used in the standard output of Japanese Windows computers (mainly command prompt, like anaconda prompt). So probably this error will not be occurred English environment or other OS, like Ubuntu.
- This error happen both on the command line and JupyterLab's Extension Manager.

# Temporary solution

When I use `--debug` argument to inspect, then build completed successfully (probably the change of the standard output contents during the build seems to solve this problem).

```
$ jupyter labextension install "@krassowski/jupyterlab-lsp" --debug
```

krassowski:
Thanks for the detailed report and a workaround!

Unfortunately, the traceback is not very informative (everything seems to happen in the Python standard library). Also, this extension has no Python code, so it is not a simple "wrong encoding of a .py file" issue :)

simon-ritchie:
Hi, @krassowski 
Thx for reply!

> Also, this extension has no Python code, so it is not a simple "wrong encoding of a .py file" issue :)

Ah...that's right.  
Maybe it needs to fix code related to build on the JupyterLab side.  
Since it is not an issue of this extension, I will close the issue.

Thx.

krassowski:
Other than the installation issue, does the extension work well for you?

simon-ritchie:
@krassowski 

Yes, extension worked well (amazing works!!).  

> automatic signature suggestions
> ...
[...]

PR #315, opened 2020-08-11
Add completion settings (Hinterland mode, documentation, suppression)

### LINKED ISSUE #314: Allow to disable the documentation preview in completer
## What are you trying to do?

Add a setting disabling the documentation preview in the completer

## How is it done today, and what are the limits of current practice?

The new completer would always show documentation, which might not be what the user wants

krassowski:
## References

Implements #25, closes #314, addresses a long standing feature request by @simon-ritchie in early feedback [here](https://github.com/krassowski/jupyterlab-lsp/issues/7#issuecomment-523880845) (the Hinterland mode).

## Code changes

- store global references/state in `LSPExtension` to simplify passing stuff around (especially settings)
- add `IFeatureSettings` interface to allow isolate access to subset of settings by the features (implemented by `LabFeatureSettings`)
   - the interface is provisional as I imagine that server/language-specific behaviour would be nice to have (soon!) and for now it only does global stuff. I think that its responsibility will be to merge properties from global settings and language-specific settings (and give the feature whichever it wants).
- one hard-coded variable less (invocation suppression tokens)

## User-facing changes

- settings now include "features" property
  - display of documentation in the completer can be disabled
  - user can now decide which tokens to ignore
  - continuous hinting (suggestion after typing each key) is now implemented ("Hinterland mode")
- reused the diagnostics icon for settings
  - it may not be a perfect icon for LSP as a whole, but it seems good enough in this context; I think that most users may not be too familiar with LSP term (even though they may understand what LSP is when they are installing it - unless it comes pre-installed) so giving them a visual hint of the diagnostics icon could be useful (in settings context). Though I think it would be misleading if we did the same on the statusbar...

completion_settings

- `JupyterLabWidgetAdapter` constructor was rewritten

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [x] changelog entry


krassowski:
I guess I will have to write tests for that at some point... I guess I will do one for the documentation in completer behaviour tomorrow so that we have at least one test using the settings before we merge this.

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -1,12 +1,55 @@
...
+          }
+        }
+      }
+    }
+  },
   "properties": {
     "language_servers": {
       "title": "Language Server",
       "description": "Language-server specific configuration, keyed by implementation, e.g: \n\npyls: {\n  serverSettings: {\n    pyls: {\n      plugins: {\n        pydocstyle: {\n          enabled: true\n        },\n        pyflakes: {\n          enabled: false\n        },\n        flake8: {\n          enabled: true\n        }\n      }\n    }\n  }\n}\n\nAlternatively, using VSCode's naming convention:\n\npyls: {\n  serverSettings: {\n    \"pyls.plugins.pydocstyle.enabled\": true,\n    \"pyls.plugins.pyflakes.enabled\": false,\n    \"pyls.plugins.flake8.enabled\": true\n  }\n}",
-      "type": "object",
```
bollwyvl, line 9: it's usually good practice to leave the base type (even if it's obvious from having `properties` or `items`)...

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -16,20 +59,16 @@
...
-          "title": "Language Server Configurations",
-          "description": "Configuration to be sent to language server over LSP when initialized: see the specific language server's documentation for more",
-          "type": "object",
-          "default": {}
+    },
+    "features": {
+      "title": "LSP features",
+      "description": "Client-side features configuration",
+      "$ref": "#/definitions/features",
+      "default": {
```
bollwyvl, line 67: I am not sure if the `defaults` are actually preserved all the way down... it might just be the top-level object. would have to check.

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -1,12 +1,55 @@
...
+            "showDocumentation": {
+              "type": "boolean",
+              "default": true
+            },
+            "continuousHinting": {
+              "type": "boolean",
+              "default": false
+            },
+            "suppressInvokeIn": {
+              "type": "array",
```
bollwyvl, line 39: some description here, e.g. `an array of ...`

are these:
- lsp token kinds?
- codemirror types?

Do these need to be configured per-language server?

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -46,44 +46,32 @@ const lsp_commands: Array<IFeatureCommand> = [].concat(
...
-    IDocumentManager,
-    ICompletionManager,
-    IRenderMimeRegistry,
-    IPaths,
-    ILabShell,
-    IStatusBar
-  ],
-  activate: (
-    app: JupyterFrontEnd,
+export class LSPExtension {
```
bollwyvl, line 49: so i have lots of thoughts on this. I think the existing `LanguageServerManager` in `manager.ts` class can already take on this duty, and is starting to on the `add-language-server-kernel` branch...

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -46,44 +46,32 @@ const lsp_commands: Array<IFeatureCommand> = [].concat(
...
+  connection_manager: DocumentConnectionManager;
+  language_server_manager: LanguageServerManager;
+  settings: ISettingRegistry.ISettings;
+
+  constructor(
+    public app: JupyterFrontEnd,
```
bollwyvl, line 55: This signature is already way too complicated. A more common pattern would be to have a `ILanguageServerManager.IOptions` which defined all these pieces, and didn't conflate the variable scopes (public/private) in the constructor. Putting this in a (new) `tokens.ts` will help downstreams understand what the language server manager can provide to them.

The main advantage here is that it is easier to add a property to an unordered set of named arguments rather than getting the perfect ordering of many disparate types.

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -235,36 +209,69 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
+    ISettingRegistry,
+    ICommandPalette,
+    IDocumentManager,
+    ICompletionManager,
+    IRenderMimeRegistry,
+    IPaths,
+    ILabShell,
+    IStatusBar
+  ],
+  activate: (app, ...args) => {
```
bollwyvl, line 259: i see what you're doing here, and yes, relying on so many things is onerous. but this seems like a very roundabout way to leverage the type system... and basically guarantees that the signature, and therefore the API, will break the next time this changes. Sure, you're not returning this superobject yet, but this would all but need to be the public API.

my experience has been to minimize the number of external APIs that are actually called inside the manager class, and favor connecting to top-level signals in the `extension.ts`. this centralizes the potential API surface to the best

packages/jupyterlab-lsp/style/icons/diagnostics.svg
```diff
@@ -1,5 +1,5 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="24" viewBox="0 0 24 24">
-  <g class="jp-icon3" fill="#616161" shape-rendering="geometricPrecision">
+  <g class="jp-icon3 jp-icon-selectable" fill="#616161" shape-rendering="geometricPrecision">
```
bollwyvl, line 2: :+1:

packages/jupyterlab-lsp/src/adapters/jupyterlab/file_editor.ts
```diff
@@ -44,20 +41,11 @@ export class FileEditorAdapter extends JupyterLabWidgetAdapter {
   }
 
   constructor(
+    extension: LSPExtension,
```
bollwyvl, line 44: same thing here: when we start changing constructors, we should just go ahead and adopt an `IOptions` someplace so the options can grow more organically over time. The closer we can start making `<ActivityName>WidgetAdapter` invocation to be more regular, I think the happier we'll be...

packages/jupyterlab-lsp/src/adapters/jupyterlab/file_editor.ts
```diff
@@ -44,20 +41,11 @@ export class FileEditorAdapter extends JupyterLabWidgetAdapter {
...
+    jumper: FileEditorJumper
   ) {
-    super(
-      app,
-      editor_widget,
-      rendermime_registry,
-      'completer:invoke-file',
-      connection_manager
-    );
+    super(extension, editor_widget, 'completer:invoke-file');
```
bollwyvl, line 48: we should probably collect all of our external command references someplace...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -32,6 +36,37 @@ import {
   ISocketConnectionOptions
 } from '../../connection_manager';
 import { Rename } from '../codemirror/features/rename';
+import { LSPExtension } from '../../index';
+import { ISettingRegistry } from '@jupyterlab/settingregistry';
+
+class LabFeatureSettings implements IFeatureSettings {
```
bollwyvl, line 42: this screams generic to me... if it accepted an object, you could basically get away without having to cast anything anywhere.

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -32,6 +36,37 @@ import {
...
+    return (this.lab_settings.get(this.key).composite as any)[
+      this.feature_name.toLowerCase()
+    ];
+  }
+
+  get(setting: string) {
+    return this.get_obj()[setting];
+  }
+
+  set(setting: string, value: any) {
```
bollwyvl, line 64: here, you could reus the `T` from above, and always set to a `Partial<T>`: fewer magic strings, and no looping (outside of here) to set multiple values.

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -32,6 +36,37 @@ import {
...
+    let obj = this.get_obj();
+    obj[setting] = value;
+    this.lab_settings.set(this.key, obj).catch(console.warn);
```
bollwyvl, line 67: it seems it would be reasonable to make this return a `Promise<void>` instead of swallowing the error... at some point, we should figure out how to actually validate these...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -119,18 +154,31 @@ export abstract class JupyterLabWidgetAdapter
...
   protected constructor(
-    protected app: JupyterFrontEnd,
+    protected extension: LSPExtension,
     protected widget: IDocumentWidget,
-    protected rendermime_registry: IRenderMimeRegistry,
-    invoke: string,
-    connection_manager: DocumentConnectionManager
+    invoke: string
   ) {
+    this.app = extension.app;
```
bollwyvl, line 171: if possible, we should try to keep references to the application itself to a minimum...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -119,18 +154,31 @@ export abstract class JupyterLabWidgetAdapter
...
+    this.rendermime_registry = extension.rendermime_registry;
+    this.connection_manager = extension.connection_manager;
```
bollwyvl, line 173: indeed, many of these things could just become dumb getters that wrap `this.extension.X` (or `this.language_server_manager.X`)

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -119,18 +154,31 @@ export abstract class JupyterLabWidgetAdapter
...
     this.document_connected = new Signal(this);
     this.invoke_command = invoke;
     this.adapters = new Map();
     this.status_message = new StatusMessage();
-    this.connection_manager = connection_manager;
+    this.completion_settings = new LabFeatureSettings(
```
bollwyvl, line 178: this seems like sort of a weird place to land this...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
```diff
@@ -473,7 +521,8 @@ export abstract class JupyterLabWidgetAdapter
         virtual_document,
         connection,
         this,
-        this.status_message
+        this.status_message,
```
bollwyvl, line 524: this probably _also_ needs an `IOptions`

packages/jupyterlab-lsp/schema/plugin.json
krassowski: Where do I put the description in the schema for items on a deep level? I could not find this in schema docs (nor in any JupyterLab core extension).

packages/jupyterlab-lsp/schema/plugin.json
krassowski: Jut followed the pattern from core; for me, the reasoning is that it is visually confusing when looking where to put types., but could add it back.

packages/jupyterlab-lsp/schema/plugin.json
krassowski: yes, they seem to be for the practical reason; how else can if specify them? `properties` → `completion` and then default?

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
krassowski: I do not understand. Could you throw me some example, plese?

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
krassowski: It is, but it will get more complex to allow for merging values from both the language-specific and general-level settings for features.

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
krassowski: see https://github.com/krassowski/jupyterlab-lsp/pull/315/files#diff-768f6aea86149c0aa51e05111ea0968cR160. I don't like it either but seems the best thing to maintain separation of lab-specific and codemirror-specific code without major refactoring. I mean it should probably go like Feature = LabFeature + CMFeature + Settings (which are relevant to both) - I can work on it in a subsequent PR.

packages/jupyterlab-lsp/src/index.ts
krassowski: I think that it should stay this way. `LanguageServerManager` sounds like backend connector, and `LSPExtension` duty is to be a global state holder (see settings).

packages/jupyterlab-lsp/src/index.ts
krassowski: I get that there are a few advantages of `IOptions` pattern, but it does not seem greatly useful for a function which gets called once ever (and with arguments that we do not control). Besides, having `IOptions` here would just move the problem of relying on a specific ordering of arguments to the `activate()` function.

packages/jupyterlab-lsp/src/index.ts
krassowski: 1. not returning because activate has `void` signature
2. as long as the `requires` and `as []` stay in sync this will not break (ideally there would be a `const TOKENS = [...]` on `type TOKENS = [...]` defined in a single line, but I could not get it working)
3. yes the public API will change the next time we add a token dependency, but is this a bad thing? `LSPExtension` is only meant to be used in the activate function (can document this). I see that I export it but maybe if I exported an interface with the things that are meant to be passed down, then `LSPExtension` would no longer be public. On the other hand, I do not feel convenient having the extra interface if I can just have an extension object. The extension class just seems so natural to have (and it to contain app and settings references)

krassowski:
Thanks for the feedback! I will try to adress the points tomorrow.

krassowski:
@bollwyvl Most of the issues you point out are can be solved by #316. However, it is a substantial rewrite and I would like to publish this version tomorrow to get up to speed with Lab 2.2.

packages/jupyterlab-lsp/schema/plugin.json
bollwyvl: yeah, there's some weirdness in choices around these files (json5 wut?) in core. i highly recommend a spin through the [official docs](https://json-schema.org/understanding-json-schema/index.html) some point.

packages/jupyterlab-lsp/schema/plugin.json
bollwyvl: anything that can have a `type` can have a `title` and/or `description`, `examples`, `default`, etc. The issue is, again, that core has chosen to only show docs for the top level. if advanced settings can't be made useful, i say we ignore how bad it is, and do our own UI, which we want _anwyay_.

Or, we could _further_ accept the awful that is the dotted-notation from vscode, and hoist all the docs up that way, gaahhh

packages/jupyterlab-lsp/src/index.ts
bollwyvl: ehm. ok man.

packages/jupyterlab-lsp/src/index.ts
bollwyvl: advantages, off the top of my head:
- easier to document
- easier to extend
- better completion
- more like core

`activate` is needfully the special case, at the application level, so that things _can_ be replaced.

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
bollwyvl: we've already got a model (e.g. settings) for doing these kinds of merges... i think you would _still_ expect, for a given language, to get back an object that has a predictable type. better still, we should be deriving the types _from_ the json schema...

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
bollwyvl: i'll see if i can dig something up. Basically, if you know `T`, you can ask for `setting: keyof T`, and then the value can be inferred by type inference. I think `lumino` has some stuff like that.

packages/jupyterlab-lsp/src/index.ts
krassowski: Quick note: 1 no longer the case in the refactor, this was my limited understanding.

packages/jupyterlab-lsp/src/adapters/jupyterlab/jl_adapter.ts
krassowski: I had a go on schema-to-ts yesterday did not seem a great benefit for now, but will try again after features refactor :)

bollwyvl:
_this_ is also a substantial rewrite... i mean, whatever, but it's going to get pretty headachey to start bringing the kernel branch back up to snuff.

krassowski:
> i's going to get pretty headachey to start bringing the kernel branch back up to snuff

I know, but I am happy to resolve conflicts/apply the changes on top of the refactor to birng it back up

Issue #316, opened 2020-08-12
Maybe refactor features


krassowski:
Feature = LSP functionality,  e.g. completion, jumper, diagnostics.

Make it possible to split up the features and install the extra features.

## What are you trying to do?

Expose a token `LSPFeatureManager` which will be used to register the features. It would have a register method:

```typescript
class LSPFeatureManager implements TokenPattern {
    register(LSPFeatureManager.IFeatureOptions) {
        // add the feature to the list of features
    }
}

interface LSPFeatureManager.IFeatureOptions {
    feature: IFeature;
    /** ids of the features this feature relies on  */
    requires?: string[];
    /** ids of the features this feature wants to disable; 
    use it to override the default feature implementations with your custom implementation
    (e.g. a custom completer from Kite)  */
    supersedes?: string[];
}
```

The new feature interface:

```typescript
import { CodeEditor } from '@jupyterlab/codeeditor';
import { CodeMirrorEditor } from '@jupyterlab/codemirror';

export interface IFeature {
    id: string;
    // used in the GUI
    name: string;
    commands: Commands;
    // each feature can be written in mind with mind to support one or more editors;
    // by default we target CodeMirrorEditor, but let's think about allowing others too
    // supportedEditors: typeof CodeEditor[] = [CodeMirrorEditor];
    editorIntergration: Record<typeof CodeEditor, typeof FeatureEditorIntegration<CodeEditor>>;   
    labIntegration?: IFeatureLabIntegration;
    // note: settings could be one-per plugin rather than centrally managed by LSP (as proposed here);
    // the per-plugin settings may be nicer for the user, and would keep the schema near the plugin
    // but then combining per-language settings etc would be a pain
    register(settings: Settings): void;
}

export class CodeMirrorIntegration implements FeatureEditorIntegration<CodeMirrorEditor> {
   // some staff useful for CM integraion
}
```

each feature would then be an extension:

```typescript
const FEATURE_ID = '@jupyterlab-lsp/feature-completion:plugin'

import {BaseFeature, CodeMirrorIntegration} from 'jupyterlab-lsp'


class CompletionCMIntegration extents CodeMirrorIntegration {
     // register cm-specific actions
}

class CompletionLabIntegraion implements IFeatureLabIntegration {
    register() {
        // register custom data connector for completion
    }
}

class CompletionFeature extends BaseFeature implements IFeature {
   // does not have to be the same, but why not
   id = FEATURE_ID,
   name = 'Completion'

[...]

Issue #310, opened 2020-08-08
LSP Servers always Connecting...


walidsi:
Hi,

I followed the instructions at https://jupyterlab-lsp.readthedocs.io/en/latest/Installation.html in order to get jupyterlab-lsp up and running. Installation steps were successful without errors but upon opening a notebook in juypter lab, I noticed that the LSP Servers status is always "Connecting..."

![image](https://user-images.githubusercontent.com/42148514/89716285-02a21380-d9ac-11ea-9eae-32aba223854c.png)

What is wrong please? and how can I get it to work?

krassowski:
I think it is all ok (I see that you got the code diagnostics/linting) for
Python, I think that it says "connecting" because it cannot find JavaScript
language server. I will investigate.

EnchoC:
![image](https://user-images.githubusercontent.com/28055454/90272011-52db1300-de8f-11ea-8c9d-3f07837a6062.png)

Me too, server connecting。。

krassowski:
Thank you for the reports it should be fixed in the 2.0 version (to be released today).

walidsi:
Thanks for the update. How do I upgrade to version 2.0?

walidsi:
> 
> 
> Thank you for the reports it should be fixed in the 2.0 version (to be released today).

Thanks for the update. How do I upgrade to version 2.0? I tried removing everything and installing again but it seems v0.9 is still the one that gets installed.

krassowski:
Sorry for the delay, I have not released it yet, waiting for #328. @bollwyvl do you need a hand with testing, or maybe should I go ahead and make a release now?

bollwyvl:
Sorry for the delay, I have not released it yet, waiting for #328
> <https://github.com/krassowski/jupyterlab-lsp/pull/328>. @bollwyvl
> <https://github.com/bollwyvl> do you need a hand with testing, or maybe
> should I go ahead and make a release now?
>
Sorry, was heads-down on a talk the last couple days.

Haven't had the chance to start the tests, and agree it's a little weird to
just have one feature configurable. Go ahead and release, we'll fix it in
post!

>

krassowski:
@walidsi you can update now with:

```
pip install jupyter-lsp==0.9.2
jupyter labextension install @krassowski/jupyterlab-lsp@2.0.0
```

[...]

PR #318, opened 2020-08-13
Features refactor

krassowski:
## References

See #316 

## Code changes

- allows to add, replace or improve any of the LSP features as small independent extensions (by providing tokens: `ILSPFeatureManager` - for feature registration, and `ILSPAdapterManager` - for reacting to the widget (notebook/editor) adapter changes).
- enables editor-agnostic implementation of features (i.e. for editors like Monaco instead of CodeMirror, as long as they are wrapped in the `CodeEditor.IEditor` interface).
- greatly improves the code organisation
- slowly moves towards camelCase convention in public interfaces.
- [x] adds settings for:
  - [x] diagnostics
    - [x] `defaultSeverity` (no longer hard-coded)
    - [x] `ignoreCodes` (new, long-overdue string[])
    - [x] `ignoreMessages` (new, long-overdue regexp[] - can schema have regexp?)
  - [x] hover: `modifierKey` (no longer hard-coded)
  - [x] completer (see #315)

## User-facing changes

New statusbar icon:
Screenshot from 2020-08-20 23-57-57

Status popover icons restored & the document locator widget instead of document ids in the popover:
Screenshot from 2020-08-21 00-06-58

Settings:
Screenshot from 2020-08-21 00-04-22

Icons in the context menu (following the addition of icons to file editor context menu):
Screenshot from 2020-08-21 00-08-17

Everything but connection-related stuff was refactored, moving forward to 2.0 version on the frontend.

## Checklist:

- [ ] <s>split `style.css`?</s> (not this time)
- [ ] organise the features into directories (one file for implementation, one for the extension, <s>one for styles</s>)
   - [x] somewhat - for the two features which already grown in size; not splitting the hair for others for now (the very deep directories structure due to monorepo can be annoying, lets not exacerbate it for the sake of beauty)
- [ ] documentation of the public tokens and options of their methods:
   - [x] features
      - [x] ILSPFeatureManager (token)
      - [x] IFeatureOptions (registration options)
      - [x] IFeature
   - [x] code editors
      - [x] ILSPVirtualEditorManager (token)
      - [x] IVirtualEditorType (registration options)
      - [x] IVirtualEditor
   - [ ] document widgets
      - [ ] ILSPAdapterManager (token)
      - [ ] IAdapterTypeOptions (registration options)
      - [ ] WidgetAdapter (abstract class) - not documented, may be refactored to IWidgetAdapter (and documented) in the next iteration; low priority as no one has formally requested additional DocumentWidget integrations.
- [x] "extending frontend" documentation, including "implement your LSP feature"
- [x] make the tests pass
   - [x] jest
   - [x] robot
- [x] plugin-based DocumentWidget adaptors:
   - [x] refactor CommandManager (bring closer to AdapterManager?) to remove the tokens dependency in the extension
   - [x] refactor commands entrypoints so that those are types exported by adapters rather than enum
   - [x] create the plugins
- [x] further limit the number of public members of the extension (constrained via interface `ILSPExtension` interface)
- [x] add icons for features
- [x] create editor manager which will register virtual editor implementations.
   - [x] <s>The virtual editor implementation should state which adapters is it for.</s> Let's make `VirtualEditor` adapter-independent
     - [x] `VirtualEditor` cleanup (goal: keep only editor implementation-specific methods in the interface):
         - move document creation to adapter (it knows the path/lang),
         - create `VirtualDocument.UpdateManager` and move update functions there (except for the actual update)
         - replace cell-based notebook-specific methods with editor-based generic interface
     - [x] make the `VirtualEditor` return a collection of editors and implement the multi-editor (i.e. multi-cell) actions outside to be able to provide a single editor implementation for all the adapters
         - note: the `VirtualEditor` still has to be aware of existence of multiple editors but now it does not care where they came from (i.e. cells or file editor).

Future work (not in the scope of this PR):
- refactor `EditorAdapter` (maybe rename it so there no confusion on which adapter is meant). `EditorAdapter` is feature instance manager, but naming is such might cause confusion with `FeatureManager` which is a token that registers and providing feature implementations (classes); `EditorAdapter` probably should take more responsibility for the feature's creation (away from the `WidgetAdapter`) and get better documentation.
   - the way the features require a callback to act on changes is sub-optimal, it would be better with a signal
- the virtual editor plugins could provide the virtual editor tokens so that features may require them (not necessarily to use them in the implementation, but to state the dependencies), or
- <s>maybe the feature could register itself to a virtual editor implementation rather than to the feature manager - this could simplify things a bit.</s> but then the features would need to be redefined for each editor, which would confuse the user, so let's keep it as is.
- add a setting allowing to disable specific features from the extension; it should accept an array of identifiers (strings) which will be validated at runtime; this is a responsibility of features manager (in future we should have GUI for that because the user has no way to know the identifiers without looking into the code).

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


packages/jupyterlab-lsp/src/features/jump_to.ts
```diff
@@ -135,3 +148,78 @@ export class JumpToDefinition extends CodeMirrorLSPFeature {
     }
   }
 }
+
+class JumperLabIntegration {
```
bollwyvl, line 152: This is a key opportunity to do the './tokens.ts', thing, to make it possible to provide an alternate implementation:

```ts
export interface IJumperLabIntegration {
...
}
export namespace IJumperLabIntegration {
  export interface IOptions {
    fileEditorTracker: IEditorTracker,
    notebookTracker: INotebookTracker,
    documentManager: IDocumentManager
  }
}

IJumperLabIntegration = new Token()
```

Otherwise, if people have to _subclass_ your thing, they are stuck with whatever choices we make about private/public etc.

I imagine the bulk of the interface methods _mostly_ take LSP messages and (Virtual)Documents at the boundaries.

Also I _highly_ recommend splitting up the extension and the feature implementation in two files, and forcing the extension to only use the interface methods. Further, passing high-level token-implementers into implementations makes the graph pretty messy.  Looking, for example, at this, it can only ever speak editor and notebook, and couldn't have (say) the SettingsEditor attached to it. Further, the implementation is almost identical. So maybe,

```
interface IJumperLabIntegration {
  registerTracker(IWidgetTracker, )
}
```

so that this feature no longer knows about notebooks or editors, just widgets that want to be jumped to.

krassowski: > splitting up the extension and the feature implementation in two files

If you mean `jump_to/index.ts` (extension) `jump_to/feature.ts`(implementation) (+ any additional files required) then yes, I have that in my plans (not sure if in this PR or subsequent).

krassowski: > passing high-level token-implementers into implementations makes the graph pretty messy. Looking, for example, at this, it can only ever speak editor and notebook, and couldn't have (say) the SettingsEditor attached to

Yeah, I agreed, but the _almost_ and lack of the underlying implementation of a the jumper for `SettingsEditor` makes me postpone such refactor. Note, it is not public (not exported).

krassowski:
Is it worth the effort to split CSS styles and move icons closer to the features too?

bollwyvl:
> Is it worth the effort to split CSS styles and move icons closer to the features too?

Yeah, been thinking about this on some other stuff. If they become separate packages/plugins, you'd want to only install the CSS/etc if the plugin actually gets loaded (saving the hit, and additional selectors... everything has a cost, even if its not really being used). I'd think while they are in the same tree, and are instantiated as part of the same file, there aren't a lot of good tools for it, so may not be worth it.

Another relatively important thing, though, is how the initialization params would get created. Trying to do magic name stuff, e.g. `featureName + 'Provieder` seems like it would get tiring. I'm thinking each feature gets a chance to modify the object, e.g. 

```ts
ILanguageClientFeature {
  ...
  initializeParams(params: InitializeParams): InitializeParams
}

class Connection {
  ...
  initializeParams(): InitializeParams {
    let capabilities = this.defaultInitializeParams();
    for (const feature of this.whatever.features) {
       params = feature.initializeParams(params);
    }
    return params;
  }
}
```

...and these in turn can react to settings. Or, maybe `lodash.mergeWith` can take care of this and they could be fixed at registration time, but I think some escape hatch preserve some dynamism is important.

krassowski:
I like the idea of initialization params transformer! This way we can have language-specific LSP settings as part of the feature settings. However, as `IFeature` is language-agnostic (server-agnostic) we need either:
- to make the settings language-aware and only provide the feature with the settings that are specific to the currently connected server
- to make the initParams transformation static (i.e. not dependent on the server in use)

Another layer of troubles is in settings GUI - if each feature settings become language-aware, how do we make the per-language settings pleasant for the user? Should we hide the server for features settings (as it is an implementation detail), show per-language options and only allow to change the per-server initialization params in the global LSP settings?

I feel that this would need a separate issue/PR.

packages/jupyterlab-lsp/package.json
```diff
@@ -29,7 +29,11 @@
   },
   "scripts": {
     "build": "jlpm build:schema && tsc -b",
-    "build:schema": "json2ts ../../py_src/jupyter_lsp/schema/schema.json --unreachableDefinitions | prettier --stdin-filepath _schema.d.ts > src/_schema.d.ts",
+    "build:schema": "jlpm build:schema-backend && jlpm build:schema-completion && jlpm build:schema-hover && jlpm build:schema-diagnostics",
+    "build:schema-backend": "json2ts ../../py_src/jupyter_lsp/schema/schema.json --unreachableDefinitions | prettier --stdin-filepath _schema.d.ts > src/_schema.d.ts",
+    "build:schema-completion": "json2ts schema/completion.json | prettier --stdin-filepath _completion.d.ts > src/_completion.d.ts",
+    "build:schema-hover": "json2ts schema/hover.json | prettier --stdin-filepath _hover.d.ts > src/_hover.d.ts",
+    "build:schema-diagnostics": "json2ts schema/diagnostics.json | prettier --stdin-filepath _diagnostics.d.ts > src/_diagnostics.d.ts",
```
krassowski, line 36: This works fine now, but to get a nice name of the interface (e.g. `LSPCompletionSettings`) I needed to change the title of the schema (e.g. to `"title": "LSP Completion settings"`) which is a bit ugly in the interface:

Screenshot from 2020-08-14 14-16-29

I think that keeping the prefix is fine because this makes all the settings neatly close to each other (and I am thinking about keeping the schema simple and core-compliant to get nice GUI settings once they land in the core).

The "Settings" suffix is ugly. Well, we could always alias imports... Or is there a better trick?

bollwyvl:
So on settings: yeah, `Advanced Settings` is kind of a mess... while the schema goes all the way down, the docs it generates give up at the top.

I think we'll just need to own it, eventually, and have our own settings UI that uses rjsf and/or a JSON editor that has a client-side language server (a la #184). This is where that jumper would come in, e.g. going from a place in the form UI into a specific place in the schema. Definitely a different PR.

Also, we should probably avoid the acronym in strings the user sees in the UI... so perhaps `Language Client Completion Settings` vs LSP.

Note that we can go the _other_ direction, and build a schema from a typescript type, if you find that it's too restrictive... i don't think we want anything that goes all the way back to the server (since the server response already lives over there) but within the frontend, we should do whatever is best for DX/UX.

bollwyvl:
As for per-language-server stuff, i think a pattern like so would do the trick:

```yaml
type: object
default:
  global:
    enabled: true
properties:
  global:
    $ref: "#/definitions/completion-settings"
  language_servers:
      patternProperties:
        "*":
          $ref: "#/definitions/completion-settings"

definitions:
  completion-settings:
    properties:
      enabled:
        type: boolean
      ...
```

krassowski:
Here is how it would look like with "Code" prefix (for brevity):

Screenshot from 2020-08-14 19-49-36

krassowski:
I am happy to change any of the icons (completion/diagnostics/hover/lang server) but at least we have the boilerplate in. Suggestions welcome! These are all from http://materialdesignicons.com/ (and probably best to keep it this way for consistency).

bollwyvl:
Yeah, any icon is better than the gear. 

Actually, amending the above schema, it's probably best to forgo the `global`... so something more like...

```json
{
  "title": "completion",
  "allOf": [
    {"$ref": "#/definitions/completion-settings"},
    {
      "properties": {
        "language_servers": {
          "$ref": "#/definitions/language-servers"
        }
      }
    }
  ],
  "definitions": {
    "completion-settings": {
      "type": "object",
      "properties": {
        "continuousHinting": {
          "type": "boolean"
        }
      }
    },
    "language-servers": {
      "type": "object",
      "patternProperties": {
        ".*": {
          "$ref": "#/definitions/completion-settings"
        }
      },
      "additionalProperties": {
        "$ref": "#/definitions/completion-settings"
      }
    }
  }
}
```

krassowski:
would it be possible to make it work if we had definitions for completions in another file (as it is now on this PR?)

bollwyvl:
> definitions for completions in another file

Looking, but can you point a link to what you mean? For this purpose, I can't recommend having cross-file schema `$refs`, for portability and versioning sake, even though they are supported by the spec... it is highly implementation- (and environment-specific) as to how the refs resolve (shades of XML parsing nightmares). 

Also, on these settings, it's _really_ hard to do versions/migrations properly. Better to be super lax (vs on the server side, where we benefit from being very precise).

packages/jupyterlab-lsp/schema/completion.json
```diff
@@ -0,0 +1,32 @@
...
+      "title": "Suppress invoke in specific code fragments",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "default": ["comment", "string"],
+      "description": "An array of CodeMirror tokens for which the auto-invoke should be suppressed. The token names vary between languages (modes)."
+    }
+  },
+  "additionalProperties": false,
```
bollwyvl, line 30: can't recommend `additionalProperties:false`, can become very hard to upgrade.

packages/jupyterlab-lsp/schema/completion.json
```diff
@@ -0,0 +1,32 @@
...
+      "description": "Whether to enable continuous hinting (Hinterland mode)."
+    },
+    "suppressInvokeIn": {
+      "title": "Suppress invoke in specific code fragments",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "default": ["comment", "string"],
+      "description": "An array of CodeMirror tokens for which the auto-invoke should be suppressed. The token names vary between languages (modes)."
```
bollwyvl, line 27: awesome. I'll sometimes point to a codemirror [theme](https://github.com/codemirror/CodeMirror/blob/master/theme/solarized.css) when trying to explain cm tokens.

bollwyvl:
Yeah, i guess I'd just keep the defs for completion _in_ completion, rather than trying to shoehorn everything back into the "here's how i connect to servers" thing. It's slightly awkward, but we can do a better job when we decide to do a proper GUI.

bollwyvl:
Or, go the other way, and have typescript be the source of truth, and generate the schema, duplicating the defs in two files... but i don't think the "main" schema should plan on being able to know even what features it supports. Sigh.

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -3,222 +3,305 @@ import {
...
+const WIDGET_ADAPTER_MANAGER: JupyterFrontEndPlugin<ILSPAdapterManager> = {
+  id: PLUGIN_ID + ':ILSPAdapterManager',
+  requires: [IEditorTracker, INotebookTracker, ILabShell],
+  activate: (
+    app,
+    fileEditorTracker: IEditorTracker,
+    notebookTracker: INotebookTracker,
+    labShell: ILabShell
+  ) => {
+    const widgetTypes: IWidgetTypeOptions<IDocumentWidget>[] = [
```
krassowski, line 207: Just need to convert this into the Token pattern. From major architectural issues there is also VirtualEditor-as-a-token (and making adapters editor-implementation aware so they only register a cell/editor which is from an editor with an associated VirtualEditor (e.g. CodeMirror with VirtualCodeMIrrorEditor)).

krassowski: Also, move this out of index.

packages/jupyterlab-lsp/src/adapters/notebook/index.ts
```diff
@@ -0,0 +1,46 @@
+import { CommandEntryPoint } from "../../command_manager";
+import { JupyterFrontEndPlugin } from "@jupyterlab/application";
+import { ILSPAdapterManager, PLUGIN_ID } from "../../tokens";
+import { INotebookTracker } from "@jupyterlab/notebook";
+import { NotebookAdapter } from "./notebook";
+import { IDocumentWidget } from "@jupyterlab/docregistry";
+import { IAdapterTypeOptions } from "../../index";
```
krassowski, line 7: Should go to adapter manager or tokens file

packages/jupyterlab-lsp/src/feature.ts
```diff
@@ -0,0 +1,177 @@
...
+   * Will allow the user to disable specific functions
+   */
+  abstract isEnabled(): boolean;
+
+  /**
+   * Remove event handlers on destruction
+   */
+  abstract remove(): void;
+
+  // TODO: replace with a signal
```
krassowski, line 153: Should be resolved before merge

packages/jupyterlab-lsp/src/features/completion.ts
```diff
@@ -0,0 +1,202 @@
...
+import completionSvg from '../../style/icons/completion.svg';
+
+export const completionIcon = new LabIcon({
+  name: 'lsp:completion',
+  svgstr: completionSvg
+});
+
+export class CompletionCM extends CodeMirrorIntegration {
+  private _completionCharacters: string[];
+  // TODO chek if this works if yest then remove settings from options
```
krassowski, line 35: It works and indeed settings should be a responsibility of the widget only and removed from the init options

bollwyvl:
A lot of awesome work here, I gave up trying to review per-commit :blush:!

Perhaps my goal, once it simmers down a bit, will be to update the architecture stuff as a critical (in the literary, not personal, sense, of course!) read of what ends up being built and transferring it to the diagram. This will guide the frontend equivalent of the server-side documentation on extending.

We can stay with the graphviz approach... or we could consider adopting the [fork of jupyterlab-drawio](https://github.com/deathbeds/jupyterlab-drawio/) that I have been trying to get into a more capable and testable state. The disadvantage of drawio (however you serialize it) is _much_ harder to revision control than graphviz... but is really quite pleasant to edit and read for tree/graph-style documents.

Onward!

krassowski:
I was thinking the same! But on the other hand, could we (semi-)automate this? Should we?

I would love to have an interactive UML if possible (I mean with links to respective documentation pages). The only drawback being less control over the layout  I know it's possible to embed HTML links in dot.

bollwyvl:
Ha, this is encroaching on my day job stuff. Automatic layout is good, but
is also np-hard. The graphviz stuff is "good enough" for most cases, since
we're not trying to sell anything. The best thing I've found is elkjs which
is part of the eclipse ecosystem.

Automatic diagram generation is good for that which _does_ exist, but less
good for that which _might_ exist. If the tool generates an underlying
model, not just a picture, you can do more after the fact, but I don't have
a solid in-lab tool for model-based engineering... I hope to mature the
drawio stuff to that at some point, probably though ipywidgets first.

Drawio can also do collapsible containers, which, in lieu of a model, at
least can give you several appropriate views... But doesn't work when
embedded as a static svg (even in iframe). Graphviz doesn't really do this.

Drawio can embed links, but much like graphviz, you have to jump through
some hoops (iframe) to make them "nice" in docs... But I'm not sure they
work at all in the drawio editable svg, which is coming to be my preferred
in-repo format, at it is easy trip link/preview. embedded in a notebook is
a cool parlor trick, but has limitations like not being a picture browsers
know how to draw.

packages/jupyterlab-lsp/src/components/statusbar.tsx
```diff
@@ -459,7 +483,9 @@ export namespace LSPStatus {
       if (!this.adapter) {
         return stopIcon;
       }
-      return iconByStatus[this.status.status];
+      return codeCheckIcon.bindprops({
```
krassowski, line 486: Taking the opportunity given the new shiny LSP icon (which can be changed at a later time)

packages/jupyterlab-lsp/src/components/statusbar.tsx
```diff
@@ -437,10 +460,11 @@ export namespace LSPStatus {
       let status: StatusCode;
       if (detected_documents.size === 0) {
         status = 'waiting';
-        // TODO: instead of detected documents, I should use "detected_documents_with_LSP_servers_available"
       } else if (initialized_documents.size === detected_documents.size) {
         status = 'initialized';
-      } else if (connected_documents.size === detected_documents.size) {
+      } else if (initialized_documents.size === documents_with_servers.size) {
+        status = 'initialized_but_some_missing';
+      } else if (connected_documents.size === documents_with_servers.size) {
```
krassowski, line 467: While could be done in a subsequent PR, I needed a better reporting for the tests sake. Solves #310.

krassowski:
Re-graphs - this thing seems promising: https://github.com/sverweij/dependency-cruiser/blob/develop/doc/real-world-samples.md

krassowski:
Two pending things are:
 - confirming that all the tests pass on CI (i.e. coercing the Azure Pipelines to kick in...)
 - updating architecture in docs (@bollwyvl let me know if you still want to have a go; have look at the changing to "Extending" I added, these should be a good starting point).

krassowski:
Maybe I should change the PR base to master to get the CI going? Or, should we change something in the Azure CI settings?

krassowski:
Hooray, passing now! Except for the "warm up tectonic cache" step failing (randomly?).

krassowski:
This will need  improvements in dispose steps as there are race conditions when closing files:
Screenshot from 2020-08-27 01-34-02

krassowski:
This time tectonic cache failed on download/DNS resolution:

```python
/bin/bash --noprofile --norc /home/vsts/work/_temp/e3d07215-4c98-4b2c-858a-97644611174e.sh
note: this is a BETA release; ask questions and report bugs at https://tectonic.newton.cx/
note: connecting to https://archive.org/services/purl/net/pkgwpub/tectonic-default
note: resolved to https://tectonic.newton.cx/bundles/tlextras-2018.1r0/bundle.tar
note: downloading index https://tectonic.newton.cx/bundles/tlextras-2018.1r0/bundle.tar.index.gz
error: https://dataverse.harvard.edu/api/access/datafile/:persistentId?persistentId=doi:10.7910/DVN/GFFCIX/ENK9FL: error trying to connect: failed to lookup address information: Temporary failure in name resolution
caused by: failed to lookup address information: Temporary failure in name resolution
  File "scripts/tectonic_cache.py", line 28, in <module>
    tectonic_cache()
  File "scripts/tectonic_cache.py", line 24, in tectonic_cache
    subprocess.check_call(["tectonic", str(tex)], cwd=td)
  File "/usr/share/miniconda/lib/python3.6/subprocess.py", line 311, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['tectonic', '/tmp/tmpciv3a7li/example.tex']' returned non-zero exit status 1.
```

I'm going to add a re-try in Python cache script for that.

krassowski:
Why on earth the tectonic is not getting installed and mamba does NOT raise an error due to failed resolution?

krassowski:
Ironically, tectonic only works on Windows now and it does not install on Linux and MacOS :facepalm:

Issue #191, opened 2020-02-08
Formalize and extend language transclusion


bollwyvl:
## Elevator Pitch

Allow a user to define rules for finding parts of files in a host language that embed other languages. Additionally support referring to the outer language within the inner language. Language servers, libraries, and other things a user might have installed should be easily and robustly able to ship and activate these on installation.

## Motivation

Much like with #190 for #189, there are more languages that embed other languages than we'll be able to (or want to) define/maintain. The [existing list](https://github.com/krassowski/jupyterlab-lsp/blob/f8eaa9a6193710c3a161b74e45ac3dbb14d45270/packages/jupyterlab-lsp/src/extractors/defaults.ts#L26) includes things from IPython built-ins, but doesn't fully capture the scope of other magics, much less what other languages do, especially _extremely_ crazy things like [allthekernels](https://github.com/minrk/allthekernels), [sos](https://github.com/vatlab/sos-notebook) or [metakernel](https://github.com/Calysto/metakernel).

A number of these kinds of things also allow for referring to the outer language, such as `{variable}` references, up to full-blown templating languages, and having access to completion/diagostics/reference/hover for these nested scopes would be very valuable.

## Design Ideas

Add a separate spec to the JSON schema, e.g. `transclusions` which can be defined/modified without re-compiling lab by traitlets config or `entry_points`. Move the existing "simple" ones from the hard-coded typescript into defaults on the python side, or simple JSON.

### Notional "Simple" Example
```yaml
LanguageServerManager:
  language_servers: ...
  transclusions:
    ipython-shell-magics:
      host:
        mime_types: 
          - text/x-python
        languages: 
          - python
        file_extensions:
          - .ipy
          - .ipynb
      guest:
        mime_types:
          - application/x-sh
        languages: 
          - sh
        file_extensions:
          - .sh
      patterns:
        line-magic:
          regex: "(?<=\\s)!(.*)$"
          extract: $1
          isolate: true
          host_inclusions:
            variable:
              pattern: "(?<!\\$)\\{([^\\}]+)\\}"
              extract: $1
        cell-magic:
          regex: "^%%bash[^\n]*\n([\n.]+)$"
          extract: $1
          isolate: true
          host_inclusions:
            variable:
              pattern: "(?<!\\$)\\{([^\\}]+)\\}"
              extract: $1
```

### More complex things
[...]

PR #321, opened 2020-08-22
Make code extractors and replacements modular

krassowski:
This is a small step towards #191. I would postpone any major refactor until the JupyterLab 3.0 with the pip-based extension system ships, which seems that could change how we plan things a bit.

Pros:
- makes the code organisation a bit more sane (so that the different IPython extensions are not mixed together),
- fixes an `%%html --isolated` magic bug (thanks to new tests!)
- adds support for extraction of SQL from %%sql magics
- provides tentative tokens for extending the transclusions:
  - I am not sure if having two tokens (as implemented here) rather than one large "TransclusionsManager" is the way to go in the future; it is, however, the way to make some progress on a topic that no one wanted to tackle for some time and to reduce technical debt; we can change the API in the future.

Cons:
- it introduces an interface which we already plan to evolve and possibly break

## Checklist
- [x] Add `ILSPCodeExtractorsManager` token
- [x] Add `ILSPCodeReplacementsManager` token
- [x] Reorganise transclusions:
  - [x] split the defaults into two packages:
     - [x] `jupyterlab-lsp:extractors-ipython`, scope extractors for IPython built-in magics and expressions only
     - [x] `jupyterlab-lsp:extractors-rpy2`
  - [x] implement `jupyterlab-lsp:extractors-ipython-sql` to support https://github.com/jupyterlab/jupyterlab/issues/8879 (not an IPython built-in)
- [x] Document both the tokens
- [ ] Maybe implement host_inclusions as proposed in #191 (would benefit SQL!)

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


packages/jupyterlab-lsp/src/extractors/manger.ts
```diff
@@ -0,0 +1,29 @@
+import { JupyterFrontEndPlugin } from "@jupyterlab/application";
```
bollwyvl, line 1: typo in file name?

krassowski:
IPython-SQL has a connection concept which can be set per each cell or via line magic. It would be nice to be able to extract it and pass it to the virtual document so that it can be passed on to the language server during document initialization.

packages/jupyterlab-lsp/src/extractors/types.ts
```diff
@@ -40,7 +40,7 @@ export interface IExtractedCode {
  */
 export interface IForeignCodeExtractor {
   /**
-   * The foreign language.
+   * The foreign language. TODO what is language, what is host language, ILanguageInfo
```
krassowski, line 43: This. We need to document what are valid language strings. Changing the host language to accept `ILanguageInfo` would be difficult and not very practical (as it would require matching of the extractors/overrides by the language version/mimetype and it is not obvious how to act on a combination of those), but having foreign language as `ILanguageInfo` would be very helpful as it could be used to provide preferred mimetype for syntax highlighting, and it might replace the need specify the file extension; this is also a subset of already established Jupyter interface, so lot of benefits here!

PR #322, opened 2020-08-22
Implement completer icons and add a stub for themes system

### LINKED ISSUE #303: Choose and then add icons for the completion items
Completion items list can now use items, as implemented in https://github.com/jupyterlab/jupyterlab/pull/8140. We should take use of that as this is very low-effort high-gain feature. Currently the first letter and a colour code is used to indicate completion type. It will make it easier and more efficient to choose the right completion.

Costs:
- We need to bring the icons. The types/icon set is well defined by the LSP specification.

Potential icons:
- [gnvim](https://github.com/vhakulinen/gnvim/tree/master/assets/icons) uses feather which is MIT
- official [VSCode](https://github.com/microsoft/vscode-icons) icons are release under CC BY 4.0
   - some users will like it
   - we would need to think where to put the attribution - I do not see a good place for it...
- any other options?

**Please do propose other options and discuss** - this is the kind of issue anyone can get involved in - no extension coding skills required!

krassowski:
## References

Closes #303

## Code changes

- adds `ILSPCompletionThemeManager` token which registers themes implementing `ICompletionTheme`
- replaces namespaces used for constants mapping (and `inverse_namespace` helper) to TypeScript enums
- adds `eslint-plugin-import` with `import/no-cycle` because enums break when there are cycles (don't ask)

## User-facing changes

completer_icons

| VSCode | Material |
|--------|-------|
| Screenshot from 2020-08-24 10-40-57 | Screenshot from 2020-08-24 10-37-43 |

## Checklist:

- [x] implement "Material"/mdi theme
   - [ ] maybe set it to default?
   - [x] remove `jp-icon-selectable` as it is of no use there
- [x] implement "About Completion Themes" dialog window (it is a command palette)
   - [x] add short licence info generated from the metadata of `ILSPCompletionThemeManager` to comply with agreement
- [x] add acceptance tests, for three settings cases:
   - [x] `theme: null`
   - [x] `theme: "vscode"` or `icons: "material"`
   - [x] `theme: "a-non-existing-theme"`
- [x] add acceptance test for dark/light mode icons (probably with vscode theme)
- [x] add short documentation entry on creating custom completer icon themes

Wish list of completer UX issues which should be addressed ASAP (before release):
- [x] add a test for completer in FileEditor as it seems to be off
   - <s>add a test for completer rebinding after switching focus in multi-panel setting</s> would be nice, but not needed for this PR since the above one solved the issue

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry


krassowski:
@bollwyvl I was experimenting with splitting the themes into packages so that we would have:

packages/
- jupyterlab-go-to-definition
- jupyterlab-lsp
- lsp-ws-connection
- metapackage
- theme-material
- theme-vscode

However, the themes depend on (tokens of) `jupyterlab-lsp` so getting their plugins and re-exporting them from `jupyterlab-lsp` would create circular dependencies (is that very bad?). An alternative would be to start distributing metapackage rather than the extension itself (so that these two default themes don't need to be manually installed - as they are supposed to be there by default). What direction (introduction of a circular dependency/metapackage) is preferable?

bollwyvl:
It'll be this evening before I can really dig into it, fully...



Recall, as well, that a big goal of doing all this refactoring, and the
venerable #76, is that new features can be added by third-party extension
that we don't know about, yet, e.g symbol tree, formatting, etc. Or even
things that are unofficial extensions to LSP, like the latex PDF forward
search. So to that end, I think something purely cosmetic like the icon
themes, or otherwise unknowable like per-language server config, need to be
best-effort, and not too dominated by engineering rigor... a fallback to
letters (eg, first letter + first uppercase) isn't the worst thing in the
world.


On Tue, Aug 25, 2020, 11:16 Michał Krassowski <notifications@github.com>
wrote:

> @bollwyvl <https://github.com/bollwyvl> I was experimenting with
> splitting the themes into packages so that we would have:
>
> packages/
>
>    - jupyterlab-go-to-definition
>    - jupyterlab-lsp
>    - lsp-ws-connection
>    - metapackage
>    - theme-material
>    - theme-vscode
>
> However, the themes depend on (tokens of) jupyterlab-lsp so getting their
> plugins and re-exporting them from jupyterlab-lsp would create circular
> dependencies (is that very bad?).
>

There's really no harm in having packages that are just interfaces, if it
avoids circular imports, which are indeed the 😈, and force
ts/webpack/yarn/lerna to make bad decisions sometimes.

They wouldn't need to expose extensions, either... Just be real dependency
packages.

> An alternative would be to start distributing metapackage rather than the
> extension itself (so that these two default themes don't need to be
> manually installed - as they are supposed to be there by default).
>

What direction (introduction of a circular dependency/metapackage) is
> preferable?
>
The existing Metapackage is only for dev purposes, and should never be
published. We should feel free to do whatever makes development easy there

If anything, we'd have a brand new package, which would re-export all the
tokens. Indeed, this would probably get the name jupyterlab-lsp, for
continuity, and the other stuff could end up in "jupyterlab-lsp-core",
which wouldn't even be an extension... Is a pain before it gets published,
then usually okay.

Anyhow... Hacking up some  diagrams/fake docs might help put a point on it

krassowski:
> There's really no harm in having packages that are just interfaces, if it avoids circular imports,

Good point. This led me to an idea of promoting the completion themes manager to a package which breaks the cycle. It might not stay here long (after another round of features refactor it might get back into `@org/feature-completion/themes`) but it can wait IMO.

bollwyvl:
Very good stuff! Let me know how I can be of further rubber-duck assistance!

bollwyvl:
Waiting for binder and local to build... 

binder:
- [Binder](https://gke.mybinder.org/v2/gh/krassowski/jupyterlab-lsp/completer-icons?urlpath=lab)
- I have observed it die trying to install the env 
  - we may need to move to mamba in binder, as we have done on [_fam](https://github.com/deathbeds/_fam), or pre-compute the envs with conda-lock.

screenshots:
- colors
  - ...maybe pull through the vs colors into the material theme... 
    - i don't know if people have actually pinned those things in their brains 
      - _functions are purple, darnit, but hands off my gold enums!_ 
      - but it's a little more interesting, perhaps 
    - luckily, it isn't like there's a green _block_ and a red _block_ that mean drastically different things
  - ...and/or offer an `icon_colors` config option
    - `colorful` to ape the vs colors
    - `monochrome` (the existing grey)
  - ...and/or offer `opacity`
  - ...and/or offer a "no_icons" icon theme, for the truly curmudgeonly, but that seems rather austere 
  - ...and at least have _some_ meaning for theme `null`
  - like all these things, _very_ subjective: I'd probably pick: `material`, `grey`, and `0.5` for daily driving. 
- rosetta stone dialog
  - continuing my crusade, please no more dialogs for things that don't potentially lose data without undo
  - a closable main area widget would 
  - ideally show all themes next to each other
  - more ideally let you click the theme to activate it

More when i get stuff building, somewhere...

bollwyvl:
copy:
- probably prefer "Language Server" to "LSP" acronym in layed out-content

package names:
- probably _should_ have lsp in them
- probably _isn't_ specific to completion
  - whenever we get around to symbol list/tree, we'll want these, too, so they need to be available by some other means...
- naming might help, as they aren't jupyterlab themes (just icons) and don't work without the rest of the kit
  - `lsp-symbol-icons`
  - `lsp-symbol-icons-material`
  - `lsp-symbol-icons-vscode`

bollwyvl:
Some issue with the symbol kinds of non-selected items in vscode theme:

| vscode | material |
|---|----| 
| Screenshot from 2020-08-27 09-25-00 | Screenshot from 2020-08-27 09-24-26 |

bollwyvl:
Offtopic, probably doesn't need fixing here:

- The hint area should probably be `overflow-y: auto` instead of `overflow: scroll`... the list of completer items is probably fine as scroll, however. 
- Subjectively, as well, the symbol names could likely also be in the ui (vs code) font, as they are part of the narrative.
- It would also be rad if the hint area was syntax highlighted... 
  - having a persistent codemirror item that can be be lit up would be great, _especially_ if those tokens were clickable, e.g. to be able to jump to the reference of the return type of a method...

packages/completion-theme/src/index.ts
```diff
@@ -0,0 +1,175 @@
+import { kernelIcon, LabIcon } from '@jupyterlab/ui-components';
```
bollwyvl, line 1: recommend splitting up into:
- `tokens.ts`
  - tokens, constants, etc
- `manager.ts`
  - the manager implementation
- `extension.ts`
  - just the labextension machinery
- `preview.ts`
  - widget to do the config UI (not dialog)

packages/completion-theme/style/index.css
```diff
@@ -0,0 +1,18 @@
+.lsp-completer-themes .lsp-licence {
+  display: inline;
+}
+
+.lsp-completer-themes ul {
+  list-style: none;
+  padding-left: 10px;
```
bollwyvl, line 7: use theme variable (i like notebook padding for things like this)

packages/jupyterlab-lsp/schema/completion.json
```diff
@@ -25,6 +25,12 @@
       },
       "default": ["comment", "string"],
       "description": "An array of CodeMirror tokens for which the auto-invoke should be suppressed. The token names vary between languages (modes)."
+    },
+    "theme": {
```
bollwyvl, line 29: guess this would be a new file, if it became `lsp-symbol-theme`, etc.

packages/jupyterlab-lsp/src/features/completion/completion.ts
```diff
@@ -71,15 +72,20 @@ export class CompletionLabIntegration implements IFeatureLabIntegration {
...
+    settings.changed.connect(() => {
+      completionThemeManager.set_theme(this.settings.composite.theme);
+    });
   }
 
   private swap_adapter(
     manager: ILSPAdapterManager,
     adapter: WidgetAdapter<IDocumentWidget>
   ) {
+    console.log('ADAPTER CHANGED TO ', adapter);
```
bollwyvl, line 88: get rid of, put behind flag, etc.

packages/jupyterlab-lsp/src/features/completion/completion.ts
```diff
@@ -150,6 +164,7 @@ export class CompletionLabIntegration implements IFeatureLabIntegration {
     }
     this.current_completion_connector = new LSPConnector({
       editor: editor,
+      icons_manager: this.completionThemeManager,
```
bollwyvl, line 167: `symbol_manager`?

packages/jupyterlab-lsp/src/features/completion/completion_handler.ts
```diff
@@ -277,6 +286,16 @@ export class LSPConnector
     };
   }
 
+  protected icon_for(type: string): LabIcon {
+    if (!this.options.settings.composite.theme) {
+      return undefined;
+    }
+    if (typeof type === 'undefined' || type == '<unknown>') {
+      type = 'Kernel';
```
bollwyvl, line 294: hoist some of these magic strings

packages/jupyterlab-lsp/src/features/completion/completion_handler.ts
```diff
@@ -291,18 +310,15 @@ export class LSPConnector
...
+          icon: this.icon_for(item.type as string)
         };
       });
     } else {
       items = reply.matches.map(match => {
         return {
           label: match,
           insertText: match,
-          icon: kernelIcon
+          icon: this.icon_for('Kernel')
```
bollwyvl, line 321: hoist magic string

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -223,6 +226,9 @@ const plugins: JupyterFrontEndPlugin<any>[] = [
   FILE_EDITOR_ADAPTER,
   VIRTUAL_EDITOR_MANAGER,
   CODEMIRROR_VIRTUAL_EDITOR,
+  COMPLETION_THEME_MANAGER,
```
bollwyvl, line 229: i assume these won't be hardcoded here later?

packages/completion-theme/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+export const COMPLETION_THEME_MANAGER: JupyterFrontEndPlugin<ILSPCompletionThemeManager> = {
+  id: PLUGIN_ID,
+  requires: [IThemeManager, ICommandPalette],
+  activate: (
+    app,
+    themeManager: IThemeManager,
+    commandPalette: ICommandPalette
+  ) => {
+    let manager = new CompletionThemeManager(themeManager);
+    app.commands.addCommand('lsp-completer-about-themes', {
```
bollwyvl, line 161: prefer a prefixed thing, e.g. `lsp-symbols:show-about`

packages/completion-theme/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+      label: 'Display the completer themes',
+      execute: () => {
+        manager.display_themes();
```
bollwyvl, line 164: as suggested, hoist to dedicated widget

packages/completion-theme/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  protected is_theme_light() {
+    const current = this.themeManager.theme;
+    if (!current) {
+      // assume true by default
+      return true;
+    }
+    return this.themeManager.isLight(current);
+  }
+
+  get_icons_set(
```
bollwyvl, line 40: `get` and `set` in the same function name is a little confusing...

packages/completion-theme/src/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  private current_theme_id: string;
+  private icons_cache: Map<string, LabIcon>;
+
+  constructor(protected themeManager: IThemeManager) {
+    this.themes = new Map();
+    this.icons_cache = new Map();
+    themeManager.themeChanged.connect(this.update_icons_set, this);
+  }
+
+  protected is_theme_light() {
```
bollwyvl, line 31: i guess the whole point of the LabIcon trainwreck was to _not_ make us worry about light/dark, by delegating to the `jp-icon-*` class names... ideally wouldn't have to have this...

bollwyvl:
Semi-functional prototype:
Screenshot from 2020-08-27 14-18-46

packages/jupyterlab-lsp/src/index.ts
krassowski: Sure!

krassowski:
> Semi-functional prototype:
> Screenshot from 2020-08-27 14-18-46

I thought about such alignment to. The disadvantage is that the user would need to scroll horizontally if they have more themes.

krassowski:
> Some issue with the symbol kinds of non-selected items in vscode theme:
> 
> vscode	material	null
> Screenshot from 2020-08-27 09-25-00	Screenshot from 2020-08-27 09-24-26	Screenshot from 2020-08-27 09-27-46

It's a feature, not a bug. The themes evolved from `iconThemes` to `completerThemes`. The VSCode theme not only provides the VSCode icons, but also replicates some of the VSCode style/behaviour experience.

krassowski:
> copy:
> 
> * probably prefer "Language Server" to "LSP" acronym in layed out-content
> 
> package names:
> 
> * probably _should_ have lsp in them
> * probably _isn't_ specific to completion
>   
>   * whenever we get around to symbol list/tree, we'll want these, too, so they need to be available by some other means...
> * naming might help, as they aren't jupyterlab themes (just icons) and don't work without the rest of the kit
>   
>   * `lsp-symbol-icons`
>   * `lsp-symbol-icons-material`
>   * `lsp-symbol-icons-vscode`

I am against adding lsp because we will become `@jupyterlab-lsp` and the package names will remain `@jupyterlab-lsp/theme-vscode`. The themes will be able to style other components as well, requesting appropriate tokens but will NOT be limited to icons alone (see the vscode completer theme), therefore I am also against renaming them to symbols (they are meant to be proper themes for the LSP components).

krassowski:
To be clear, anyone who develops a third-party theme should probably name the package `lsp-theme-x` indeed. Just would be redundant for us given it will be in the org name (same as jupyterlab does not prefix packages with jupyterlab).

krassowski:
> ...maybe pull through the vs colors into the material theme...
> i don't know if people have actually pinned those things in their brains
> functions are purple, darnit, but hands off my gold enums!
> but it's a little more interesting, perhaps
> luckily, it isn't like there's a green block and a red block that mean drastically different things
> ...and/or offer an icon_colors config option
> colorful to ape the vs colors
> monochrome (the existing grey)

Yeah, I though about it too. I am happy to either have this enabled by the `completerThemeManager`, or for individual themes to implement their colouring scheme via CSS. We cannot match the previous experience of everyone because the colors change between kernels (they are sequentially generated to a list of types returned by the kernel) and as so do not have a fixed meaning (especially in multi-language setting). I was thinking that the settings (color on/off) might belong to the theme, but of course, if implemented at the completerThemeManager icon level, it would belong over there. In my view this would be another PR though.

packages/jupyterlab-lsp/schema/completion.json
krassowski: I think it should stay this way. See my comments about the LSP theme packages not being about icons alone, nor about completion alone.

A theme package will provide themes for specific features/components (because each feature is modular and independent of each other) and features which support the themeManager pattern can opt in to allow to change the theme/look-and-feel of the specific widgets. One can like the appearance of the completer from VSCode, but prefer the hover box to use material-based, or PyCharm-emulating theme.

packages/completion-theme/src/index.ts
krassowski: We could rename IconSet to IconPack, but `set` seem to fit better though...

packages/completion-theme/src/index.ts
krassowski: I don't think it is possible with LabIcon (especially for two completely different variants of icons, see light/dark version for VSCode theme - which are much different from "let's change a contrast a bit" approach in Lab).

packages/jupyterlab-lsp/schema/completion.json
bollwyvl: i think for the specific case of icons, it's _way_ more confusing to have multiple, simultaneous icons for the same concept in different places.

packages/completion-theme/src/index.ts
bollwyvl: then call it an `get_iconset` perhaps?

packages/jupyterlab-lsp/schema/completion.json
krassowski: I agree, but just to re-iterate it is not about icons only, its about the ability to customise the LSP components to a much greater extent. I am staring with icons and presentation of the completion list, but there is much more to it. For example, in the completer I want to allow the themes to place the documentation in different places (below the suggestion rather than next to it); diagnostics themes would have different icons for severity but also different highlight mechanisms (underlines/wavy/color sets).

Also, the modular design reflects the core approach, for example with the terminal (and other core extensions) having separate themes and the overall "meta-themes" agggregating this to provide quick-switch experience.

packages/completion-theme/src/index.ts
bollwyvl: >  which are much different from "let's change a contrast a bit" 

yeah, would have to go through and classes to things, for sure. but right now: if a user had a purplish base  theme (have seen it), a lot of the icons would disappear. 

> light/dark version for VSCode theme

it's all just svg under the hood: typically i've just transformed the raw, incoming SVG, swapping out the icon classes for different things to get different flavors. anyhow, i care less about exactly color matching some other experience if it falls apart under certain, easily observed conditions. 

heck, that might be how icon color schemes are actually implemented, e.g. a map of `[Kind]=>'jp-icon-*'`.

bollwyvl:
> via CSS. We cannot match the previous experience of everyone 

yeah, i wouldn't worry about trying to recreate (or even particulary document) the behavior the kernel has. theoretically, everything should _just work_ even if this theme thing _isn't_ there, which means we should do as little as possible to the default behavior.

>  The themes will be able to style other components as well, requesting appropriate tokens but will NOT be limited to icons alone (see the vscode completer theme), therefore I am also against renaming them to symbols (they are meant to be proper themes for the LSP components).

ok, didn't realise the scope of this extended beyond icons, or had planned for completion. I just don't want us to end up having `foobar-theme` and `hover-theme` and... anyhow, so perhaps `code-theme(-*)`, then? do you in fact intend it to be optional in the downstream components?

packages/completion-theme/src/index.ts
krassowski: I like the direction of your "semi-functional" prototype (though we need the licence and extra data there). I would like to also include some space for extra settings for each of the themes (e.g. "only show the type of highlighted entry" and "show the documentation underneath the active item" for vscode theme). Should I work on my own widget implementation (i.e. was your effort more of a mock-up or do you want to PR?)

bollwyvl:
>  The disadvantage is that the user would need to scroll horizontally if they have more themes.

ha: can worry about that happy problem when it arises, i'd say. with a little jiggery-pokery, that layout could show 10 themes. 

> However, it should provide the information about licence and presence 

oh for sure, hence "barely working prototype" or whatever i said. should have an actual _link_ as well, too. 

> of dedicated dark-mode set.

as mentioned, would prefer the use of the theme-aware semantic labicon classes vs advertising sub-variants... if an extension really wants to do something crazy, they can just make another theme.

krassowski:
> ok, didn't realise the scope of this extended beyond icons, or had planned for completion. I just don't want us to end up having foobar-theme and hover-theme and... anyhow, so perhaps code-theme(-*), then? do you in fact intend it to be optional in the downstream components?

So my idea is that each feature may provide a theme manager. A theme will then accept tokens of these feature-specific theme managers (not `require` but `optional`) and register itself with one or more such managers.

bollwyvl:
> So my idea is that each feature may provide a theme manager.

then i'm doubling down own `symbol-icons` vs `completion-icons` + `tree-icons` + `hover-icons` and making each theme manager beholden to the same set of symbols: this is case where sacrificing configurability for cross-feature consistency is pretty important. if that means they need to get hoisted even higher, so be it.

krassowski:
> as mentioned, would prefer the use of the theme-aware semantic labicon classes vs advertising sub-variants... if an extension really wants to do something crazy, they can just make another theme.

How do I implement the dark mode VSCode icons with labicon classes via interface then? I see the inconvenience of having to manage theme switching but if there is a better way (which retains the advantage of interface for assurance that all needed icons are supplied and there are not typos) I just cannot see it.

bollwyvl:
> How do I implement the dark mode VSCode icons with labicon classes via interface then? 

I don't know what "via interface" means (it's just CSS classes, nothing programmatic), but:

- dark `method`
```xml
<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M13.51 4L8.51001 1H7.51001L2.51001 4L2.02002 4.85999V10.86L2.51001 11.71L7.51001 14.71H8.51001L13.51 11.71L14 10.86V4.85999L13.51 4ZM7.51001 13.5601L3.01001 10.86V5.69995L7.51001 8.15002V13.5601ZM3.27002 4.69995L8.01001 1.85999L12.75 4.69995L8.01001 7.29004L3.27002 4.69995ZM13.01 10.86L8.51001 13.5601V8.15002L13.01 5.69995V10.86Z" fill="#B180D7"/>
</svg>
```

- labicon would require the following change:
```xml
<path ... fill="#B180D7" class="jp-icon-contrast0 jp-icon-selectable"/>
```

yeah, the purple will not be exactly `#b180d7`, but the intent will be preserved, and it won't need multiple files.

if we do go down the "icon color scheme" path: i would imagine defaulting the svg to the most boring class e.g. `jp-icon-3` and then hot-swapping the color with regexen would answer the mail.

packages/completion-theme/src/index.ts
bollwyvl: > space for extra settings for each of the theme

Welp, as a user, i didn't expect changing the icons to have anything to do with hiding parts of my ui. Sample size n=1, but...  i feel like other switches like that would be hoisted to other config widgets, akin to the color scheme, opacity, and other junk we've mentioned. making configuration itself _too_ configurable means we'd lose much of the value of the settings system doing validation for us.

Downstream theme consumers should have no knowledge of _which_ theme they are using, or the whole house of cards falls down, so it sounds like the completion theme has to know about all the possible things one might configure.

So an over-arching `Use Code Theme: VSCode` might be a command that updates a bunch of settings

```yaml
# at the global level
{"symbol-icons": "vscode"}
# at the completion level 
{"only_show_active_suggestion_kind": true, "docs_under_active_suggestion": true}
```
?

> was your effort more of a mock-up or do you want to PR?

ehm, it's semi-decent, but burned much of your stuff to the ground (because dialog)... i can get it worked up, writing to the settings, etc.. but likely would be the weekend.

packages/jupyterlab-lsp/schema/completion.json
bollwyvl: Ok, if the focus has shifted, title of the PR is still "completer icons". As suggested earlier: a new docs diagram of how you see this going down, and what _this_ PR is achieving of that vision would probably help... and this sounds complicated enough that for anyone else to follow what they have to do, we'll need it anyway.

krassowski:
But would not this meant that only one color can be supported per icon? By interface I mean:

```typsecript
type requiredIcons = {
  [key in CompletionItemKindStrings]: SvgString;
};

export interface ICompletionIconSet extends requiredIcons {
  Kernel?: SvgString;
}
```

which guarantees that all the icons will be covered. If the swapping of colors for the dark mode is implemented in CSS then I do not see an easy way to enforce that all icons are provided with dark theme equivalents.

bollwyvl:
>  only one color can be supported per icon

all the current ones are monochrome anyway, no? 16px is _really small_, anyway.

> with dark theme equivalents.

There would _be_ no xxx mode "equivalent", because:

- themes are responsible for _implementing_ the correct semantic classes that look _good_ 
- icons are responsible for _using_ the semantic classes, e.g. `jp-icon-focus`, `-warn`, etc

krassowski:
Well, not the VSCode ones. They have up to two colors (although I think that we do not use these for completer).

bollwyvl:
> They have up to two colors

sry, man, just going off what i see on this pr... but i don't see it as a huge loss. again, 16px.

krassowski:
> The hint area should probably be overflow-y: auto instead of overflow: scroll... the list of completer items is probably fine as scroll, however.

Agree the scrollbars which are not used for most of the time are a tad ridiculous. This sounds like a core PR, but we can also apply the override here at no cost (I am thinking `.jp-Completer-docpanel {overflow; auto}`).

> It would also be rad if the hint area was syntax highlighted... having a persistent codemirror item that can be be lit up would be great, especially if those tokens were clickable, e.g. to be able to jump to the reference of the return type of a method...

Needs a change upstream (JupyterLab), currently, the interface accepts strings only. We would need to pass string + the mimetype. The added complexity is because the documentation string sometimes mixes up code (signature) and prose. Ideally, it would be only markdown and signature would be provided in a different field (but this requires upstream changes to LSP protocol).

bollwyvl:
> the interface accepts strings only. 

eek, yeah that's gross... should be a mimebundle.

> changes to LSP protocol

I wouldn't hold my breath... but perhaps a call to signature help or something could be used?

packages/completion-theme/src/index.ts
krassowski: > So an over-arching Use Code Theme: VSCode might be a command that updates a bunch of settings

I was thinking that we could keep it to a minimum and keep the settings in the actual theme plugins. The only thing that we would change is the ICompletionTheme registration interface. It would now include 'settings: string[]' with the identifiers of settings defined by the appropriate JSON schema file. The interface would now take the labels and type of accepted values from the setting registry and display an appropriate widget (let's say we only support a true/false switch for now).

krassowski:
So in order to:
- move the documentation below the completion item
- add background color reusing the lab implementation

we would need something like this:

```typescript
export interface ICompletionRenderOptions {
  /**
   * Whether the background should be rendered.
   */
  renderBackground: boolean;
  /**
   * Whether the documentation data should be added.
   */
  addDocumentationData: boolean;
}

/**
 * The number of colors defined for the completer type annotations.
 * These should be in sync with upstream in completer/src/widget.ts.
 */
const N_COLORS = 10;

class LSPRenderer extends Completer.Renderer {

  constructor(protected options: ICompletionRenderOptions) {
    super();
  }

  createCompletionItemNode(
    item: CompletionHandler.ICompletionItem,
    orderedTypes: string[]
  ): HTMLLIElement {

    let node = super.createCompletionItemNode(item, orderedTypes);
    // allow themes to display the documentation below the item (overriding the default documentation panel)
    if (this.options.addDocumentationData) {
      node.setAttribute('data-documentation', item.documentation)
    }
    if (this.options.renderBackground) {
      let typeNode = node.querySelector('.jp-Completer-type')
      if (!typeNode.hasAttribute('data-color-index')) {
        const colorIndex = (orderedTypes.indexOf(item.type) % N_COLORS) + 1;
        typeNode.setAttribute(`data-color-index`, colorIndex.toString());
      }
    }
    return node;
  }
}
```

But this depends on https://github.com/jupyterlab/jupyterlab/issues/8926 as it is not possible to change renderers at the moment.

krassowski:
I propose that we merge this as is. I am happy to subsequently accept PRs which:
- change the handling of the dark mode icons (assuming all colours for VScode theme are preserved)
- replace the themes dialog with an area widget (or I could do a minimal effort, but a fully functional one myself if you do not have time for it @bollwyvl)

bollwyvl:
presently taking a look at the things mentioned above...

bollwyvl:
Woof... rebuilt my env, and now jedi is exploding, so I can't even really get many completions. BAH!

bollwyvl:
I did get everything back up and running on the config widget. Branch is up, but not quite PR-ready. 

Things I changed (sry, no screens, on different box):
- done
  - [x] moved the config widget next to the completion feature in `jlsp`
    - reduces duplication of the category name for command palette
    - also moved the style
  - [x] got the immediate round-trip to work for `theme`
    - added `get_current_theme_id` and a `current_theme_changed` signal to the IconThemeManager
    - the completion plugin listens for this
  - [x] added a command for _Set Completion Icon Theme`, starting the `CommandIds` pattern
  - [ ] add command palette items for themes as registered 
  - [x] added the license, pretty name. 
    - i wonder about making the ids `redmond` and `mountainview` :smiling_imp: 
- ongoing
  - [ ] add dummy themes for `kernel` and `null`
  - [ ] thinking harder about the layout... 
    - it would actually be simpler if it was a wide table, rather than long.
    - or could be "cards" that wrap, though would lose the comparison advantage of tables, and hover/select is kinda gross 
    - this will force thinking about scroll sooner... since we _know_ how big the icons are, it's a little easier
  - [ ] started looking at the icon color schemes
    - will also get the theme manager dependency out
    - started a notebook to find/replace the colors based on color scheme... regarding two-tone, we could have "base" and (optional) "active" icon classes

krassowski:
Thank you for the update @bollwyvl, I am excited about these improvements! I am merging this one in one hour into `features_refactor` branch (and would love to merge #319 now and #321 later today once ready) so that I can resolve the conflicts and iterate on some loose ends to finish off with a beta/RC tonight. Please target `features_refactor` with the subsequent PRs. It would be great to hear from you on #319 soon (but if you are AFK I can address any comments later on - this is NOT a merge into master yet).

PR #323, opened 2020-08-26
Try to use the most recent build of tectonic again

krassowski:
## References

Follows up #320 and https://github.com/conda-forge/tectonic-feedstock/issues/21#issuecomment-680309037.


krassowski:
Thanks to some magic I do not fully comprehend (marking builds as faulty upstream?), all systems now install tectonic after removing the pin, resolving to build 0. Windows now resolves to:

tectonic                  0.1.12               h11def9f_0    conda-forge

Linux to:

tectonic                  0.1.12               hbc1ed39_0    conda-forge

which works for now...

bollwyvl:
Hooray (I guess)! If this comes back up, we can take a look at doing more
conditional environment and test stuff... Also, hopefully, my smoke test pr
(which the maintainer is now extending to run the full upstream test suite)
will help ensure that broken releases don't go out... finding upstream
issues might end up being one of the more important and lasting things we
do, as it benefits people that aren't even using our code.

On some other work I'm looking into things like conda-lock and conda-pack,
which should be much more durable... Once we have solves that work, and a
place to put them. Lock files are good and easy to check in, while whole
packed envs are... Harder. But since it can pack everything, we might be
able to sneak in things like the tectonic and Jedi caches... Basically
everything short of our packages and built lab.

PR #326, opened 2020-08-29
Add sql-language-server

### LINKED ISSUE #325: Add support for an SQL langauge server
- [ ] decide which server to use
  - https://github.com/joe-re/sql-language-server seems good first choice as it is an npm package; it is not clear which features it supports (but clearly it does support diagnostics and completion),
  - https://github.com/lighttiger2505/sqls is a nice alternative, but written in go and might slow down our CI; also it does not support diagnostics yet,
  - https://github.com/microsoft/vscode-mssql/ is a no go as it supports a single dialect only.
- [ ] allow the extractors to pass language-specific arguments to the connection initialization params see https://github.com/krassowski/jupyterlab-lsp/pull/321#issuecomment-683190128

krassowski:
## References

Closes #325 . Provides a default connection configuration using an empty in-memory SQLite3 database so that it works straight away. It's simple on features (completion + diagnostics) and the diagnostics do not really work on Node 10 as code uses features only implemented in Node 11 (see https://github.com/joe-re/sql-language-server/issues/56), but we might be fine bumping the node requirement soon I think.

## Code changes

None

## User-facing changes

None

None

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
Speaking of node, it seems that docs do not like the version bump of jest:

```
error jest@26.4.2: The engine "node" is incompatible with this module. Expected version ">= 10.14.2". Got "10.13.0"
```

krassowski:
And installation notebook is outdated saying that Node >= 8 is fine.

krassowski:
Jest has a weird dependency update cycle... they say that they dropped Node 8 in 26.0 but then they require 10.X in point releases and then they downgrade it (https://github.com/facebook/jest/pull/10352) to 10.13, but only in the worker, so here we are.

py_src/jupyter_lsp/specs/__init__.py
```diff
@@ -1,11 +1,13 @@
 """ default specs
 """
 # flake8: noqa: F401
+
 from .bash_language_server import BashLanguageServer
 from .dockerfile_language_server_nodejs import DockerfileLanguageServerNodeJS
 from .javascript_typescript_langserver import JavascriptTypescriptLanguageServer
 from .pyls import PythonLanguageServer
 from .r_languageserver import RLanguageServer
+from .sql_langserver import SQLLanguageServer
```
bollwyvl, line 10: maybe make `sql_language_server` for consistency....

bollwyvl:
Modulo the one comment regarding the spec module name, which is not a blocker, LGTM. 

Heroic work on the RTD node stuff :clap:  :clap:  :clap: ... can't wait until Lab 3.0 makes it less of a life-ruining end-user experience, though of course we're rather wedded to it... for now. Before the next release, we should have a dedicated burn-down-the-yarn-lock party and figure out what is broken-by-default-unless-in-our-magic-garden.

Potential follow-ons: 
- fancy handling of the third-party magic seems like a separate PR, unless you've already snuck that in someplace.
- a heavier-duty config test against e.g. postgres, a real `.sqlite` file, might be good in the future, as I'm a little unclear on where the [config](https://github.com/joe-re/sql-language-server#using-configuration-files) stuff needs to go. 
- explore if we can make it demo nicely with [xeus-sqlite](https://github.com/jupyter-xeus/xeus-sqlite) (which unfortunately can't be installed next to xeus-python for now)

py_src/jupyter_lsp/specs/__init__.py
krassowski: I was also thinking whether we need to have nodejs prefix on `DockerfileLanguageServerNodeJS` etc. I guess that it might have been used to distinguish potential alternative implementations, right?

krassowski: And more user-facing, we could change the display name of `vscode-html-languageserver-bin` to (at least) remove `-bin` suffix.

krassowski:
Interestingly, the tectonic Zlib error issue broke Windows 3.8 here (but the two other Windows builds proceeded normally)... So this is because it installed h4f32bc4_3 (bot other Windows runs used h11def9f_0). I feel i understand the way conda/mambda resolves dependencies less and less... (though this is still an upstream bug...)

PR #328, opened 2020-08-30
[wip] Add completion configurer

bollwyvl:
## References

- start of #196
- follow on from https://github.com/krassowski/jupyterlab-lsp/pull/322#issuecomment-683313754

## Code changes

  - [x] moved the config widget next to the completion feature in `jlsp`
    - reduces duplication of the category name for command palette
    - also moved the style
  - [x] got the immediate round-trip to work for `theme`
    - added `get_current_theme_id` and a `current_theme_changed` signal to the IconThemeManager
    - the completion plugin listens for this
  - [x] added a command for _Set Completion Icon Theme`, starting the `CommandIds` pattern
  - [x] add command palette items for themes as registered 
  - [x] added the license, pretty name. 
    - i wonder about making the ids `redmond` and `mountainview` :smiling_imp: 
    - [x] add `modifications?` to theme definition for licensing compliance
  - [x] add dummy themes for `kernel` and `null`
  - [x] thinking harder about the layout... 
    - ~~it would actually be simpler if it was a wide table, rather than long.~~
    - or could be "cards" that wrap, though would lose the comparison advantage of tables, and hover/select is kinda gross 
    - this will force thinking about scroll sooner... since we _know_ how big the icons are, it's a little easier
  - [x] started looking at the icon color schemes
    - will also get the theme manager dependency out
    - started a notebook to find/replace the colors based on color scheme... regarding two-tone, we could have "base" and (optional) "accent" icon classes
  - [x] [hoist color scheme implementation ](https://github.com/krassowski/jupyterlab-lsp/pull/328#discussion_r479797965)

## User-facing changes

Instead of a dialog, presents a main area widget, and offers appropriately interactive direct manipulations of all of the items (no save button)

Screenshot from 2020-08-30 10-36-47

A number of things changed... 
- making things-we-might-have-n-of asynchronous
- replacing some ILabIcon with LabIcon (they aren't really compatible)

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

## Discussion

While I actually _enjoy_ this old school kind of work, and tweaking how every pixel looks, doing this per feature is going to be insane, especially once we add per language overrides, etc. It might be fine to iterate on, test, and land this so we have _something_ to look at, and then refactor with RJSF. We would be able to refactor the existing "novel" things (e.g. the icon preview, the token finder) work, and make them stuff that shows up in our `uiSchema`, so we'd get the best of both worlds.


krassowski:
Well if we are adding continuous hinting and documentation here, we indeed may start looking into the other features as well. My rationale of implementing the themes preview at all was only to gather the icons in one place and display the licence and wait for the upstream effort of bringing configuration GUI. I am afraid that if we go too far with the settings configuration we may end up having two GUIs for that...

krassowski:
Though yeah, the layout for icons is looking nice. Note: "suppress in" is not only about continuous hinting (after each keystroke) but also about auto hinting after trigger characters (which are provided by the LSP server).

krassowski:
So it could be:

"Hinting"
  - enable continuous
  - suppress auto-invoke in

packages/completion-theme/src/types.ts
```diff
@@ -51,21 +51,21 @@ export interface ICompletionIconSet extends requiredIcons {
 
 export interface ILicenseInfo {
   /**
-   * Licence name.
+   * License name.
    */
   name: string;
   /**
-   * Abbreviation of the licence name;
+   * SDPX identifer of the license.
```
krassowski, line 58: Good, I did not know about SPDX. Maybe would be good to keep "(abbreviation)" or provide an example.

packages/theme-vscode/style/icons/file.svg
```diff
@@ -1,3 +1,3 @@
 <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path fill-rule="evenodd" clip-rule="evenodd" d="M10.57 1.14L13.85 4.44L14 4.8V14.5L13.5 15H2.5L2 14.5V1.5L2.5 1H10.22L10.57 1.14ZM10 5H13L10 2V5ZM3 2V14H13V6H9.5L9 5.5V2H3ZM11 7H5V8H11V7ZM5 9H11V10H5V9ZM11 11H5V12H11V11Z" fill="#424242"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M10.57 1.14L13.85 4.44L14 4.8V14.5L13.5 15H2.5L2 14.5V1.5L2.5 1H10.22L10.57 1.14ZM10 5H13L10 2V5ZM3 2V14H13V6H9.5L9 5.5V2H3ZM11 7H5V8H11V7ZM5 9H11V10H5V9ZM11 11H5V12H11V11Z" fill="#C5C5C5" class="jp-icon3" />
```
krassowski, line 2: I went a bit conservative route of not touching these files as they already have fill (which is all that is needed for the labicon system to work); not sure if touching this has any licence implications, but if it has not then we could also remove the `fill="none"` from the svg tag (and maybe the fill rules?).

packages/theme-vscode/style/icons/symbol-variable.svg
```diff
@@ -1,3 +1,3 @@
 <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path fill-rule="evenodd" clip-rule="evenodd" d="M2 5H4V4H1.5L1 4.5V12.5L1.5 13H4V12H2V5ZM14.5 4H12V5H14V12H12V13H14.5L15 12.5V4.5L14.5 4ZM11.76 6.56995L12 7V9.51001L11.7 9.95996L7.19995 11.96H6.73999L4.23999 10.46L4 10.03V7.53003L4.30005 7.06995L8.80005 5.06995H9.26001L11.76 6.56995ZM5 9.70996L6.5 10.61V9.28003L5 8.38V9.70996ZM5.57996 7.56006L7.03003 8.43005L10.42 6.93005L8.96997 6.06006L5.57996 7.56006ZM7.53003 10.73L11.03 9.17004V7.77002L7.53003 9.31995V10.73Z" fill="#75BEFF"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M2 5H4V4H1.5L1 4.5V12.5L1.5 13H4V12H2V5ZM14.5 4H12V5H14V12H12V13H14.5L15 12.5V4.5L14.5 4ZM11.76 6.56995L12 7V9.51001L11.7 9.95996L7.19995 11.96H6.73999L4.23999 10.46L4 10.03V7.53003L4.30005 7.06995L8.80005 5.06995H9.26001L11.76 6.56995ZM5 9.70996L6.5 10.61V9.28003L5 8.38V9.70996ZM5.57996 7.56006L7.03003 8.43005L10.42 6.93005L8.96997 6.06006L5.57996 7.56006ZM7.53003 10.73L11.03 9.17004V7.77002L7.53003 9.31995V10.73Z" fill="#75BEFF" class="jp-icon-brand1" />
```
krassowski, line 2: I would avoid going this route. This is certainly a modification (needs to check license) and there are plans to change these variables, see https://github.com/jupyterlab/jupyterlab/issues/8832#issuecomment-683198582. I would create a set of dedicated variables fully reflecting the exact colors as shipped by MS.

bollwyvl:
> we indeed may start looking into the other features as well... may end up having two GUIs for that...

Gotta have one first, and starting with a complicated/rich one is as good as any.

packages/completion-theme/src/types.ts
bollwyvl: spdx is really google-able... if anything would put a link to the list:
https://spdx.org/licenses/

packages/theme-vscode/style/icons/file.svg
bollwyvl: CC-BY-4.0 allows for modification. fill doesn't do anything on an svg itself.

packages/theme-vscode/style/icons/symbol-variable.svg
bollwyvl: If the upstream variable names change, we'll update them, and it will work for all themes. If we make new variables, every theme will have to (read: won't) adapt to them. The primary motivating case: if a theme designer decides they need the vscode purple as their background, all the purple icons would disappear. However, as that purple shows up in, for example, the "new markdown file" icon, they'll realize that they have to own _the rest_ of the colors as well, and create an overall harmonious experience: we aren't building vscode. 

If we _do_ (I don't) want to build something that beat-matches vscode, it would be a theme, and _could_ define `jp-icon-whatever` to be _exactly_ that color.

packages/completion-theme/src/types.ts
bollwyvl: 6ef20ca

packages/theme-vscode/style/icons/file.svg
krassowski: Well the fill on svg tag seems to be allowed (https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg) but IMO is not needed. As for the fill on path it needed for the lab icons see https://github.com/jupyterlab/jupyterlab/issues/8413.

krassowski: Yes, CC-BY-4.0 allows to modify but requires to clearly state all modifications (see [s3(a)(1)](https://creativecommons.org/licenses/by-sa/4.0/legalcode#s3a1)). However, as long as we only modify the technical bits and not the essence (e.g. change how the colors are encoded or how we set them but not the colors themselves) we are not creating an an Adapter Material per [s2(a)(4)](https://creativecommons.org/licenses/by-sa/4.0/legalcode#s2a4) and do not need to state the modifications along the licence.

bollwyvl:
> "Hinting"
> - enable continuous
> - suppress auto-invoke in

Didn't dig too much into these... so does continuous affect suppress in any way?

packages/theme-vscode/style/icons/file.svg
bollwyvl: Adding a `modifications?` member to the theme interface seems reasonable, and we can show it here.

We can add an `unthemed` option which removes the classes altogether, and just uses what's in the svg. Still feel strongly we'd want `themed` to be the default, as it encourages good practice for the ecosystem.

Looking across the core icons, they all have _something_ in `fill`... as it's really hard to edit something you can't see.

bollwyvl: Actually, regarding that: the `colorScheme` thing is presently really ugly (magic strings, etc). I should hoist them to the theme manager, e.g. 
```ts
mgr.register_color_scheme({
  id: 'unthemed',
  title: 'Unthemed',
  description: 'Use raw colors from SVG (may not be compatible with all themes)',
  transform: (svg) => svg.replace(mgr.ICON_CLASS_REGEX, '')
});

mgr.register_color_scheme({
  id: 'muted',
  title: 'Muted',
  description: 'Use muted default theme icon color',
  transform: (svg) => svg.replace(mgr.ICON_CLASS_REGEX, 'jp-icon4')
});
```

We would still document them in the schema, but it would be harder to validate, conclusively, but that's probably ok.

krassowski: I meant that while we are editing the files, we can remove the _other_ fill. This one:

`<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">`

Specifically this bit is not needed:

```
fill="none" xmlns="http://www.w3.org/2000/svg"
```

The xmlns ca be removed because:

> Note: The xmlns attribute is only required on the outermost svg element of SVG documents. It is unnecessary for inner svg elements or inside HTML documents. (https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg)

bollwyvl: > The xmlns ca be removed because:

My recollection is that it actually gets parsed as XML (not just `innerHTML`ed) when becoming a `LabIcon`... pretty sure it fails if `xmlns` is missing. 

My initial thing was purely additive... there's too many icons to worry  _too_ much about bits and bytes. Making the loading of the actual SVG strings was more important, from a size perspective... that was part of what lead me to get rid of the comparison table, as loading that would load/parse every theme.

That being said, it _is_ important to get rid of all `id`s, as that leads to all kinds of haywire crap, since they all go into the same DOM. Icons built for the web _usually_ don't do that, though.

Another thing: adding classes to `body` gives me the willies... maybe we set a `data-` tag on the app node? So we'd have `data-lsp-completion-theme="material"`, selected with `[data-lsp-completion-theme="theme-id"] .lsp-whatever`. This helps ensure that multiple themes don't stomp on each other if some cleanup doesn't work correctly.

bollwyvl:
Some progress:

Screenshot from 2020-08-31 22-52-49

bollwyvl:
Palette items:

Screenshot from 2020-08-31 23-38-46

bollwyvl:
Handling null:

Screenshot from 2020-09-01 00-03-27

bollwyvl:
Ok, I've made all the changes i wanted to do... losing steam to fire up the :robot: this evening, however :zzz: hopefully i can do the few tests i want tomorrow: really just need to validate that the settings update, and the bits appear on the form, as the completer tests should take care of the rest (though I may have broken something).

krassowski:
Should the colour scheme part get greyed out when null theme/no icons is set?

bollwyvl:
> Should the colour scheme part get greyed out when null theme/no icons is
> set?
>
While we could add it cosmetically, this might be a good one to defer until
we have rjsf, as it can support conditional schema fairly easily...

>

bollwyvl:
Hopefully unbroke the _existing_ tests... will proceed tomorrow/this weekend with the new ones...

bollwyvl:
Also, newest `black` went ape on our python... i generally like it, but we should probably resolve that on a separate PR and/or do the yarn.lock refresh...

bollwyvl:
Not sure what's going on in windows... might need to try some stuff like https://www.crummy.com/software/BeautifulSoup/bs4/doc/#inconsistent-encodings

atest/logcheck.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+from bs4 import UnicodeDammit
+
+
+def file_should_not_contain_phrases(filename, offset=0, *phrases):
+    """don't fail _too_ hard if the file can't be read for soem reason"""
```
krassowski, line 7:
```suggestion
    """don't fail _too_ hard if the file can't be read for some reason"""
```

krassowski:
Looks better now! The failure at the moment is:

`NotADirectoryError: [WinError 267] The directory name is invalid: 'D:\\a\\1\\s\\atest\\output\\windows_37_4\\user-settings\\@krassowski\\plugin id=@krassowski\\jupyterlab-lsp:completion'`

which I guess might because `Configure JupyterLab Plugin` takes `plugin id`, while `Reset Plugin Settings` takes `plugin` (but is being called with `plugin id`).

bollwyvl:
Indeed, that would break things! I'l look at normalizing how we work with/clean up the settings ids/files...

bollwyvl:
ok, affb602 and some follow-on should have us back up on robot... we should be better set up to handle more plugins, in the future, too.

bollwyvl:
Just watching the windows build, it has at _least_ gotten past the settings cleanup stuff, as well as some places where the log reading was dying before...

bollwyvl:
So everything's looking good, thus far, but we're running up on the 60 minutes (`Duration: 57m 36s`) with a _single_ fail on the first trip through win38/win37... and I'm about to actually _add_ a few more tests for the new stuff. 

A quick win might be not running the `jest` tests on windows: https://github.com/krassowski/jupyterlab-lsp/issues/311#issuecomment-688951006

bollwyvl:
Oh, and full local run completed.

PR #345, opened 2020-09-08
First attempt to setup GitHub actions

krassowski:
## References

Trying to migrate to GitHub actions #311. It will be trial and error I guess. Not sure If I can do this without pushing to master though...

## User-facing changes

None

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


.github/workflows/job.lint.yml
```diff
@@ -0,0 +1,52 @@
+name: Check lint
+on: [push, pull_request]
```
bollwyvl, line 2: I recommend:
```yaml
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

.github/workflows/job.lint.yml
```diff
@@ -0,0 +1,52 @@
...
+  lint:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-16.04]
+        python: [3.8]
+        nodejs: ['>=12,<13.0.0a0']
+        lab: ['>=2,<3.0.0a0']
+    steps:
+      - name: Set up Python and conda
```
bollwyvl, line 23: There's a cache example that pairs up quite nicely for this.

.github/workflows/job.lint.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: goanpeca/setup-miniconda@v1
+        with:
+          python-version: ${{ matrix.python }}
+          channels: conda-forge, anaconda
+          channel-priority: true
+          mamba-version: "*"
+          activate-environment: jupyterlab-lsp
+          auto-update-conda: false
+
+      - name: testing dependencies
```
bollwyvl, line 33: We'll likely want to revisit this, and generate a _known_ environment, rather than doing the solves multiple times.

.github/workflows/job.test.yml
```diff
@@ -0,0 +1,146 @@
...
+          conda list
+          conda config --show-sources
+          conda config --show
+          printenv | sort
+
+      - name: update conda environment with test dependencies
+        run: conda env update -n base --file requirements/github-actions.yml && conda install jupyterlab"${{ matrix.lab }}" nodejs"${{ matrix.nodejs }}"
+
+      - name: install npm dependencies
+        run: jlpm
```
bollwyvl, line 66: another spot worth caching... this doesn't solve the lab dependency download (c'mon lab 3!)

```yaml

      - name: cache (yarn)
        uses: actions/cache@v1
        with:
          path: .yarn-packages
          key: yarn-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
          restore-keys: |
            yarn-${{ runner.os }}-
            yarn-
```

.github/workflows/job.test.yml
```diff
@@ -0,0 +1,146 @@
...
+        run: python setup.py sdist bdist_wheel
+
+      - name: build npm bundles
+        run: jlpm lerna run bundle
+
+      - name: install python wheel
+        run: cd dist && python -m pip install jupyter_lsp-${{ env.PY_JLSP_VERSION }}-py3-none-any.whl --no-deps
+
+      - run: python scripts/jedi_cache.py
+        name: warm up jedi cache
```
bollwyvl, line 81: This can probably be cached.

.github/workflows/job.test.yml
```diff
@@ -0,0 +1,146 @@
...
+
+      - name: warm up tectonic cache
+        run: python scripts/tectonic_cache.py
```
bollwyvl, line 84: I'd love to know _where_ this was writing... worth investigating and caching...

.github/workflows/job.test.yml
```diff
@@ -0,0 +1,146 @@
...
+
+      #- task: PublishTestResults@2
+      #  name: publish browser test results
+      #  inputs:
+      #    testResultsFiles: atest/output/*.xunit.xml
+      #    testRunTitle: 'Robot ${{ env.name }}${{ python.name }}'
+      #    mergeTestResults: true
+      #  condition: always()
+
+      #- task: PublishPipelineArtifact@0
```
bollwyvl, line 141: Fairly easy to bring back publishing..

```yaml
      - name: publish (linux)
        uses: actions/upload-artifact@v2
        with:
          name: robot ${{ github.run_number }} # needs more of the matrix
          path: ./atest/output
```

krassowski:
@bollwyvl Would you like to take over to make conda update/install work? I am not an advanced conda user and it is getting more trial and error. I removed generation from the template as thought it would be easier, but I am confused on when to create the env ad how (note: I am happy for the generation from the template to be restored, I guess it would safe some run time).

.github/workflows/job.test.yml
```diff
@@ -0,0 +1,146 @@
...
+
+jobs:
+  acceptance:
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }} ${{ matrix.python }}
+    strategy:
+      matrix:
+        python: [3.6, 3.7, 3.8]
+        os: [ubuntu-16.04, macos-10.14, vs2017-win2016]
+        nodejs: ['>=10,<11.0.0.a0', '>=2.2.0,<3.0.0a0', '>=13,<14.0.0a0']
```
bollwyvl, line 38: probably just 
```yaml
nodejs: ['>=10,<11.0.0.a0', '>=13,<14.0.0a0']
```

requirements/github-actions.yml
```diff
@@ -27,7 +22,6 @@ dependencies:
   - chktex
   # test tools
   - pytest-asyncio
-  - pytest-azurepipelines
```
bollwyvl, line 30: Really is too bad... I haven't found anything to replace it yet on github actions (short of a dedicated ci bot that publishes gists)

requirements/github-actions.yml
```diff
@@ -5,11 +5,6 @@ channels:
   - defaults
 
 dependencies:
-  # runtime dependencies
```
bollwyvl, line 8: Sure, that's fine... I'm more and more thinking we want to just drive down the _runtime_ cost of all these env shenanigans... something along the lines of https://github.com/conda-incubator/conda-lock/issues/42

bollwyvl:
Solider on! Let me know if you need any support!

I'm still investigating my issues on #328, but shouldn't be stepping on each others' toes.

krassowski:
Interesting, `##[error]ENOENT: no such file or directory, open '/home/runner/work/jupyterlab-lsp/jupyterlab-lsp/requirements/github-actions.yml'`... I do not get it...

bollwyvl:
> take over to make conda update/install work

Sure, I'll take a look at something...

bollwyvl:
I was able to spend a little time working the conda-lock piece, but it's not ready to push:
- offline, solve all the environments we need based on
  - on the actions matrix
  - the `requirements/*.yml`
- check lock files in (right now in `.github/conda.locks`
- use `setup-miniconda` to handle the "plumbing" (setting shell variables etc) for the env `-n jupyerlab-lsp` (with just e.g. `python`)
- `conda create -n` from the lockfile, derived from the matrix
   - use `mamba` on linux/osx, `conda` on windows
- otherwise, should stay the same

Right now, these are the locks its generating with `mamba` on linux in about three minutes:
```
conda.ubuntu-16.04-3.6-10-2.2.lock
conda.ubuntu-16.04-3.6-13-2.2.lock
conda.ubuntu-16.04-3.7-10-2.2.lock
conda.ubuntu-16.04-3.7-13-2.2.lock
conda.ubuntu-16.04-3.8-10-2.2.lock
conda.ubuntu-16.04-3.8-13-2.2.lock
conda.macos-10.14-3.6-10-2.2.lock
conda.macos-10.14-3.6-13-2.2.lock
conda.macos-10.14-3.7-10-2.2.lock
conda.macos-10.14-3.7-13-2.2.lock
conda.macos-10.14-3.8-10-2.2.lock
conda.macos-10.14-3.8-13-2.2.lock
conda.vs2017-win2016-3.6-10-2.2.lock
conda.vs2017-win2016-3.6-13-2.2.lock
conda.vs2017-win2016-3.7-10-2.2.lock
conda.vs2017-win2016-3.7-13-2.2.lock
conda.vs2017-win2016-3.8-10-2.2.lock
conda.vs2017-win2016-3.8-13-2.2.lock
```

krassowski:
Looking at 12fe40da492a4a7b350de89c78b3a06738ae85d6 to backport it here...

krassowski:
> offline, solve all the environments we need based on

Ideally we would have a GitHub action which would create a PR with the lock files I guess... It could trigger on push and then check if any of the requirements files has changed.

krassowski:
Or, could we just cache the solved environments?

.github/workflows/job.lint.yml
krassowski: On both lint and tests?

.github/workflows/job.lint.yml
krassowski: Done on test workflow.

.github/workflows/job.test.yml
krassowski: Done

.github/workflows/job.test.yml
krassowski: Very different places on different OSes: https://github.com/tectonic-typesetting/tectonic/issues/159#issuecomment-367107341

krassowski:
This seems so cool: https://github.com/marketplace/actions/lint-action (auto fix and annotations!). What do you think @bollwyvl?

bollwyvl:
> GitHub action which would create a PR with the lock files 

I dunno... we don't do this with `yarn.lock` (but could), and there will be less tooling on this. Frankly, what I've liked doing (for both) is refreshing both locks during an "epoch" of development (usually a release) or when something "important" comes up... if it gets too automated, it can be hard to catch what's going on.

> cache the solved environments?

Environments are not generally movable on disk. If they go back to the _exact_ same path, it can work however. [conda-pack](https://github.com/conda/conda-pack) can create an archive which needs one command to run (`conda-unpack`) after uncompressing to get over most of this (and works with more than just conda).

So a run would be like:
- attempt to uncache the pack, based on the entropy of the lockfile
  - if that fails, build and cache the pack
  - if it succeeds do nothing, unpack and install

> linter... auto fix and annotations

We shouldn't _rely_ on anything unique on CI for anything, as then it gets really hard to reproduce locally. if the tool in question can respect our settings and not generate lots of false positives, that's great, and saves us work that we have _decided_ to do. 

I have found that a single-command workflow (e.g. `make`, or `doit` because windows), used locally and on CI, is more effective than investing more in exotic third-party CI-only gizmos... locally, it would just always lint before building and testing, whereas on CI, it could skip the lint if that's not that job's... job.

bollwyvl:
The conda-lock approach has windows down to ~6 minutes conda stuff on windows, and I think i can shave another minute off...

requirements/github-actions.yml
krassowski: We already use it, see https://github.com/krassowski/jupyterlab-lsp/pull/345/files#diff-edeb9086d351b1b16b9627919912df52R88 :)

krassowski:
> conda.vs2017-win2016-3.8-13-2.2.lock

We should use Node 14 if possible, 13 is EOL I think. The CI running right now should tell us if we can bump the version easily.

krassowski:
Looking at finalising this... Any preferences on coverage uploads? I am looking at https://github.com/marketplace/actions/codecov (no annoying comments, just uploads to the third-party servers). I am also happy to upload the xmls as artifacts if you need this.

requirements/github-actions.yml
bollwyvl: great. i'll start up the conda-forge process for it, the lockfile approach can't handle pip deps... yet

bollwyvl: https://github.com/conda-forge/staged-recipes/pull/12620

bollwyvl:
It's fine if you want to land it... i've been tinkering with mine over here, not yet ready to go, and would not merge cleanly with the pip dep:

https://github.com/bollwyvl/jupyterlab-lsp/pull/6

I can PR on top of this, if you keep it open, or we can hit it later: once again, it pulled the new `black`, which inflates the delta some.

> Any preferences on coverage uploads? 

No opinion really: python side of the house is already pinned at 100, so it fails... npm side... well... if you're not failing things based on coverage, then it's just window dressing, and the jest tests aren't using real browsers... so i might not even care.

> We should use Node 14 if possible, 

have started that on above branch, appears to be working, so i'd say all the warnings are just... scary.

krassowski:
Merging as is, looking forward to even faster Windows builds and the conda package for pytest-github-actions-annotate-failures!

PR #348, opened 2020-09-10
[wip] [squash] Add conda locks for CI

bollwyvl:
## References

- continuation of #45
- implements ideas from #311, including `dodo` and `conda-lock`
 
## Code changes

- [x] adds conda-lock
    - [x] checked in locks for all current CI excursions
    - [x] `lock.py` is implemented as a `dodo` task generator
       - could be extended in the future to capture more of the workflow, from checkout to release
- [x] using the locked env above (as show in the docs), ran the new `black`
  - had already encountered this over on #328, so landing it here is as good as anywhere

## Anecdata:
- _absolute_ best case (cold cache on `master`, hot cache on the PR)

### lint: linux py38

| branch        | conda cache | conda | env  | yarn cache | yarn | total |
|---------------|-------------|-------|------|------------|------|-------|
| master        |        0:00 | 3:08  | 1:54 | 0:00       | 1:05 | 9:33  |
| pr            |        0:10 | 1:49  | 0:24 | 0:02       | 1:52 | 5:59  |
| pr (run2)     |        0:06 | 1:31  | 0:25 | 0:01       | 1:58 | 5:46  |

### test: win-64 py3.8

| branch        | conda cache | conda | env  | yarn cache | yarn | total  |
|--------------:|------------:|------:|-----:|-----------:|-----:|-------:|
| master        |        0:00 | 7:38  | 0:00 | 0:00       | 3:36 | 50:17  |
| pr TODO       |        0:00 | 0:52  | 5:52 | 0:04       | 0:00 |  ?:??  |

- the yarn thing is troubling...

> will update with a better windows benchmark set

## User-facing changes

N/A

N/A

## Chores

- [x] linted
- [x] tested
- [x] documented
- [ ] changelog entry


krassowski:
Btw before adding conda cache, conda was about 2:50, so it feels like download of packages was not a bottleneck. I even wondered if having local packages could result in slower times to solve the env., but then I was not using mamba yet (and the lock files were already on horizon).

bollwyvl:
Still need a few more pushes to kick the tires on the `CACHE_EPOCH` thing... but otherwise this is coming together. Should certainly be ready for review come this weekend.

bollwyvl:
Still having issues with the cache keys, especially on windows, which is causing those to mostly fail... still investigating.

.github/workflows/job.test.yml
```diff
@@ -103,18 +106,23 @@ jobs:
...
 
       - name: Build npm bundles
+        shell: bash -l {0}
         run: jlpm lerna run bundle
 
       - name: Install python wheel
+        shell: bash -l {0}
         run: cd dist && python -m pip install jupyter_lsp-${{ env.PY_JLSP_VERSION }}-py3-none-any.whl --no-deps
 
       - name: Cache jedi cache
```
krassowski, line 128: Maybe could just keep bash as default and specify powershell here?

bollwyvl: the issue is bash on windows, as it interacts badly with the cache mechanism and GNU tar... e.g. it thinks `D:\` is a network location or something.

bollwyvl:
Noticed this in the logs:
```
 WARNING in @jupyterlab/coreutils
      Multiple versions of @jupyterlab/coreutils found:
        4.0.2 ./~/@krassowski/jupyterlab-lsp/~/@krassowski/jupyterlab_go_to_definition/~/@jupyterlab/coreutils from ./~/@krassowski/jupyterlab-lsp/~/@krassowski/jupyterlab_go_to_definition/~/@jupyterlab/coreutils/lib/index.js
        4.2.3 ./~/@jupyterlab/coreutils from ./~/@jupyterlab/coreutils/lib/index.js
```

bollwyvl:
This seems like a reasonable change:
```diff
diff --git a/packages/jupyterlab-go-to-definition/package.json b/packages/jupyterlab-go-to-definition/package.json
index a1fd1b7..40d4ef8 100644
--- a/packages/jupyterlab-go-to-definition/package.json
+++ b/packages/jupyterlab-go-to-definition/package.json
@@ -50,7 +50,7 @@
     "@jupyterlab/apputils": "~2.2.0",
     "@jupyterlab/codeeditor": "~2.2.0",
     "@jupyterlab/codemirror": "~2.2.0",
-    "@jupyterlab/coreutils": "~4.0.0",
+    "@jupyterlab/coreutils": "~4.2.0",
     "@jupyterlab/docmanager": "~2.2.0",
     "@jupyterlab/docregistry": "~2.2.0",
     "@jupyterlab/fileeditor": "~2.2.0",

```

.github/workflows/job.test.yml
krassowski: If only cache of jedi cache paths are troublesome on Windows, could we just ignore it and have the benefit of neat consistency of using bash everywhere? I do not that the gain from now warming up jedi cache on Windows is worth fighting for (dozens of seconds max). Unless I am missing something...

bollwyvl:
Ha, _finally_ got a cache hit on windows for the conda packages. Now to see if it makes a lick of difference...

bollwyvl:
Haven't dug into the repeated `TabError` fail (first on osx, then on win), but otherwise starting to look good.

I've also warmed back up the `setup-miniconda` PR which would directly support the lock files:

https://github.com/conda-incubator/setup-miniconda/pull/74

krassowski:
Stale?

bollwyvl:
Definitely stale here, though the approach is sound. I'm working on some
related stuff... Hopefully will have more to show soon.

bollwyvl:
I'm not at my keyboard, but this can be safely closed!

PR #359, opened 2020-09-13
Faster updates on partial changes in notebooks

### LINKED ISSUE #336: Virtual document updates are slow for large notebooks
## Description

In notebooks with many editors (150+) each letter typed triggers an update of the entire virtual document, which can be very expensive. Instead, it should only patch the `VirtualBlock` that has changed to minimise the delay - we should only rebuild the document on editors were moved around OR if a change of a virtual documents number inside the edited block was detected; this should help speed up the completer too (see #272). In addition, we need profiling to see what are the other bottlenecks.

## Context

- Operating System and version: Ubuntu 20.04
- Browser and version: Chrome 85
- JupyterLab version: 2.2.6

<pre>
config dir: /home/krassowski/.pyenv/versions/3.8.1/envs/pelvic_pain/etc/jupyter
    jupyter_lsp  enabled 
    - Validating...
      jupyter_lsp 0.9.2 OK
    jupyterlab  enabled 
    - Validating...
      jupyterlab 2.2.6 OK
</pre>
</details>

<details><summary>Required: installed lab extensions</summary>
<pre>
        @krassowski/jupyterlab-lsp v2.0.0  enabled  OK
        @krassowski/jupyterlab_go_to_definition v1.0.0  enabled  OK
</pre>
</details>

krassowski:
## References

Aims to fix #336

### Perform partial VirtualDocument updates rather than a full update each time

- [x] Measure current performance using a medium-length notebook (~50 code cells)
- [ ] Refactor virtual document updates to use code blocks which can be selectively updated (rather than triggering an update of the full document) - done, not committed yet to do the above
- [ ] Measure the performance after the change

Result: the js-side updates are not the bottleneck, even if done in the most basic linear and non-partial way; however using the partial updates can reduce the amount of data transferred between to the backend (server extension) and then to the LSP server. It is not clear if this will be worth the effort.

## User-facing changes

Aims:
- medium, long and very long notebooks will be much faster (including the completer)
- short notebooks (<10 cells) will be imperceptibly slower

None

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


krassowski:
Closing, the virtual document development is going to continue in JupyterLab core.

Issue #297, opened 2020-07-25
Add Robot Tests for hover


bollwyvl:
## Elevator Pitch

We should test hover behaviors.

## Motivation

Hover is an important feature for understanding code that should be pretty responsive.

## Design Ideas

- In `01_Editor.robot`, each server that supports hover should have a `hover=Message here`. 
- A new keyword, `Editor Shows Hover  ${selector}   ${message}`. Maybe have a configurable timeout... some language servers probably have a cache.

krassowski:
Working on this now, will PR soon (though not for all servers and in notebooks only - feel free too pick up and add tests in `01_Editor.robot` afterwards!)

krassowski:
Kind of depends on https://github.com/robotframework/SeleniumLibrary/issues/1639.

Though I think we could have a setting to show the hover tooltip after `n` seconds anyways. `n > 5`

bollwyvl:
Action chains do work, but are non-obvious... I had to ctrl+shift drag
something once (nightmares). I'll see i can dig it up.

krassowski:
Actually I just realised that I implemented hover in the way that it is sufficient to press the modifier after hover to bring the tooltip up, which is intended so that the user does not have to move the mouse again :)

krassowski:
Extracted the idea of adding delay-based activation to #362.

krassowski:
As of today hover is tested in `Hover.robot`:

https://github.com/jupyter-lsp/jupyterlab-lsp/blob/7ae3c056ed5092b8081719138ead8d12902434e5/atest/05_Features/Hover.robot#L13-L45

This issue proposed to add a basic test of hover per-server. Given the simplicity of this feature I am not convinced it is necessary; I do think that testing some features for all supported servers is a good thing (e.g. diagnostics as a sanity check "is this server working" + rename because it is very complex) just not sure if the added burden of adding hover to the list is worth it -> I will not be working on it, but if anyone wants to, I am happy to review the pull request :)

PR #363, opened 2020-09-14
Fix highlighting in hover tooltips and start testing the hover feature

krassowski:
## References

A step towards #297

## Code changes

- Fixed missing new lines in markdown generation which were causing the code in the tooltips to be rendered without syntax highlighting
- Added setting for hover tooltip debouncer.
- Removed unneeded margin/padding on the bottom and right side of the tooltip
- Added `className` option to the custom tooltip subclass
- Added tests for hover in notebooks; they check that the syntax highlighting is correctly applied
- Fix the underline
- Improve multi-element hover display (deprecated API)

## User-facing changes

| before | after |
|--------|-------|
| before_lsp   |  after_hover   |

Final:
hover

None

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
Actually, we used to get these nice underlines indicating whether hover is available or not. This seems to be broken in 2.0.

krassowski:
Underlines are now fixed; the only remaining problem is with tests, as the improved implementation does not allow to work it around easily. Will need to implement the delay-based hover for tests...

krassowski:
@bollwyvl, how would you approach the action chains issue? There is https://github.com/hmalphettes/robotframework-selenium2library-extensions but it was not updated in ages nor released. We could try to extend SeleniumLibrary https://github.com/robotframework/SeleniumLibrary/blob/master/docs/extending/extending.rst (lot's of reading it seems, no time at the moment..)

bollwyvl:
Sorry, didn't have a chance to dig into this yet, nor have I been able to
find my old code.

I'd probably make a quick and dirty library that used raw selenium stuff,
e.g. hover_with_modifier(selector, modifier=None) and hide the action chain
stuff in there. We should not take that library you mention as a
dependency, it's caused me issues in the past.

Speaking if freaky dependencies: I have also been looking into:

https://github.com/MarketSquare/robotframework-browser

...which uses the nodejs-based playwright, which theoretically supports
more browsers, though with less control, with more rando binary blobs.

Anyhow, its "Hover" keyword already accepts modifiers:

*https://marketsquare.github.io/robotframework-browser/Browser.html#Hover
<https://marketsquare.github.io/robotframework-browser/Browser.html#Hover>*

I'll be trying it out on some greenfield stuff, and if it's useful, looking
into packaging it, yadda yadda.

PR #381, opened 2020-10-08
Upgrade linters, language servers

bollwyvl:
## References

Generally referenced in a few places, some of the other open PRs have already applied the python changes. Hopefully we can get all these diffs in one PR and move forward on other things.

## Code changes

- updates `eslint`, `typescript` and various plugins
- updates `yaml-language-server`, which hopefully works with schema now
- clean refresh of `yarn.lock`
- runs formatters, linters
- adds all `_*.d.ts` to `.gitignore`

## User-facing changes

N/A (aside from yaml working better)

N/A

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
+++ b/packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
@@ -100,7 +100,7 @@ export abstract class CodeMirrorIntegration
   // TODO use better type constraints for connection event names and for responses
   protected readonly connection_handlers: Map<
     string,
-    (response: object) => void
+    (response: Record<string, unknown>) => void
   >;
   protected readonly wrapper_handlers: IHTMLEventMap;
   protected wrapper: HTMLElement;
--- a/packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
+++ b/packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
@@ -100,7 +100,7 @@ export abstract class CodeMirrorIntegration
   // TODO use better type constraints for connection event names and for responses
   protected readonly connection_handlers: Map<
     string,
-    (response: Record<string, unknown>) => void
+    (response: Record<string, any>) => void
   >;
   protected readonly wrapper_handlers: IHTMLEventMap;
   protected wrapper: HTMLElement;
--- a/packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
+++ b/packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
@@ -15,7 +15,7 @@ import {
 } from '../positioning';
 import * as lsProtocol from 'vscode-languageserver-protocol';
 import { PositionConverter } from '../converter';
-import { DefaultMap } from '../utils';
+import { DefaultMap, uris_equal } from '../utils';
 import { CodeEditor } from '@jupyterlab/codeeditor';
 import {
   CodeMirrorHandler,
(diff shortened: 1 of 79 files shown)


krassowski:
It seems that the typescript upgrade gives as a difficult time on:

```
@krassowski/jupyterlab-lsp:     TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
@krassowski/jupyterlab-lsp:     src/features/diagnostics/diagnostics.ts:268:3 - error TS2610: 'settings' is defined as an accessor in class 'CodeMirrorIntegration', but is overridden here in 'DiagnosticsCM' as an instance property.
```

bollwyvl:
Yeah, ick, will check it out next chance i get.

bollwyvl:
Giving this another shot... passes locally...

bollwyvl:
ah, i didn't grok that it was failing during tests... will investigate...

bollwyvl:
Some progress here: found some [upstream language server issues](https://github.com/redhat-developer/yaml-language-server/issues/345), and have made some headway on fixing the property re-definitions... jest still failing, though...

bollwyvl:
Really sorry this fell off for a sec... got distracted by... things. 

Anyhow: helluva yak shave, but everything just ran locally (though pulled it a bit hot without linting)

- the TS2610 "fix" just overloads the `get` for `settings` and `lab_integration` with the feature types
  - we could make the features generic over their settings schema, and pass that down in a few places
- `jest`(`-ts`) was annoying, but works now
  - some paths were weird (e.g. `file://dummy.py` vs `file:///dummy.py`)... 
    - we should capture that as constant someplace, i guess, 
      - ... and probably not use `/` (it's `/unit-test` someplace, which is better)
- yaml-language-server speaks schema now, so those tests _can_ pass :tada: (or fail, and provide insight)
  - ... but vscode dropped a hot version of the json language service breaking backcompat :ghost:  
    - our pin, is... fine. for now. for testing purposes. but i actually _really_ want that to work
      - ... so i have a pr open upstream, which might drop before i get this monster done
- `bash-language-server` disabled their error highlighting by default... 
  - ... i turned it back on in the spec...
    - ... but then decided it's just better to patch the env during test, rather than force the opinion
- bumped all the notebook metadata to the hottest python 3.7

Here's hoping on CI finally going through... or at least getting to where it can tell me something broken on a non-linux platform.

bollwyvl:
so yeah: i kinda just don't care about jest on windows, i guess...

bollwyvl:
Aw

bollwyvl:
Huzzah, back to green.

packages/jupyterlab-lsp/src/editor_integration/codemirror.ts
```diff
@@ -279,10 +279,7 @@ export abstract class CodeMirrorIntegration
     for (let change of changes) {
       let uri = change.textDocument.uri;
 
-      if (
-        decodeURI(uri) !== decodeURI(current_uri) &&
-        decodeURI(uri) !== '/' + decodeURI(current_uri)
-      ) {
+      if (!uris_equal(uri, current_uri)) {
```
krassowski, line 282: I'm very apprehensive of such changes, but let's try it as tests pass...

krassowski:
Thank you!

bollwyvl:
Awesome!

Stay tuned for a small follow-on with an updated setup-miniconda action to
squash more warnings which will soon become fails.

>
Regarding the uri_equals: if you look at that code, it's doing the same
thing, but ensuring windows path normalization, which may well have been
causing issues for folks... I just missed it, when refactoring, presumably.

PR #390, opened 2020-10-25
Attempt to fix TabError failures on CI

krassowski:
It's just a guess, nothing more.


krassowski:
It seems to be working :)

Explanation: we were frequently getting results from the native JupyterLab completer because our was not ready yet. The binding of completer in notebook happens on cell activation, so the cell has to be activated after the initialization of LSP is fully complete.

Issue #398, opened 2020-11-05
Context menu does not get initialized when the extension is registered last


mnowacki-b:
## Description

There may be a potential race condition when initializing the context menu. It currently get only registered when an even is fired when a new type is added to the `adapterManager`.
https://github.com/krassowski/jupyterlab-lsp/blob/a65019a7a5138510c568198f8179c5af1e077a5f/packages/jupyterlab-lsp/src/index.ts#L171
If the notebook adapter is added before the extension (jupyterlab-lsp) get registered; that even will be fired and potentially processed before the connection is made and the context mane will not be registered for that type.

## Reproduce

1. Edit your JupyterLab to make sure the extension (jupyterlab-lsp) is added at the end.
2. Build and run

## Expected behavior

Context menu should be initialized regardless of the order of extension initialization.

## Context

- Operating System and version: rhel7
- Browser and version: Chrome 86.0.4240.111
- JupyterLab version: 2.2

krassowski:
Sorry it took a while; releasing today!

PR #399, opened 2020-11-05
Register Context Commands with already added types

mnowacki-b:
## References

Issue: https://github.com/krassowski/jupyterlab-lsp/issues/398

## Code changes

Registering Context Commands Manager with already added types.

## User-facing changes

None

Backword compatible.

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/index.ts
+++ b/packages/jupyterlab-lsp/src/index.ts
@@ -5,6 +5,7 @@ import {
 import { ICommandPalette } from '@jupyterlab/apputils';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { IDocumentManager } from '@jupyterlab/docmanager';
+import { IDocumentWidget } from '@jupyterlab/docregistry';
 import { Signal } from '@lumino/signaling';
 import { LanguageServerManager } from './manager';
 import '../style/index.css';
@@ -13,6 +14,7 @@ import { IStatusBar } from '@jupyterlab/statusbar';
 import { LSPStatus } from './components/statusbar';
 import { DocumentConnectionManager } from './connection_manager';
 import {
+  IAdapterTypeOptions,
   ILSPAdapterManager,
   ILSPCodeExtractorsManager,
   ILSPFeatureManager,
@@ -26,7 +28,7 @@ import { SIGNATURE_PLUGIN } from './features/signature';
 import { HOVER_PLUGIN } from './features/hover';
 import { RENAME_PLUGIN } from './features/rename';
 import { HIGHLIGHTS_PLUGIN } from './features/highlights';
-import { WIDGET_ADAPTER_MANAGER } from './adapter_manager';
+import { WidgetAdapterManager, WIDGET_ADAPTER_MANAGER } from './adapter_manager';
 import { FILE_EDITOR_ADAPTER } from './adapters/file_editor';
 import { NOTEBOOK_ADAPTER } from './adapters/notebook';
 import { VIRTUAL_EDITOR_MANAGER } from './virtual/editor';
@@ -166,9 +168,7 @@ export class LSPExtension implements ILSPExtension {
         console.error(reason.message);
       });
 
-    adapterManager.registerExtension(this);
-
-    adapterManager.adapterTypeAdded.connect((manager, type) => {
+    const registerContextCommandManager = (type: IAdapterTypeOptions<IDocumentWidget>): void => {
       let command_manger = new ContextCommandManager({
         adapter_manager: adapterManager,
         app: app,
@@ -179,6 +179,18 @@ export class LSPExtension implements ILSPExtension {
         ...type.context_menu
       });
       this.feature_manager.registerCommandManager(command_manger);
+    };
+
+    // Register context commands with already added types
+    adapterManager.types.forEach((type: IAdapterTypeOptions<IDocumentWidget>) => {
+      registerContextCommandManager(type);
+    });
+
+    adapterManager.registerExtension(this);
+
+    // Register context commands with any types that may be added later
+    adapterManager.adapterTypeAdded.connect((manager: WidgetAdapterManager, type: IAdapterTypeOptions<IDocumentWidget>) => {
+      registerContextCommandManager(type);
     });
   }
 
--- a/packages/jupyterlab-lsp/src/index.ts
+++ b/packages/jupyterlab-lsp/src/index.ts
@@ -28,7 +28,7 @@ import { SIGNATURE_PLUGIN } from './features/signature';
 import { HOVER_PLUGIN } from './features/hover';
 import { RENAME_PLUGIN } from './features/rename';
 import { HIGHLIGHTS_PLUGIN } from './features/highlights';
-import { WidgetAdapterManager, WIDGET_ADAPTER_MANAGER } from './adapter_manager';
+import { WIDGET_ADAPTER_MANAGER } from './adapter_manager';
 import { FILE_EDITOR_ADAPTER } from './adapters/file_editor';
 import { NOTEBOOK_ADAPTER } from './adapters/notebook';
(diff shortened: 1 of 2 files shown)


bollwyvl:
Thanks for wading into this! That constructor is getting to be a bruiser (which is not your fault, at all). I wonder if we should start wrapping some of these long constructors into `IOptions` kinds of patterns. There's just a lot going on! I'll comment on the code directly...

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -166,9 +168,7 @@ export class LSPExtension implements ILSPExtension {
         console.error(reason.message);
       });
 
-    adapterManager.registerExtension(this);
```
bollwyvl, line 169: if the ordering of this line and the `connect` below doesn't matter, perhaps we could:
- `connect` first
- `resigterExtension`
  - have `registerExtension` (which I assume can only be called once?) do `this.types.foreach((t) => this.adapterTypeAdded.emit(t))`

krassowski: I like this idea! In addition, it is worth noting that `registerExtension` already does two loops (one to connect to already registered adapters and second to watch for future registrations), so it seems like a natural place to address this problem.

mnowacki-b: Thank you both for your quick response!
It sounds like a good idea; but I do have concerns (maybe they are unfounded).
If we do reemit `adapterTypeAdded` signal won't we cause potential problems if another extension decides to connect to it and then starts to receive multiple events with the same type?
Also, can we really assume that `resigterExtension` can be called only once? If so should we have something in place to enforce that?
Thank you!

mnowacki-b: @bollwyvl @krassowski 
I am more then happy to make the requested changes; but I thought I will make a small proposal first :)
I made some alterations please have a look and let me know what you think.
Thank you!

bollwyvl:
Checking out locally... may I recommend a merge/rebase of `master`, as we bumped a lot of deps and applied a lot of formatting on #381?

packages/jupyterlab-lsp/src/index.ts
```diff
@@ -167,19 +169,22 @@ export class LSPExtension implements ILSPExtension {
...
-        adapter_manager: adapterManager,
-        app: app,
-        palette: palette,
-        tracker: type.tracker,
-        suffix: type.name,
-        entry_point: type.entrypoint,
-        ...type.context_menu
-      });
-      this.feature_manager.registerCommandManager(command_manger);
+  registerAdapterType(
```
bollwyvl, line 174: :+1: to extraction to a method.

packages/jupyterlab-lsp/src/index.ts
bollwyvl: Gah, lost what i was typing. So it goes.

So: it looks like `registerExtension` is expecting an `LSPExtension`, rather than an `ILSPExtension`. I think this kinda signals that it's meant for in-house usage, or would be entirely replaced, rather than having multiple. Perhaps @krassowski can suggest some tactical docstrings. If it can be multiple, then `registerAdapterType` probably needs to be added to the interface...

bollwyvl:
Hooray green. So... how might we test this? 

We don't expect folks to wade into non-trivial robotframework stuff, but I guess we can have a simple python function that rewrites `index.out.js`. I don't know off-hand what other extensions this might conflict with... have any in mind?

packages/jupyterlab-lsp/src/index.ts
krassowski: > it looks like `registerExtension` is expecting an `LSPExtension`, rather than an `ILSPExtension`. I think this kinda signals that it's meant for in-house usage

yes, this is an accurate description.

krassowski (approved):
Thank you @mnowacki-b, this looks great to me!

@bollwyvl is this ok with you too?

krassowski:
@bollwyvl re-testing - IMO this is very much on the integration-level testing level so I am happy to accept as-is on the JS side.

@mnowacki-b did you encounter this when looking through the code or was it an interface glitch that led you to discover the race? If the latter then we might be able to create a GUI (robotframework) test for it (though testing race conditions can be tricky).

mnowacki-b:
Thank you both for the quick turnaround!

@krassowski It was an interface glitch. We have a unique way of loading new extensions in our flavor of JupyterLab. We enable them during runtime based on configuration (during JupyterLab initialization). And so those extensions activation is slightly delayed.

bollwyvl:
> our flavor of JupyterLab

We're _very_ interested in co-developing a stable baseline for an ecosystem around language server features in Lab, but it's pretty tough, as it seems folks are just taking the code and running, and making stuff that isn't compatible with this work... which is totally cool, and why it's open source!

> enable them during runtime

Can you show a tiny example that demonstrates this behavior? If your hacks are occurring too far down the stack (e.g `@lumino/application` though... we might not be able to support it.

mnowacki-b:
By 'our flavor' I meant a custom list of extensions. We do not modify the core JLab we only specify a list of custom extensions that we are interested in. Most of the extensions we use are open sourced with a handful of internal extensions.

I still believe this is an issue that may occur as the plugins are loaded asynchronously and the list of adapter type can be updated before `registerExtension` occurs.

bollwyvl:
> the list of adapter type can be updated before registerExtension occurs.

Perhaps another signal/PromiseDelegate is needed? A thing we resort sometimes for these topics is ASCII art sequence diagrams, drawio pictures, etc.

More broadly: I hope I'm not coming off as belligerent: we're at a disadvantage here, as we haven't seen this (mis)behavior in action. If we can work together to reproduce and test it, in a real browser, we can make sure it keeps working. It may mean this is a "custom" lab extension/excursion in CI, but it's work worth doing.

Our semi-ridiculous integration suite has been the only we have been able to confidently add new features and extension points, and handle migrations between different JupyterLabs... and things we _don't_ test have definitely regressed.

This kind of testing will be _much_ better with the forthcoming jupyterlab 3 work being discussed on #400, where flipping the installed extensions will be measured in milliseconds, not minutes, but I can totally understand if you've got commitments to lab2 (i still ship lab 1 on some projects, because some in-house extensions never updated).

mnowacki-b:
Totally understand, I can try to see if I can come up with something that can reproduce the issue.
We are actually looking forward to migrating to 3.0 as well; however, we have to wait a little bit before we can do that.
We are also very excited about adding juputerlab-lsp and we just don't want to wait :).

mnowacki-b:
I put together this rough diagram to demonstrate how this race condition can occur.
As you can see in this scenario all it takes is for INotebookTracker to initialize before all of the ILSPFeatureManager dependencies are initialized. This will result in Notebook's context menu without LSP menu items.

lsp-race

mnowacki-b:
So another aproceh to fix this would be require `NOTEBOOK_ADAPTER` and `FILE_EDITOR_ADAPTER` to wait for `ILSPFeatureManager`

mnowacki-b:
@krassowski @bollwyvl 
Any thought on https://github.com/krassowski/jupyterlab-lsp/pull/399#issuecomment-723480685 ?

mnowacki-b:
Is there anything I can do to expedite this PR?

krassowski:
I think that we can merge it as is. I think the current approach is a step forward and even if we do not test against the faulty behaviour, the changes improve the quality of code. Thank you for all the help.

mnowacki-b:
Thank you!
I am still fairly new to open source contributions; it there is anything I can help with please let me know.

If I may ask; what it your deployment process? how often do you cut releases?

krassowski:
I will probably do one once #406 is in. Sorry extremely busy week for me.

mnowacki-b:
@krassowski Thank you!

PR #401, opened 2020-11-06
remove travis, update setup-miniconda

bollwyvl:
## References

- follow-on to #381

## Code changes

- no functionality changes
- upgrades to pre-release of setup-miniconda@v2
  - removes extra channels in workflows (throws warnings)
- removes travis-ci
  - https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works
  - we lose
    - py 3.5, was EOL
    - pip-only testing (could restore on a separate workflow with setup-python)

## User-facing changes

- removes travis badges from readme

N/A

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


bollwyvl:
I'd like to add py39, but it looks like there are a few remaining gotchas in the solve. I will investigate further, but we'll want to get this through to squash the warnings...

bollwyvl:
Still can't figure out the py39 thing, or how to get conda to stop complaining, but have squashed a fair number of warnings... this is certainly a step forward!

krassowski (approved):

bollwyvl:
Thanks for the check. I really wish we could avoid the 

    Warning: 'defaults' already in 'channels' list, moving to the top

Junk, otherwise we'd be coming up clean, and the warnings list would be empty on a good run, rather than having to muck about...

bollwyvl:
I think i may have gotten it down to a single error per job, which is... better...

bollwyvl:
Nope, not really. force pushed back to the approved commit... guess it's going to fire again, but have no other stuff planned.

bollwyvl:
Hm, `master` looks on fire. Haven't dug into it yet, hopefully will be able to resolve on #406... might not be tonight...

Issue #271, opened 2020-05-14
Windows development challenges


bollwyvl:
At present, I think all of our contributors are working on Linux/MacOS, and while we test fairly exhaustively on Windows in CI, it's not much fun to develop locally.

Some things I've encountered trying to test #245:
- `binder/postBuild` is written in bash
  - we could write it in python
  - as more things seem error prone (anything involving node), `doit` (a la #183) would probably be helpful
- `prettier` doesn't work
  - we can change the globs to be `cmd`-friendly

will update as i find more...

bollwyvl:
Opening the python example:
```
[W 20:04:32.280 LabApp] [lsp] error in listener <function setup_shadow_filesystem.<locals>.shadow_virtual_documents at 0x0000024320AEC318> for message {'jsonrpc': '2.0', 'method': 'textDocument/didChange', 'params': {'textDocument': {'uri': 'file:///c:/git/lsp/jupyterlab-lsp/.virtual_documents/examples/Python.ipynb.py', 'version': 1}, 'contentChanges': [{'text': 'def square(x):\n    """Can you see me?"""\n    return x*x\n\n\nresult = square(2)\n\n\nfrom statistics import mean\n\n\nundefined_variable\n\n\nclass Dog:\n    \n    def bark(self):\n        print(\'🐕 woof woof\')\n\n\n\n\n\nDog().bark()\n\n\nclass Cat:\n\n    def miaow(self):\n        print(\'miaow\')\n\n\nCa\n\n\nCat.__\n\n\nCat\n\n\ntest = 1\n\n\ntest\n\n\ndef abc():\n    test = 2\n    test\n'}]}}
    Traceback (most recent call last):
      File "c:\git\lsp\jupyterlab-lsp\py_src\jupyter_lsp\types.py", line 96, in __call__
        manager=manager,
      File "c:\git\lsp\jupyterlab-lsp\py_src\jupyter_lsp\virtual_documents_shadow.py", line 172, in shadow_virtual_documents
        await editable_file.write()
      File "c:\git\lsp\jupyterlab-lsp\py_src\jupyter_lsp\virtual_documents_shadow.py", line 36, in write
        return await convert_yielded(self.write_lines())
      File "C:\git\lsp\envs\default\lib\concurrent\futures\thread.py", line 57, in run
        result = self.fn(*self.args, **self.kwargs)
[...]

PR #405, opened 2020-11-09
JupyterLab 3 update

martinRenou:
This is the quick path to get this working with JupyterLab 3, not finished yet. I still need to make jupyterlab-lsp a federated labextension.


.prettierignore
```diff
@@ -2,6 +2,7 @@
 .mypy_cache
 .pytest_cache
 .yarn-packages
+**/_*.d.ts
```
bollwyvl, line 5: if these end up in the dist (which they might) we will want them formatted...

packages/jupyterlab-go-to-definition/package.json
```diff
@@ -46,34 +46,37 @@
...
+    "@lumino/algorithm": "^1.3.3",
     "@types/chai": "^4.1.7",
+    "@types/codemirror": "^0.0.74",
     "@types/jest": "^23.3.11",
     "chai": "^4.2.0",
+    "codemirror": "*",
     "jest": "^26.0.0",
     "jest-junit": "^8.0.0",
-    "rimraf": "^2.6.1",
+    "npm-run-all": "^4.1.5",
```
bollwyvl, line 71: why do we need this? I've had issues on windows with this in the past. lerna seems much better at doing things with more controlled parallelism...

requirements/github-actions.yml
```diff
@@ -1,10 +1,13 @@
 name: jupyterlab-lsp
 
 channels:
+  - conda-forge/label/jupyterlab_rc
+  - conda-forge/label/jupyterlab_server_rc
   - conda-forge
   - defaults
 
 dependencies:
+  - jupyter-packaging
```
bollwyvl, line 10: would prefer to avoid, and stay on a bog-standard `setup.py`/`setup.cfg` that doesn't invoke `node` during `pip install` like _ever_. i've had trouble with jupyter_packaging, in the past, and during this transition.

requirements/lint.yml
```diff
@@ -13,3 +13,5 @@ dependencies:
   - mypy
   - robotframework-lint >=1.1
   - robotframework >=3.2
+  - pip:
+    - pytest_tornasync
```
bollwyvl, line 17: ?

setup.py
```diff
@@ -1,20 +1,59 @@
...
+    BaseCommand,
+)
+
+def install_labextension(path):
+
+    class LabExtensionCommand(BaseCommand):
+
+        def run(self):
+            log.info('Installing labextension with jlpm...')
+            run(['jlpm', 'bootstrap'], cwd=path)
```
bollwyvl, line 25: :angry:

packages/jupyterlab-go-to-definition/package.json
martinRenou: This might not be needed, at first glance, it looks like I can remove it. This was added by the Jupyterlab update script I suppose.

requirements/github-actions.yml
martinRenou: If we want the users to only have to run `pip install jupyterlab-lsp` to get everything working, without needing them to install `node`, that's the only way we can do it AFAIK.

Note that `node` would only be needed on the machine that builds the package, which is not the user's machine (unless they do a dev install of course).

requirements/lint.yml
martinRenou: That seems to be needed due to a jupyter-server issue, similar to here: https://github.com/voila-dashboards/voila/issues/728

setup.py
martinRenou: I understand what you say, it makes life a bit more difficult for the devs.

But this workflow allows a smoother user workflow, as they don't need `node` anymore on their machine to get jupyterlab-lsp working. They will only need to run `pip install jupyterlab-lsp` and it will ship the labextensions for them.

martinRenou: @bollwyvl what would you think of having a bash script around that does the `jlpm bootstrap` prior to `python setup.py sdist/bdist_wheel`? So that the setup.py don't actually depend on npm anymore. But the devs should be careful to actually build the labextension before pushing to PyPi.

.github/workflows/job.test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       matrix:
         python: [3.6, 3.7, 3.8]
         os: [ubuntu-16.04, macos-10.14, vs2017-win2016]
-        lab: ['>=2.2.0,<3.0.0a0']
+        lab: ['>=3.0.0-rc.7,<4.0.0a0']
```
jtpio, line 40: Should this be?

```suggestion
        lab: ['>=3.0.0rc.7,<4.0.0a0']
```

requirements/github-actions.yml
bollwyvl: Hm... pretty sure we just need to deliver stuff to where it can get picked up with `data_files`... this is a WIP thing I've been tinkering with that just uses `setuptools`:

```py
"""
jupyterlab_graphviz setup
"""
import re
from pathlib import Path

import setuptools

NAME = "jupyterlab_graphviz"
HERE = Path(__file__).parent.resolve()
SRC = HERE / "src_py" / NAME
STATIC = SRC / "static"
STATIC_FILES = [str(p.relative_to(HERE)) for p in STATIC.rglob("*.*")]
VERSION = SRC / "_version.py"
VERSION_RE = r"""__version__ = "([^"]+)"""
REMOTE_ENTRIES = [p for p in STATIC_FILES if "remoteEntry" in p]

INDENT = "\n    "

assert (
    len(REMOTE_ENTRIES) == 1
), f"""
Exactly one remoteEntry.*.js expected in {STATIC}, found...

    {INDENT.join(REMOTE_ENTRIES)}

Try:

    jlpm clean
    jlpm build

"""


SETUP_ARGS = dict(
    name=NAME,
    version=re.findall(VERSION_RE, VERSION.read_text())[0],
    data_files=[(f"share/jupyter/labextensions/{NAME}", STATIC_FILES)],
)


# the rest of it is poulated in setup.cfg
if __name__ == "__main__":
    setuptools.setup(**SETUP_ARGS)
```

requirements/lint.yml
bollwyvl: Welp.... https://github.com/jupyter/jupyter_server/issues/337

setup.py
bollwyvl: Aw, man, you don't have to sell me on lab3! We all want it. But, much like webpack, things go into a crazy, changes-per-month hole inside of a whl/sdist build, and I try to keep them as bog-standard as possible. My kingdom for being able to do binary-reproducible builds with `flit` and just a `pyproject.toml`, but we're stuck with `data_files` right now.

We have enough users/issues with windows (#271) that I don't think we can accept any bash-only stuff right now. While it pains me, sometimes we need as rapid development cycle, on as real a windows host, as we can get.

What I would _very much_ like to see, but haven't had the time to re-engineer, and _certainly_ don't expect you to do on this PR, is a `doit` (why not `make`? because #271...) workflow that _knew_ it couldn't build an sdist/whl so long as `remoteEntry` didn't exist, and knew it couldn't build that until it had been `tsc`'d and... Better still, if, during the build, a file that got delivered there contained the _list_ of files that were there, so we knew _exactly_ which files we were shipping, I'd be ecstatic. My super defensive thing above is to avoid the output of watch. 

Also, once it hits conda-forge, it needs to not explode, nor require `node` (or anything) in `build`, much less `host`.

So yeah: in the meantime, I'd say we want a build that does a dumb `data_files`, and _preferably_ explodes if the numbers are off in ways we can predict.

.github/workflows/job.test.yml
bollwyvl: But now there's 8!

requirements/github-actions.yml
martinRenou: Let's update my PR and do that yes.

It would be nice if we had a way to at least check that the static directory is up to date.

setup.py
martinRenou: >  I'd say we want a build that does a dumb `data_files`

Yes! I'll go for that then.

.github/workflows/job.test.yml
jtpio: Yes it is!

binder/postBuild
```diff
@@ -1,23 +1,12 @@
...
-# link all pieces to lab (no build)
-jlpm lab:link && jupyter labextension list
-
-# list extensions before build
-jupyter labextension list
-
-# actually build
-jupyter lab build --debug --dev-build=False --minimize=True
+# Do a dev install of the server side
+pip install . -vv
```
bollwyvl, line 8: Please bring back `python -m pip install -e . --ignore-installed --no-deps -vv`... it's not _too_ livehackable, today (vs #278), but would like to maintain that experience.

setup.py
```diff
@@ -1,20 +1,37 @@
...
+
+def get_data_files():
+    chdir(str(LABEXTENSIONS_DIR))
+
+    extension_files = [
+        (str(LABEXTENSIONS_INSTALL_DIR / Path(filename).parent), [str(LABEXTENSIONS_DIR / filename)])
+        for filename in iglob('**/*.*', recursive=True)
+    ]
+
+    chdir('../../../')
```
bollwyvl, line 23: ?

setup.py
```diff
@@ -1,20 +1,37 @@
...
+LABEXTENSIONS_INSTALL_DIR = Path('share') / 'jupyter' / 'labextensions'
+
+
+
+def get_data_files():
+    chdir(str(LABEXTENSIONS_DIR))
+
+    extension_files = [
+        (str(LABEXTENSIONS_INSTALL_DIR / Path(filename).parent), [str(LABEXTENSIONS_DIR / filename)])
+        for filename in iglob('**/*.*', recursive=True)
```
bollwyvl, line 20: Can just be `LABEXTENSIONS_DIR.rglob('*')`? This release won't work with py35 anyway, because of all the upstreams.

binder/postBuild
martinRenou: Sure, I'll bring it back

setup.py
bollwyvl: Also `relative_to` could be helpful

setup.py
```diff
@@ -1,20 +1,37 @@
...
+    ]
+
+    chdir('../../../')
+
+    extension_files.append(("etc/jupyter/jupyter_notebook_config.d", ["py_src/jupyter_lsp/etc/jupyter-lsp-serverextension.json"]))
```
bollwyvl, line 25: as has been mentioned elsewhere: maybe we have the structure in-package already looks like the system-of-interest, e.g. `etc/` and `share/` so the patterns are more concise.

packages/jupyterlab-lsp/package.json
```diff
@@ -28,46 +28,61 @@
     "url": "https://github.com/krassowski/jupyterlab-lsp.git"
   },
   "scripts": {
-    "build": "jlpm build:schema && tsc -b",
+    "build": "jlpm run build:lib && jlpm run build:labextension",
```
bollwyvl, line 31: not sure what all this stuff does, but we try to keep the individual packages.json clean, and only handle special cases, then hoist stuff to either `meta` or the root package.

setup.py
martinRenou: That's what I was looking for! Thanks

packages/jupyterlab-lsp/package.json
```diff
@@ -79,28 +94,30 @@
     "jest": "^26.0.0",
     "jest-junit": "^8.0.0",
     "json-schema-to-typescript": "^8.0.0",
-    "react": "*",
-    "rimraf": "~2.6.2",
+    "mkdirp": "^1.0.3",
+    "prettier": "^1.19.0",
```
bollwyvl, line 98: think we went to 2 already?

packages/jupyterlab-lsp/package.json
```diff
@@ -118,6 +135,7 @@
         ]
       }
     },
-    "schemaDir": "schema"
+    "schemaDir": "schema",
+    "outputDir": "../../py_src/jupyter_lsp/labextensions/@krassowski/jupyterlab-lsp"
```
bollwyvl, line 139: we need to add checking all of these (and probably other things) to `integrity.py`

requirements/lint.yml
```diff
@@ -13,3 +13,5 @@ dependencies:
   - mypy
   - robotframework-lint >=1.1
   - robotframework >=3.2
+  - pip:
+      - pytest_tornasync
```
bollwyvl, line 17: made https://github.com/jupyter/jupyter_server/issues/337 (and 338, and 339..) but in the near term, we can source this from `conda-forge` as `pytest-tornasync`:

https://github.com/conda-forge/pytest-tornasync-feedstock

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-name = jupyter-lsp
+name = jupyter_lsp
```
bollwyvl, line 2: wut

martinRenou: This makes it importable :) enabling `jupyter labextension develop`

bollwyvl:
Pretty sure we discussed this, but think we might want to have a dependency tree like:
```
jupyterlab_lsp ---+---> jupyter_lsp ----+---> jupyter_server
                  \---> jupyterlab ----/
```
Motivation: we have been trying really hard to make the contract between the frontend and backend packages "regular" (e.g. schema-constrained). If the frontend starts getting shipped with the backend, this would become less formal, and downstreams that wish to extend the functionality would be at a disadvantage, and more likely to fork. 

Further, if we split out and version individual features in the future, `jupyterlab_lsp` can become a dependency-only metapackage, we can add 100 packages, and `jupyter_lsp` (but more importanty, docs) wouldn't change at all.

We'll do a major version bump for `jupyter_lsp`, and a can copy that version for `jupyterlab_lsp`, and they'll have pretty hard pins between the two of them. If anything, I've considered having a _third_ package which only held "our" JSON schema version of the LSP and the definition of the extra metadata stuff we need... but that's for another day.

setup.cfg
bollwyvl: distributed package name != import name... this is explicitly different _because_ of that. like i mentioned elsewhere, if something in the stack is conflating the two, it needs to be nuked from orbit. If it _has_ to be like this, then that amplifies the need for it to be a standalone package.

packages/jupyterlab-lsp/package.json
martinRenou: Your comment here got me see that I removed `jlpm build:schema`.

This script should be `jlpm run build:lib && jlpm run build:lib && jlpm run build:labextension`, the only difference with before is that it will build the labextension and put it in the `outpudDir` path.

setup.cfg
martinRenou: I am totally fine reverting this change, just note that you won't be able to use `jupyter labextension develop`.

> then that amplifies the need for it to be a standalone package

Yes, although I wanted to keep this PR simple for now and not make another package.

binder/postBuild
martinRenou: Although `-e` will not install the `data_files`, but we need them installed

martinRenou:
> Pretty sure we discussed this, but think we might want to have a dependency tree like

Not sure I understand the tree structure you are drawing here.

But I guess you are mentioning to make a separate Python package for the labextension? It might be indeed nicer. Although I didn't want this PR to drastically change the code structure.

The PR works now. You should get it to work with:
```
jlpm bootstrap
pip install .
jupyter serverextension enable --sys-prefix --py jupyter_lsp
```

Please don't hesitate to push to my branch if you want to change/polish some things.

bollwyvl:
Yeah, data_files (however it's wrapped) is gross, for _precisely_ this
reason. Would running "labextension develop" do the thing?

I'll be digging into what is happening with labextension develop, but for
reference, I use postBuild as part of my local workflow, to ensure it keeps
working without building on binder... So I'm very motivated to have binder
end up with a ready-to-debug environment for someone's pr... We should
probably have a separate branch for binder that's made of our up-to-version
info without any development junk (but still node, gah), but that, too, is
for another day...

martinRenou:
> Would running "labextension develop" do the thing?

Yes, it should. `labextension develop` will make a symlink from the local build to the `share/jupyter/labextension` directory. But we would need to figure out the package name issue (jupyter-lsp vs jupyter_lsp). This is something that should be fixed in JupyterLab itself maybe, it tries to import the package to find the `_jupyter_labextension_paths` function. Maybe JupyterLab could use `importlib.importmodule('jupyter-lsp')` so that it works.

bollwyvl:
Welp, we need to make a new branch for you to target, in the first place...
We have users on lab 2 will need to support for the foreseeable future.

But have no fear of making drastic changes if they make sense for (roughly
in order):
- end users
- volunteer maintainer
- first time contributors
- downstream developers
- jupyterhub administrators
- downstream package maintainers

Regarding the volunteer bit: it'll probably be tomorrow before I can get
back on this!

martinRenou:
Thanks for the suggestions. Yes, don't hesitate to open a PR to my branch :)

bollwyvl:
This seems close to coming home... @krassowski et al: how do we want to handle landing this? Two options:
- create a `3.x`, and land this PR against it
- create a `2.x`, and land this against `master`

As there is still no firm release date for lab 3, i'm favoring the former, but it's kinda all the same :shrug: 

Also as this is going to be turbulent, we might want to investigate looking into the more complicated RTD setup which gives us versioned releases.

martinRenou:
As pointed out by @karlaspuldaro only the readthedocs build is triggered on the CI. Any idea why Github actions is not triggered here? Maybe it's because of the draft status, I will change this

martinRenou:
I will need to rebase and resolve conflicts

krassowski:
Yes, I think it was the draft status, but if CI does not trigger with the next push let me know and I will investigate further. I will look into branching/catching up with releases later on Wednesday.

packages/jupyterlab-lsp/package.json
krassowski: changed on follow up

martinRenou:
🎉 Thanks a lot!

setup.cfg
jasongrout: See also https://github.com/jupyterlab/jupyterlab/issues/9576

Issue #411, opened 2020-11-16
Update dev setup documentation


karlaspuldaro:
Following the dev steps in [CONTRIBUTING.md](https://github.com/krassowski/jupyterlab-lsp/blob/master/CONTRIBUTING.md) I came across a few issues:
1. Wrong command to create a new conda environment:
```
$ conda env update -n jupyterlab-lsp
SpecNotFound: Invalid name, try the format: user/package
```
Conda version: 4.9.2
Solution: `conda create -n jupyterlab-lsp`

2. Node version expected to be 12+
```
$ jlpm
yarn install v1.21.1
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error @typescript-eslint/eslint-plugin@3.10.1: The engine "node" is incompatible with this module. Expected version "^10.12.0 || >=12.0.0". Got "10.0.0"
error Found incompatible module.
```

3. Recommend Python 3.6+
As per [jupyterlab requirements](https://github.com/jupyterlab/jupyterlab/blob/master/setup.py)

krassowski:
Fantastic work! Thank you. Please note that node 12 is not required. The message you see says:

```
Expected version "^10.12.0 || >=12.0.0".
```

meaning that the old versions of node 10 are not supported. We do run CI tests on node 10 (10.15.3 at current), see https://github.com/krassowski/jupyterlab-lsp/runs/1414536659?check_suite_focus=true#step:6:156

However, now that you have changed it on #412 I am willing to consider accepting the change if @bollwyvl agrees. On one hand, it would make sense to recommend the newer version, on the other the 10.x will not be EOL before 2021-04-30.

bollwyvl:
For the current line that supports jlab2, I feel we need to keep testing on
node10, as a number of os package managers don't have anything newer, and
errors at that part of the stack are not fun at all. Better we catch them
in CI!

For jlab3, devs will need node12... Users will not need node at all...
except for the node-based language servers. So we'll probably _still_ need
to keep testing on 10 until that above-mentioned EOL. But with jlab3, it
will be safe to do all of our up-front build stuff on one Linux machine, as
fast as possible, archive the wheels, and use those on the other platforms,
so the language servers are the only thing that need to be installed with
jlpm.... And we should be able to cache the yarn-packages as well.

So anyhow: getting these right, for all involved audiences is hard. having
the version check thing in place, irrespective of what the actual number is
today, is of value in its own right, as well help us avoid problems in the
[...]

PR #412, opened 2020-11-16
Documentation update

karlaspuldaro:
Update requirements in documentation as described in #411


bollwyvl:
This is great. We need to add this to the integrity checks!

lresende:
@karlaspuldaro there are similar references to prerequisites on the [readme](https://github.com/krassowski/jupyterlab-lsp/blob/master/README.md) that should be updated as well. Below is one example, but there are others.
```
You will need to have all of the following installed:

JupyterLab >=2.2.0,<3.0.0a0
Python 3.5+
nodejs 10+
```

karlaspuldaro:
@bollwyvl I ran `python scripts/integrity.py` tests and they're all passing

bollwyvl:
> I ran python scripts/integrity.py tests and they're all passing

Right: they would, because the things you fixed :heart: are _not_ under test. I'm proposing we _add_ these checks to integrity, based on the Source Of Truth for each of the items (most of which are already available in that script). Doesn't have to be on this issue, but will save frustrations in the future, as it's going to be turbulent coming up (i think lab3 needs node12, etc).

CONTRIBUTING.md
```diff
@@ -29,15 +29,15 @@ You can contribute to the project through:
...
+- `nodejs` 10.12+
+- `python` 3.6+
 - `jupyterlab` 2
 
 It is recommended to use a virtual environment (e.g. `virtualenv` or `conda env`)
 for development.
 
 ```bash
-conda env update -n jupyterlab-lsp   # create a conda env
+conda create -n jupyterlab-lsp       # create a conda env
```
bollwyvl, line 40: at some point, in an effort to decrease the number of files in the root folder, we moved things into `binder`, so now the `--file` arg is required... `conda env update` has the advantage of being able to, well, update an environment, instead of just creating it.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = jupyter-lsp
-description = Multi-Language Server WebSocket proxy for Jupyter notebook or lab server for Python 3.5+.
+description = Multi-Language Server WebSocket proxy for Jupyter notebook or lab server for Python 3.6+.
```
bollwyvl, line 3: also dropped this in favor of `python_requires` below. Of course, it's not strictly accurate, as we're still doing some unneeded song-and-dance for python 3.5, but since we don't test there anymore...

CONTRIBUTING.md
karlaspuldaro: oh that makes sense now, when running the command I didn't realize it was expecting a file, or which one. Thx for clarifying :)

krassowski:
Wonderful! Thank you both.

bollwyvl:
Thanks for making time for review work!

PR #413, opened 2020-11-20
Improve error message when there is no language server installed for the current language

martinRenou:
When there is no language server installed, it fails with
```
TypeError: Path must be a string. Received null
```

Which is not very easy to understand without looking into the code. This PR makes it fail with a nicer error message that tells what is wrong.

### DIFF
--- a/packages/jupyterlab-lsp/src/connection_manager.ts
+++ b/packages/jupyterlab-lsp/src/connection_manager.ts
@@ -326,6 +326,10 @@ export namespace DocumentConnectionManager {
       language
     });
 
+    if (language_server_id === null) {
+      throw `No language server installed for language ${language}`;
+    }
+
     return {
       base: baseUri,
       document: URLExt.join(baseUri, virtual_document.uri),


krassowski:
Would this show up in the status bar or in the developer console? This is probably a thing that we would want to have a test for. I am short on time a few days more but otherwise happy to help with writing the robot test for it.

martinRenou:
> Would this show up in the status bar or in the developer console?

In the developer console

bollwyvl:
Our console output is noisy to the point of being useless... if this is a big problem, should we see about integrating with the in-UI log output?

martinRenou:
> should we see about integrating with the in-UI log output?

That would be a big improvement for people who don't know how to open the dev tools (most JupyterLab users I guess), so they get an idea why it does not work.

krassowski:
I just got bitten by this error when opening `.gitignore` file, for me it was leading to:

```javascript
TypeError: Cannot read property 'replace' of null
    at url.js:61
    at Array.map (<anonymous>)
    at Object.join (url.js:61)
    at Function.solve_uris (connection_manager.ts:335)
    at DocumentConnectionManager.connect_socket (connection_manager.ts:131)
    at DocumentConnectionManager.connect (connection_manager.ts:282)
    at FileEditorAdapter.connect (adapter.ts:490)
    at FileEditorAdapter.connect_document (adapter.ts:324)
    at new FileEditorAdapter (file_editor.ts:55)
    at WidgetAdapterManager.connectWidget (adapter_manager.ts:66)
```

Anyways, merging as is and I will iterate to add some indication of this in the UI.

krassowski:
Thank you @martinRenou!

PR #416, opened 2020-12-06
virtual_documents customization

### LINKED ISSUE #353: virtual_documents customization
first of, thanks a lot for this amazing extension :1st_place_medal: 

## Description

> I do not think it is a bug per se. But there may be an opportunity for improvements - once I understand clearly what is going on :wink: 

Context:
- JupyterHub installation with JupyterLab as single-user server
- All single-user servers share the same disk and the same root directory. So users can share their work as they all belong to a common group.

Bug:

```
[W 2020-09-11 07:57:06.563 SingleUserNotebookApp notebookapp:1950] Error loading server extension jupyter_lsp
      File "/opt/conda/lib/python3.7/site-packages/notebook/notebookapp.py", line 1945, in init_server_extensions
        func(self)
      File "/opt/conda/lib/python3.7/site-packages/jupyter_lsp/serverextension.py", line 48, in load_jupyter_server_extension
        setup_shadow_filesystem(virtual_documents_uri=virtual_documents_uri)
      File "/opt/conda/lib/python3.7/site-packages/jupyter_lsp/virtual_documents_shadow.py", line 115, in setup_shadow_filesystem
        rmtree(str(shadow_filesystem))
      File "/opt/conda/lib/python3.7/shutil.py", line 498, in rmtree
        onerror(os.rmdir, path, sys.exc_info())
      File "/opt/conda/lib/python3.7/shutil.py", line 496, in rmtree
        os.rmdir(path)
    PermissionError: [Errno 13] Permission denied: '/homes/.virtual_documents'
```

In that configuration, I get permission error. And more importantly it seems wrong to have the same `virtual_documents` for all users. So this raises two questions:

- Would it be possible to parametrize the virtual_documents path?
- Does it need to be a child of root_dir? For what I understand, that folder contains only transient information that are better saved in a cache or temporary folder.

## Expected behavior

The virtual document folder is configurable and it lies in a cache or temporary place.

## Context

- Operating System and version: Debian strech
- Browser and version: N/A
- JupyterLab version: 2.1.5
- Jupyter-lsp 0.9.1 jupyterlab-lsp@1.1.2

> I'm unfortunately stuck to 2.1.x due to a bug in 2.2.x with nbdime.

fcollonval:
## References

Fixes #353

## Code changes

Move the ".virtual_documents" definition to a configuration in `LanguageServerManager`. Its default value is kept
but that default value can be overwritten by the environment variable `JP_LSP_VIRTUAL_DIR`

## User-facing changes

None - except if they must know about the virtual document folder.

None

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,16 @@
 ## CHANGELOG
 
-### `@krassowski/jupyterlab-lsp 2.0.9` (???)
+### `jupyter-lsp 0.9.3` (???)
 
 - features
 
-  - The virtual document folder can be configure with `JP_LSP_VIRTUAL_DIR` or  
+  - The virtual document folder can be configure with `JP_LSP_VIRTUAL_DIR` or
     `LanguageServerManager.virtual_documents_dir`. Its default value is kept
     unchanged: *contents.root_dir* / `.virtual_documents` ([#416])
+
+[#416]: https://github.com/krassowski/jupyterlab-lsp/issues/416
+
+### `@krassowski/jupyterlab-lsp 2.0.9` (???)
 
 - bug fixes
 
@@ -14,7 +18,6 @@
     robustly in files and folder names ([#403])
 
 [#403]: https://github.com/krassowski/jupyterlab-lsp/issues/403
-[#416]: https://github.com/krassowski/jupyterlab-lsp/issues/416
 
 ### `@krassowski/jupyterlab_go_to_definition 2.0.0` (???)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,19 @@
 
 ### `@krassowski/jupyterlab-lsp 2.0.9` (???)
 
+- features
+
+  - The virtual document folder can be configure with `JP_LSP_VIRTUAL_DIR` or  
+    `LanguageServerManager.virtual_documents_dir`. Its default value is kept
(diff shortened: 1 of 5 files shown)


bollwyvl:
Good stuff!

It's unreliable to use `os.path` stuff for URIs because windows... `pathlib.Path.as_uri()` might be the better approach (we basically don't support py35 anymore), or fudge it and just use `"/".join()`, though then you have to worry about trailing slashes, etc.

The hard-coded string `.virtual_documents` may also appear in some other places as well, maybe the listener.

Anyhow, we would probably need a test of this being configured and (ab)used by the front end... e.g. spaces, non-ASCII, etc. Errors in this part of the (st|h)ack are _extremely_ hard for end users to reason about, as the errors end up coming from the language servers talk about a folder that they might not even know about...

fcollonval:
Thanks @bollwyvl for the review

After a second look on the code and to take into account your remarks, do you think the following is a good idea:

```py
virtual_documents_uri =  normalized_uri(pathlib.Path(contents.root_dir) / manager.virtual_documents_dir)
```

bollwyvl:
Yes, normalized_uri would do the thing!

fcollonval:
Sorry for the useless small commits at the end.
Don't hesitate to squash & merge this one.

krassowski:
Thank you!

Issue #417, opened 2020-12-07
Documentation outdate on plugin.json


marimeireles:
## What I am trying to do...

Earlier today I was trying to configure a plugin inside jupyter-lsp using the current explanation commented on settings:
```
    // pyls: {
    //   serverSettings: {
    //     pyls: {
    //       plugins: {
    //         pydocstyle: {
    //           enabled: true
    //         },
    //         pyflakes: {
    //           enabled: false
    //         },
    //         flake8: {
    //           enabled: true
    //         }
    //       }
    //     }
    //   }
    // }
```
It didn't work.
Came to the repo and found the new way to use it on the README. Opening a PR to fix it :)

PR #418, opened 2020-12-07
Update description on how to edit plugins

marimeireles:
## References

#417 

## Code changes

Just fix the comments for people using it as a reference.

None

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/schema/plugin.json
+++ b/packages/jupyterlab-lsp/schema/plugin.json
@@ -23,7 +23,7 @@
   "properties": {
     "language_servers": {
       "title": "Language Server",
-      "description": "Language-server specific configuration, keyed by implementation, e.g: \n\npyls: {\n  serverSettings: {\n    pyls: {\n      plugins: {\n        pydocstyle: {\n          enabled: true\n        },\n        pyflakes: {\n          enabled: false\n        },\n        flake8: {\n          enabled: true\n        }\n      }\n    }\n  }\n}\n\nAlternatively, using VSCode's naming convention:\n\npyls: {\n  serverSettings: {\n    \"pyls.plugins.pydocstyle.enabled\": true,\n    \"pyls.plugins.pyflakes.enabled\": false,\n    \"pyls.plugins.flake8.enabled\": true\n  }\n}",
+      "description": "Language-server specific configuration, keyed by implementation, e.g: \n\n{\n  \"language_servers\": {\n    \"pyls\": {\n      \"serverSettings\": {\n        \"pyls\": {\n          \"plugins\": {\n            \"pydocstyle\": {\n              \"enabled\": true\n            },\n            \"pyflakes\": {\n              \"enabled\": false\n            },\n            \"flake8\": {\n              \"enabled\": true\n            }\n          }\n        }\n      }\n    }\n  }\n}\n\nAlternatively, using VSCode's naming convention:\n\n{\n  \"language_servers\": {\n    \"pyls\": {\n      \"serverSettings\": {\n        \"pyls.plugins.pydocstyle.enabled\": true,\n        \"pyls.plugins.pyflakes.enabled\": false,\n        \"pyls.plugins.flake8.enabled\": true\n      }\n    },\n    \"r-languageserver\": {\n      \"serverSettings\": {\n        \"r.lsp.debug\": false,\n        \"r.lsp.diagnostics\": false\n      }\n    }\n  }\n}",
       "default": {},
       "patternProperties": {
         ".*": {


packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -23,7 +23,7 @@
...
+      "description": "Language-server specific configuration, keyed by implementation, e.g: \n\n{\n  \"language_servers\": {\n    \"pyls\": {\n      \"serverSettings\": {\n        \"pyls\": {\n          \"plugins\": {\n            \"pydocstyle\": {\n              \"enabled\": true\n            },\n            \"pyflakes\": {\n              \"enabled\": false\n            },\n            \"flake8\": {\n              \"enabled\": true\n            }\n          }\n        }\n      }\n    }\n  }\n}\n\nAlternatively, using VSCode's naming convention:\n\n{\n  \"language_servers\": {\n    \"pyls\": {\n      \"serverSettings\": {\n        \"pyls.plugins.pydocstyle.enabled\": true,\n        \"pyls.plugins.pyflakes.enabled\": false,\n        \"pyls.plugins.flake8.enabled\": true\n      }\n    },\n    \"r-languageserver\": {\n      \"serverSettings\": {\n        \"r.lsp.debug\": false,\n        \"r.lsp.diagnostics\": false\n      }\n    }\n  }\n}",
```
krassowski, line 26: Thank you! Not sure which way of presenting it is better - please note that this is documentation for `language_servers`, so one could argue that this is not necessary to include it as a top level; on the other hand, I think it might be less confusing this way.

I will wait for other opinions.

marimeireles: Hey @krassowski I found it very confusing tbh. This is the way that's showing up on my jupytelab:
image
Which induced me to copy the code on the left side and try it on the right, but it doesn't work.

marimeireles: But ofc :)
let's see people's inputs

krassowski: I see. Thanks for highlighting it - definitely some action is needed to improve how it is displayed.

krassowski:
Maybe we should forbid any top keys different from `language_servers`? This would at least shown an error (in the bottom-right corner). Right now it accepts everything which is indeed error-prone.

krassowski:
@bollwyvl any thoughts?

bollwyvl:
I don't know _what_ else we'd like to land in here, (e.g. `debug`, etc.), but making a schema more restrictive (e.g. removing `additionalProperties: true`) after the fact is painful, and would probably require a hard version bump. There's also no in-UI way to do migrations, and the failure modes are _gross_ when things don't work properly.

I also find the JSON5 "docs" all but worthless for anything other than bools, etc. We might be better off:
- making some docs, and pointing to those instead of the pyls-specific example
- adding a bespoke UI which manipulates the settings (with knowledge of the schema from the specs), and have a short doc, e.g. `please open the Language Server Settings Pane with _Settings -> Language Servers`

packages/jupyterlab-lsp/schema/plugin.json
Sayam753: Hi @krassowski . I am a new user of this extension. Following up the instructions in default json file and writing in user preferences, did not work for me. 

I was able to use `pyls` only by wrapping it in `language_servers`, as also suggested by this PR.

krassowski:
Maybe we could add additional comment to clarify that `pyls` needs to go inside `language_servers` using description like in https://github.com/jupyterlab/jupyterlab/blob/272b7a40911a8f62f3ba225f0c8a8835ec6f5547/packages/shortcuts-extension/src/index.ts#L87-L106

mlucool:
I like @bollwyvl idea of pointing to examples rather than inline here because it's not obvious what goes into `language_servers` and then you have more room for a lot of examples. I thought I could just copy/paste what was in `//` but that turned out to be wrong (see idea 2 below).

Other ideas:
1. If you find any language server outside of `language_servers` show the user an alert. This isn't all the way to removing `additionalProperties: true`, but hints they likely got it wrong. Maybe this is more complex than just making a small breaking release.
2. Repeat yourself more to make it copy/paste friendly.

```
    // Language Server
    // Language-server specific configuration, keyed by implementation, e.g: 
    // 
    // language_servers: {
    //   pyls: {
    //     serverSettings: {
    //       pyls: {
    //         plugins: {
    //           pydocstyle: {
    //             enabled: true
    //           },
    //           pyflakes: {
    //             enabled: false
    //           },
    //           flake8: {
    //             enabled: true
    //           }
    //         }
    //       }
    //     }
    //   }
    // }
    // 
    // Alternatively, using VSCode's naming convention:
    // language_servers: {
    //   pyls: {
    //     serverSettings: {
    //       "pyls.plugins.pydocstyle.enabled": true,
    //       "pyls.plugins.pyflakes.enabled": false,
    //       "pyls.plugins.flake8.enabled": true
    //     }
    //   }
    // }
    "language_servers": {
        "pyright": {
            "serverSettings": {
                "python.analysis.useLibraryCodeForTypes": true
            }
        }
    },
```

krassowski:
Closing in favour of the new configuration UI added in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/778.

PR #423, opened 2020-12-12
Refactor jump-to functionality, remove unused code

### LINKED ISSUE #370: Add ctrl + click to jump to definition
Currently the jump to definition feature needs to be invoke from the context menu. It should be possible to press a modifier key + mouse click to jump to the definition (which as for today requires that another older plugin is installed). Instead we should have configuration of the modifier in here and retire the jump-to-definition plugin (absorbing the jump-to-definition-core without the interface bindings).

### LINKED ISSUE #372: Is it possible to jump back after jump to definition?
## What I am trying to do...

I use Ctrl-B to jump to a definition. Then I am trying to jump back to the original location. Nothing happens.

This feature was available in the go-to-definition extension (via Alt-O). On the go-to-definition readme, it is indicated that jupyterlab-lsp makes use of this package to manage a jump list. There is also issue #370. I am nut sure whether that means that there is a way to jump back which I haven't found.

I can install the go-to-definition package and use Ctrl-Alt-B and Alt-O to jump back and forth, but that does not seem to give me the same powerful definition lookup as the Ctrl-B lookup (via the language server, I presume).

## How I would like to learn how to do it...

I would like to learn whether jumping back is currently possible for the language server go to definition feature.

Thanks so much for your great work!

krassowski:
- [x] implement jump back function
  - [x] remove it from the context menu?
  - [x] flip the icon (still displayed in the command palette)
- [x] remove unused code from `@krassowski/jupyterlab_go_to_definition`; these functions were never used in `jupyterlab-lsp` and can be still accessed from original `@krassowski/jupyterlab_go_to_definition`:
   - [x] static analysis
   - [x] kernel based-jumping (might be added back, but needs to be rewritten from scratch, as not in a maintainable state)
- [x] rename `@krassowski/jupyterlab_go_to_definition` to `@krassowski/code-jumpers`
- [ ] <s>move jumpers up the source tree to the `/src`</s>
- [x] remove unused dependencies
- [x] implement <kbd>ctrl</kbd> + click jumping
  - [x] add `modifierKey` and bindings
  - [x] solve the issues preventing it from working
- [x] tests
  - [ ] <s>test opening external files (symlinked)</s> not this time
  - [x] test jumping back
  - [x] test ctrl+click jump

## References

Fixes #370, fixes #372.

## Code changes

TBD

## User-facing changes

TBD

- renamed the jumping implementation dependency to `@krassowski/code-jumpers`
- changed interfaces in `@krassowski/code-jumpers`

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


bollwyvl:
Heavy! Let me know if I can assist in any way.

krassowski:
That would be it. I will try to merge PRs tomorrow and do a release before Monday. Feel free to review (though I will likely defer  any major changes by opening issues out of comments). I am not sure how to test symlinking (only Ubuntu? all OSes?) - though not aiming to do this now (not enough time).

bollwyvl:
> Feel free to review 

:eyes: 

>  I am not sure how to test symlinking

Should work fine on unixen... wouldn't bother on windows... guess i'll see more when i get there.

If this is about the `.lsp_symlink`... maybe we need to consider a configurable way to add additional stuff, somehow... shipping our own ContentsManager seems bad, but monkeypatching (a la `jupytext` seems worse).

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 - features
 
-  - The virtual document folder can be configure with `JP_LSP_VIRTUAL_DIR` or
+  - The virtual document folder can be configured with `JP_LSP_VIRTUAL_DIR` or
```
bollwyvl, line 7: maybe "documents"?

atest/05_Features/Jump.robot
```diff
@@ -17,6 +17,23 @@ Python Jumps between Files
...
     Capture Page Screenshot    10-jumped.png
 
+Ctrl Click And Jumping Back Works
+    [Setup]    Prepare File for Editing    Python    editor    jump.py
+    ${usage} =    Set Variable    a_variable
+    ${sel} =    Set Variable    xpath:(//span[contains(@class, 'cm-variable')][contains(text(), '${usage}')])[last()]
+    Click Element    ${sel}
+    ${original} =    Measure Cursor Position
+    Capture Page Screenshot    01-ready-to-jump.png
+    Click Element    ${sel}    modifier=CTRL
```
bollwyvl, line 27: You think :apple: people will expect ⌘ ?

package.json
```diff
@@ -38,13 +38,14 @@
   "private": true,
   "scripts": {
     "bootstrap": "jlpm --no-optional --prefer-offline && lerna bootstrap && jlpm lint && jlpm clean && jlpm build",
-    "build": "jlpm build:schema && jlpm build:completion-theme && jlpm build:theme-material && jlpm build:theme-vscode && jlpm build:meta && jlpm build:ws",
+    "build": "jlpm build:schema && jlpm build:completion-theme && jlpm build:theme-material && jlpm build:theme-vscode && jlpm build:jump && jlpm build:meta && jlpm build:ws",
```
bollwyvl, line 41: honestly would love to get rid of this command, as `build:meta` _should_ do it all...

packages/code-jumpers/src/jumpers/fileeditor.ts
```diff
@@ -0,0 +1,69 @@
...
+      editor_index: null,
+      line: position.line,
+      column: position.column,
+      contents_path: this.editor.context.path,
+      is_symlink: false
+    };
+  }
+}
+
+jumpers.set('fileeditor', FileEditorJumper);
```
bollwyvl, line 69: why side-effect?

packages/code-jumpers/src/jumpers/jumper.ts
```diff
@@ -0,0 +1,162 @@
...
+
+        // protect external files from accidental edition
+        if (is_symlink) {
+          this.protectFromAccidentalEdition(document_widget);
+        }
+      })
+      .catch(console.warn);
+  }
+
+  private protectFromAccidentalEdition(document_widget: IDocumentWidget) {
```
bollwyvl, line 91: just anglophile nitpick, but perhaps "Editing"?

packages/jupyterlab-lsp/package.json
```diff
@@ -29,10 +29,11 @@
   },
   "scripts": {
     "build": "jlpm build:schema && tsc -b",
-    "build:schema": "jlpm build:schema-backend && jlpm build:schema-completion && jlpm build:schema-hover && jlpm build:schema-diagnostics && jlpm build:schema-syntax_highlighting",
+    "build:schema": "jlpm build:schema-backend && jlpm build:schema-completion && jlpm build:schema-hover && jlpm build:schema-diagnostics && jlpm build:schema-syntax_highlighting && jlpm build:schema-jump_to",
```
bollwyvl, line 32: this is getting long... could be `lerna run build:schema` and push the various schema builds into their respective packages?

bollwyvl:
Other than the OSX thing, LGTM aside from minor, unimpactful word suggestions. There's also the potential to make some of the schema building defined where it's used (rather than an increasingly long JSON line) and use `lerna` to find'm, but is not blocking.

packages/jupyterlab-lsp/package.json
krassowski: yes, good idea; maybe not on this PR because the features are not packages just yet.

bollwyvl: ha, a very good point!

packages/code-jumpers/src/jumpers/fileeditor.ts
krassowski: legacy, keeping as is for now, but needs a refactor in future.

atest/05_Features/Jump.robot
krassowski: Indeed, it seems so. Should we mention this in readme, or is it typically obvious to :apple: people?

krassowski: Interesting post: https://nuclide.io/blog/2017/02/27/Command-Click-You-Have-One-Job/
I forgot about the multi-cursor placement! Ah, this is difficult to settle. If we keep <kbd>Ctrl</kbd> as default it might break workflow for some users; at least there is a setting for this; though indeed many IDEs use <kbd>Ctrl</kbd> + click for jump, and I *think* that fewer use for placement of additional cursors - so that seems like a tie to me.

krassowski:
Changing to <kbd>Alt</kbd> + click as default to avoid breaking UX for users using multi-cursor feature.

Issue #426, opened 2020-12-14
Fails on Python older than 3.8


krassowski:
The CI tests started failing on versions of Python lower than 3.8. Needs investigation.

```python
Traceback (most recent call last):
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/jsonschema/__init__.py", line 31, in <module>
    from importlib import metadata
ImportError: cannot import name 'metadata'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/importlib_metadata/_compat.py", line 8, in <module>
    from typing import Protocol
ImportError: cannot import name 'Protocol'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/jupyterlab_server/server.py", line 20, in <module>
    from notebook.notebookapp import aliases, flags, NotebookApp as ServerApp
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/notebook/notebookapp.py", line 85, in <module>
    from .services.contents.manager import ContentsManager
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/notebook/services/contents/manager.py", line 17, in <module>
    from nbformat import sign, validate as validate_nb, ValidationError
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/nbformat/__init__.py", line 32, in <module>
    from .validator import validate, ValidationError
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/nbformat/validator.py", line 12, in <module>
    from .json_compat import get_current_validator, ValidationError
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/nbformat/json_compat.py", line 10, in <module>
    import jsonschema
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/jsonschema/__init__.py", line 33, in <module>
    import importlib_metadata as metadata
  File "/usr/share/miniconda/envs/jupyterlab-lsp/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 15, in <module>
    from ._compat import (
[...]

PR #427, opened 2020-12-14
Investigate mitigation to failure on Python < 3.8

krassowski:
Investigating #426 

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/requirements/github-actions.yml
+++ b/requirements/github-actions.yml
@@ -5,6 +5,7 @@ channels:
   - defaults
 
 dependencies:
+  - typing_extensions
   - jupyterlab {lab}
   # build dependencies
   - nodejs {nodejs}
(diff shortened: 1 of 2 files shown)


bollwyvl:
Bringing our own `typing_extensions` seems fine, as we _do_ import it... I don't know what might have dropped it as a transient dependency.

krassowski:
But we only import it in type checking mode:

```python
if TYPE_CHECKING:  # pragma: no cover
    from typing_extensions import Protocol
```

so I guess safe to close this PR and leave as is (dependency for dev, but not in install_requires)?

bollwyvl:
Yeah this was resolved at runtime upstream with importlib[-.]metadata. If
we're not seeing any other fails, we can just close this... If someone _is_
type checking that isn't us 🤷

PR #429, opened 2020-12-16
Hotfix crash "blank screen" likely related to Mac command character

### LINKED ISSUE #428: Jupyter lab interface crash
After installing the plugin and rebuild, my jupyter lab interface can not work. It looks like this:
image

Does anyone know how to solve the problem?

krassowski:
It seems that the nice Mac Command character (⌘) was breaking JupyterLab extension system because of poor handling of encoding. Probably not worth fixing in lab because of the new extension system so just removing the offending character.

## References

Hopefully fixes #428.

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -14,7 +14,7 @@ defaults:
 env:
   # TODO extract these from files instead
   PY_JLSP_VERSION: 0.9.3
-  JS_JLLSP_VERSION: 2.1.0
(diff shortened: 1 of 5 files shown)


bollwyvl:
So it's Unicode is why we can't have nice things!

Cheers on the quick turn, will see to the conda-forge business.

bollwyvl:
oh, cool, npm only, so i _won't_ be doing any conda-forging.

PR #433, opened 2020-12-16
Add highlights debounce

### LINKED ISSUE #431: Highlights do not always update with the cell focus
## Description

highlights_do_not_awlays_wsitch_with_cell

## Expected behavior

Should always update highlights.

## Context

- Chrome
- JupyterLab 2.2
- JupyterLab-LSP 2.1.1

### LINKED ISSUE #432: Highlights are not debounced leading to annoying delays in display
## Description

no_debounce_for_highlights

## Expected behavior

Highlights should be delayed slightly until the cursor settles.

## Context

- Chrome
- JupyterLab 2.2
- JupyterLab-LSP 2.1.1

krassowski:
Improves performance of completion and highlights by minimising the number of highlight requests and GUI redraws (token checking, debouncing, acting on a single response only). Also fixes highlights not being updated on focus/blur.

## References

Fixes #432 and fixes #431

## Code changes

Added debouncer and multiple checks verifying if we want to ask server for highlights and then if we want to apply those.

## User-facing changes

Highlights do not flicker when moving cursor quickly. Completions work slightly faster.

None

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
TODO:
- [ ] convert logs to debug calls; add debug option in settings and only display logs when it is set to `true`.
   - should it be one global switch, or should it be per feature?

PR #448, opened 2020-12-31
Implement caching of shadow virtual file handles

### LINKED ISSUE #447: Cache file handles rather than opening them each time (virtual documents)
## What are you trying to do?

The current virtual documents implementation opens file for read and then write for each document modification. This takes time, even for me running on Linux and SSD. This may be a bigger performance issue for Windows users, especially with HDD. We can cache the file handles in a Python dictionary.

## What is new in your approach and why do you think it will be successful?

My preliminary benchmarking shows that opening file once reduces the entire operation from ~1 ms to ~0.7 ms. The differences are presumably more notable  on non-SSD drives and on Windows with the default NTFS where operations are known to have a higher cost than on the default Linux and Mac file systems (sources: my own experience, [this discussion from MS WSL project](https://github.com/Microsoft/WSL/issues/873#issuecomment-425272829), countless [SO](https://stackoverflow.com/a/55785595) [posts](https://stackoverflow.com/a/48356618/6646912)).

## Who cares? If you are successful, what difference will it make?

I expect that this small performance increase will compound to a noticeable improvement for users with HDD and Windows, especially when multiple LSP server queries are being sent in a row, e.g. when quickly typing code in the editor with autocomplete enabled, which is a critical feature and often pointed out as having poor performance by the users. Of note, the recent work on highlights debouncing and token checking reduced the number of queries sent to the LSP server when typing quickly allowing for better performance already (see #433).

## What are the risks?

It requires a caching mechanism for `EditableFile`, including:
- a way to close files upon exit: easy (`atexit`),
- a way to purge closed files from cache (or risk a memory leak): hard

## What are the mid-term and final “exams” to check for success?

The CI robot tests runs on average noticeably faster on Windows (~1% time decrease is a success)

krassowski:
Opening the PR to trigger CI and see if the speed up is worth fighting for.

## References

Fixes #447.

## Code changes

- [x] Cache file handles instead of opening them each time.
- [ ] Cache invalidation and deletion of unused files on `textDocument/didClose` (TODO)

## User-facing changes

- Potential performance gain, likely only visible for HDD/Widows user

None

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/py_src/jupyter_lsp/tests/test_virtual_documents_shadow.py
+++ b/py_src/jupyter_lsp/tests/test_virtual_documents_shadow.py
@@ -35,17 +35,6 @@ async def test_read_missing(tmp_path):
 
 
 @pytest.mark.asyncio
-async def test_read_deleted(tmp_path):
-    path = tmp_path / "deleted.py"
-    deleted_file = EditableFile(path)
-    path.unlink()
-
-    await deleted_file.read()
-
-    assert deleted_file.lines == [""]
-
-
-@pytest.mark.asyncio
 async def test_apply_change(tmp_path):
     # inserting text
     path = tmp_path / "test.py"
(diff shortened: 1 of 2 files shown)


krassowski:
Closing as we are moving towards not needing these files are all.

PR #451, opened 2021-01-01
Release v2.1.2

krassowski:
Release time/happy new year!


krassowski:
Anaconda is down, need to wait if want to have a final check with CI...

PR #402, opened 2020-11-09
JupyterLab 3 update

martinRenou:
Add support for JupyterLab 3

- [x] completion theme extension
- [ ] jupyterlab-go-to-definition
- [ ] jupyterlab-lsp
- [ ] lsp-ws-connection
- [ ] metapackage
- [ ] theme-material
- [ ] theme-vscode

Issue #400, opened 2020-11-05
JupyterLab 3 update


martinRenou:
## Motivation

JupyterLab 3 is coming, with a new extension system. I've been updating multiple packages around, and I have the need for this extension to be updated as well. 

Is there anyone working on it already? Or should I come with a PR myself?

krassowski:
You are welcome to PR! I am happy to help and review, though do not have time to work on it myself this month. I expect that upgrade to 3.0 should solve quite a few issues. We use a little bit more complex build system and have plugins depend on each other, so it might not be as easy as with other packages, so your help as of someone with experience in upgrades would be more than appreciated!

FYI, I just merged #381 to prevent potential merge conflicts and make sure CI will pass.

bollwyvl:
Yeah... This sucker is not lightweight, and we're happy for more hands and
fewer forks, while trying to keep shipping updates for users. I have been
trying to do _simple_ things with cm vs jl3rc, and it's already tough.

Please let us know how we can help... If a meeting/workshop makes sense, i
can carve out a day.

Some thoughts:

A big thing in this will be a thorough assessment of what we can share with
the debugger.

Another freaky wrinkle is the now-bitrotten proxy kernel pr... I think we
generally agree we want it, but a lot has happened since I got to work on
it. The wins here are that it would:
- remove the notebook dependency, clearing the path for jupyter_server
compat
- remove any hard runtime vscode is deps
- remove home-rolled websocket connection management in favor of comms,
which mostly just work
- decouple the locality/perf/lifecycle of the tornado server from the
language servers
- make it more "jupyter-like", as in more interactive and inspectable
- provide a path for willing kernels to go hybrid and provide lsp over a
well known target

At the expense of:

- adding a hard (initial) ipython or (eventual) xeus dep
- trading server extension problems for kernel spec ones

Related: having in-browser language servers will further reduce our
reliance on npm/nodejs for some use cases... e.g lab settings editor. But
making contentsmanager look like fs will be... Hard.

I don't see any of the above as a hard blockers to getting something out,
but might make sense to consider them in this work, as we'll have a unique
opportunity during this migration... e.g. byebye py35!

martinRenou:
[...]

PR #452, opened 2021-01-02
JupyterLab 3 support v2

krassowski:
Probably should have rebased rather than merging, but let's roll with this. Opening as new PR for now to spin the CI, but we might sent a PR to #402 instead. Relates to #400.

```
python -m pip install -e . --ignore-installed --no-deps -vv
jupyter serverextension enable --py jupyter_lsp
jupyter labextension develop . --overwrite
```

Punchlist:
- [x] setup new branch for JupyterLab 2.x and make RTD build another version of docs for it so that we can point users who will stick with 2.x for the time being (https://github.com/krassowski/jupyterlab-lsp/tree/2.x)
- [x] fix the docs
- [x] decide whether to publish both npm and python package or only one of those
    - probably npm is needed too as otherwise confused users will see message that this version of lab is not supported and we have no way of telling them from the interface that they need to use pip only this time.
- [x] typescript 4.1.3
- [x] eslint:
```
    "@typescript-eslint/eslint-plugin": "^4.8.1",
    "@typescript-eslint/parser": "^4.8.1",
    "eslint": "^7.14.0",
    "eslint-config-prettier": "^6.15.0",
    "eslint-plugin-jest": "^24.1.3",
    "eslint-plugin-prettier": "^3.1.4",
    "eslint-plugin-react": "^7.21.5",
```
- [x] In the CONTRIBUTING.md:
    - [x] the proposed `pip install .` needs to be invoked after every change, after compilation. This is extremely tedious and needs automation. [`jupyter labextension build` and `jupyter labextension develop . --overwrite`
](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#developing-a-prebuilt-extension) to the rescue?
- [ ] releasing instructions update
- [x]  update issue templates
- [ ] verify on [Binder](https://mybinder.org/v2/gh/krassowski/jupyterlab-lsp/jupyterlab3?urlpath=lab%2Ftree%2Fexamples%2FPython.ipynb)


krassowski:
A small conda dependency hell for Python 3.6 it looks like. Did JupyterLab 3.0 drop 3.6?

CONTRIBUTING.md
```diff
@@ -29,11 +29,11 @@ You can contribute to the project through:
 
 Development requires, at a minimum:
 
-- `nodejs >=10.12,<15`
+- `nodejs >=12,<15`
 - `python >=3.6,<3.9.0a0`
   - Python 3.6 and 3.8 are also tested on CI
```
krassowski, line 34:
```suggestion
  - Python 3.7 and 3.8 are also tested on CI
```

krassowski:
Locally getting warnings:

- [ ] ServerApp] 'LanguageServerManager' object has no attribute 'virtual_documents_dir'
- [ ] jupyter_lsp | extension was found and enabled by nbclassic. Consider moving the extension to Jupyter Server's extension paths.

krassowski:
- [ ] In the CONTRIBUTING.md:
    - [ ] the proposed `pip install .` needs to be invoked after every change, after compilation. This is extremely tedious and needs automation. [`jupyter labextension build` and `jupyter labextension develop . --overwrite`
](https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#developing-a-prebuilt-extension) to the rescue?
- [ ] releasing instructions update

bollwyvl:
In the `doit`-based automation I've been doing (ha) around lab3 on [other extensions](
https://github.com/jupyrdf/ipyelk/pull/62/files), the approach I'm somewhat happy with is:
- do a "normal" install by default (just to ensure it works)
- encapsulate the whole `watch` process in a separate command, out at the python level, which does the appropriate setup, runs the watches, etc. then kills both of them when either dies

I've been finding this easier to reason about.

On CI: node on windows (and even OSX) is boring when it breaks, as now we don't really care if webpack works on those platforms. I've been building the wheel/sdist up-front in a `build` job, and installing them in the `test` job. This helps catch subtle file omissions that the symlink thing sweeps under the covers. We may want a separate `package.json` of _just_ our npm-distributed language servers, which would be the only thing that's installed on the various test boxen, but wouldn't be installed on the `build` box. On a local dev enviornment, you'd need both, of course.

Testing: the modal command palette is kinda flaky under test (10% of the time, it seems to just disappear, depending on what has focus when invoked). The original sidebar behavior that is easy to test (and works with our existing stuff) can be restored by settings:

```robotframework
    Create File
    ...    ${SETTINGS DIR}${/}@jupyterlab${/}apputils-extension${/}palette.jupyterlab-settings
    ...    {"modal": false}
```

bollwyvl:
> A small conda dependency hell for Python 3.6 it looks like. Did JupyterLab 3.0 drop 3.6?

this weirdness:
https://github.com/conda-forge/jupyter_server-feedstock/blob/master/recipe/meta.yaml#L19

Will get on it...

bollwyvl:
Have some PRs up, one of which I can't merge (`jupyter_server`) but hopefully can be resolved quickly...

krassowski:
I could not fix an issue with karma-webpack (tried updating to 5.0-alpha and updating everything, and tried few other workarounds):

```js
yarn run v1.21.1
$ karma start
02 01 2021 15:08:10.767:ERROR [preprocess]: Can not load "webpack", it is not registered!
  Perhaps you are missing some plugin?

START:
02 01 2021 15:08:10.803:ERROR [karma-server]: Error during file loading or preprocessing
TypeError: process is not a function
    at executeProcessor (/~/jupyterlab-lsp/node_modules/karma/lib/preprocessor.js:47:11)
    at runProcessors (/~/jupyterlab-lsp/node_modules/karma/lib/preprocessor.js:60:23)
    at FileList.preprocess [as _preprocess] (/~/jupyterlab-lsp/node_modules/karma/lib/preprocessor.js:134:11)
    at async Promise.all (index 0)
    at async /~/jupyterlab-lsp/node_modules/karma/lib/file-list.js:90:11
    at async Promise.all (index 2)
02 01 2021 15:08:10.804:INFO [karma-server]: Karma v5.2.3 server started at http://localhost:9876/
02 01 2021 15:08:10.804:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
02 01 2021 15:08:10.805:ERROR [karma-server]: Error: Found 1 load error
    at Server.<anonymous> (/~/jupyterlab-lsp/node_modules/karma/lib/server.js:213:27)
    at Object.onceWrapper (events.js:421:28)
    at Server.emit (events.js:327:22)
    at emitListeningNT (net.js:1341:10)
    at processTicksAndRejections (internal/process/task_queues.js:83:21)
error Command failed with exit code 1.
```

Feel free to enable it if you have ideas how to make this work again. It would be a shame to trash the tests inherited from the original author (and I do not feel like rewriting them to jest right now).

bollwyvl:
move to jest?

krassowski:
Could it be that it uses wrong karma (should go for `packages/lsp-ws-connection/node_modules/karma` but uses `node_modules/karma`)? How would I fix that?

bollwyvl:
Yeah, i don't have strong feelings on nodejs-based test stuff... i trust you've rebuilt the whole yarn.lock... are we pulling multiple versions? does it need `resolutions` of (more) things?

krassowski:
It seems that it's just one version. I think I was wrong but on a good track:
- `karma` gets installed in `node_modules/`, but
- `karma-webpack` gets installed in `packages/lsp-ws-connection/node_modules`

so karma does not see the webpack plugin. Now I am not sure how to reconcile them.

bollwyvl:
hoist karma-webpack to root `package.json`? there are some fiddly bits that can control it in a more fine-grained fashion but :shrug: ?

bollwyvl:
and, of course, #278 dispenses with that package altogether...

krassowski:
That's it from me for today. The CI is still red, but I was able to get it running locally. I think that I needed to enable serverextension with sys prefix in order for it to work, and I am still getting `extension was found and enabled by nbclassic` (for some reason `data_files` from setup do not end up installed when using `pip .` - I don't know where is problem).

bollwyvl:
Awesome work! I'll give it some eyes. Don't sweat nbclassic stuff.
Data_files is a damn mess, not your fault. Like I said, if we run atest
against the real packages, we'll be able to make sure we don't need any
assets.

README.md
```diff
@@ -77,9 +77,9 @@ Use context menu on rows in the panel to filter out diagnostics or copy their me
 
 You will need to have all of the following installed:
 
-- JupyterLab >=2.2.0,<3.0.0a0
+- JupyterLab >=3.0.0,<4.0.0a0
 - Python 3.6+
-- nodejs 10.12+
+- nodejs 12+
```
bollwyvl, line 82: Someone _just_ running `pyls` or `r-languageserver` won't need this at all :tada:!

packages/jupyterlab-lsp/package.json
```diff
@@ -28,48 +28,63 @@
...
+    "clean": "jlpm run clean:lib",
+    "clean:all": "jlpm run clean:lib && jlpm run clean:labextension",
+    "clean:labextension": "rimraf labextension",
+    "clean:lib": "rimraf lib tsconfig.tsbuildinfo",
+    "eslint": "eslint . --ext .ts,.tsx --fix",
+    "eslint:check": "eslint . --ext .ts,.tsx",
+    "install:extension": "jupyter labextension develop --overwrite .",
     "lab:link": "jupyter labextension link . --no-build",
-    "test": "jlpm jest --coverage --coverageReporters=cobertura --coverageReporters=html --coverageReporters=text-summary"
+    "prepare": "jlpm run clean && jlpm run build:prod",
```
bollwyvl, line 53: this is one of those "magic" scripts, and gets run at "interesting" times... i recommend dropping it.

packages/jupyterlab-lsp/package.json
```diff
@@ -28,48 +28,63 @@
...
-    "@jupyterlab/docmanager": "~2.2.0",
-    "@jupyterlab/docregistry": "~2.2.0",
-    "@jupyterlab/fileeditor": "~2.2.0",
-    "@jupyterlab/notebook": "~2.2.0",
-    "@jupyterlab/rendermime": "~2.2.0",
-    "@jupyterlab/services": "~5.2.0",
-    "@jupyterlab/statusbar": "~2.2.0",
-    "@jupyterlab/testutils": "~2.2.0",
-    "@jupyterlab/tooltip": "~2.2.0",
+    "@jupyterlab/application": "^3.0.0",
```
bollwyvl, line 69: we should probably hoist (at least) this to `dependencies`...

setup.py
```diff
@@ -4,17 +4,26 @@
 
 import setuptools
 
+
+LABEXTENSIONS_DIR = Path('py_src') / 'jupyter_lsp' / 'labextensions'
+LABEXTENSIONS_INSTALL_DIR = Path('share') / 'jupyter' / 'labextensions'
+
+
+def get_data_files():
+    extension_files = [
```
bollwyvl, line 13: `data_files` is an awful API. You have to give it each _directory_, so if there's any nesting, it doesn't do what one would expect. Here's [one approach](https://github.com/jupyrdf/ipyelk/pull/62/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7).

bollwyvl:
Also, the upstream conda-forge PRs are coming along... should have something where we can drop the jedi pin, soon, and hopefully the `jupyter_server` one gets reviewed soon...

bollwyvl:
At least part of the CI fail:

```
Run jupyter labextension install --debug --no-build packages/jupyterlab-lsp/krassowski-jupyterlab-lsp-2.1.2.tgz
```

Don't have to do this anymore :tada:, and even better, don't have to do `jupyter lab build` :partying_face: .

krassowski:
So the key to enabling the extension with jupyter_server and not notebook is to add a space: `jupyter server extension enable` not `jupyter serverextension enable`. Subtle.

Also it turns out that installation from a wheel works ok with data, but `jupyter serverextension list` returns:

```
config dir: /~/.pyenv/versions/3.7.9/envs/jupyterlab-lsp-lab3/etc/jupyter
    jupyterlab  enabled 
    - Validating...
      jupyterlab 3.0.0 OK
```

whereas: `jupyter server extension list`:

```
Config dir: /~/.jupyter

Config dir: /~/.pyenv/versions/3.7.9/envs/jupyterlab-lsp-lab3/etc/jupyter
    jupyter_lsp enabled
    - Validating jupyter_lsp...
      jupyter_lsp 0.9.3 OK
    jupyterlab enabled
    - Validating jupyterlab...
      jupyterlab 3.0.0 OK
    nbclassic enabled
    - Validating nbclassic...
      nbclassic  OK

Config dir: /usr/local/etc/jupyter
```

So jupyterlab install itself to both notebook and server. The `jupyter_server` migration guide also [shows such an example](https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html#migrating-an-extension-to-use-jupyter-server). However, lab extension cookicutter only enables [itself for server, not notebook](https://github.com/jupyterlab/extension-cookiecutter-ts/blob/master/%7B%7Bcookiecutter.python_name%7D%7D/setup.py).

I think that it should be ok.

New punchlist items:
- [ ] update issue template to use `jupyter server extension` for 3.0
- [ ] update manual enable in docs and readme to use `jupyter server extension` for 3.0

bollwyvl:
Yeah, that's among the reasons I don't use the cookiecutter.

bollwyvl:
Also, let's go ahead and bump `jupyter_lsp` to `1.0.0`, and add a hard dep on `jupyterlab >=3.0,<4`, I kinda don't trust the pip solver much these days, especially for upgrades...

bollwyvl:
The work's paying off... my homebrew automation scripts have gotten me up to robot... it cracked a browser instead of using the headless one...

krassowski:
Opened https://github.com/jupyterlab/jupyterlab/issues/9539 to as lab extension (by similarity to server extension) kind of works and sooner or later someone is gonna use it.

bollwyvl:
Yeah, the problem is these are detected via `entry_points`, and it's bad business if two projects try to provide the same one...

krassowski:
> it cracked a browser instead of using the headless one...

Got it, forgot to rename `jupyter_notebook_config.json` to server in one place

krassowski:
So there is this failure:
```
Hover works in notebooks                                              | FAIL |
Keyword 'Trigger Via Hover With Modifier' failed after retrying 4 times. The last error was: WebDriverException: Message: TypeError: rect is undefined
------------------------------------------------------------------------------
Hover can be triggered via modifier key once cursor stopped moving    | FAIL |
Keyword 'Trigger Via Modifier Key Press' failed after retrying 5 times. The last error was: WebDriverException: Message: TypeError: rect is undefined
```

It would seem familiar for `Click Element`, but here it happens on `Mouse Over` and the element is definitely visible (except that `Element Is Not Visible` indicates its not). It is visible on screenshots and when debugging in Chrome but when debugging in robot it has size 0 x 0 **but** after painting blue rectangle over it it magically recovers its dimensions and works properly.

```robot
    ${width}  ${height} =   Get Element Size   ${sel}
    Run Keyword If  ${width} == 0   Cover Element   ${sel}    # don't know why but otherwise it raises Message: TypeError: rect is undefined
    Mouse Over    ${sel}
```

this is probably one of the weirdest workarounds I did ni a while, but it works. Any ideas what might be the root issue?

bollwyvl:
that is a doozy... i am familiar with/dread the `rect is undefined` thing, and usually it's just a timing thing that needs longer. Seems to sometimes be related to react virtual rendering, etc. However, CodeMirror usually seems to have stuff on the page... i guess you could force a `refresh` or `repaint` or whatever with the CM API? At any rate, good to know about `Cover`.... very sneaky!

bollwyvl:
Quick thing, probably need to add `py_src/jupyter_lsp/labextensions/` to `.gitignore`.

Less quick: I don't know the magic trick to make `jupyter_server` be the thing launched by binder.... will do some looking...

krassowski:
> Quick thing, probably need to add py_src/jupyter_lsp/labextensions/ to .gitignore.

Already there.

Tests are now passing :tada: (well, not 3.6 but others do).

krassowski:
@bollwyvl do you think that using `jupyter-lsp` as an umbrella for lab extension and server extension is ok, or should we try to split it into `jupyter-lsp` (server) and `jupyterlab-lsp` (lab)? I guess better to make this decision now rather than later.

I see both pros (one less thing to install, one less thing users will be confused about) and potential cons (in a distant future one may wish to provide lsp for other jupyter frontents and have frontend-specific packages).

krassowski:
I think that having a separate `jupyterlab-lsp` (which depends on `jupyter-lsp`) will also avoid some confusion during upgrade for existing users. I now think that's the way to go.

bollwyvl:
Yeah, two packages sounds good for now.

So maybe something like
- jupyterlab-lsp 3.0, requires
  - jupyter-lsp >=1.0.0,<2.0.0, requires
    - jupyter_server >=1.1.3
  - jupyterlab >=3.0.0,<4.0.0

Basic install instructions wouldn't even mention jupyter_lsp, probably.

krassowski:
What about organization in repo:

```
/
|-- docs/
|-- packages/   <- typescript stuff
|-- python_packages/
   |-- jupyter_lsp/
   |  |-- src/*
   |  |-- setup.py
   |  `-- setup.cfg
   |-- jupyterlab_lsp/
   |  |-- src/
   |    |-- __init__.py
   |    `-- install.json
   |  |-- setup.py
   |  `-- setup.cfg
```

?

bollwyvl:
Looks about right! I'm happy to take a crack at the re-org, if you're getting a little tired of the exercise: just let me know. As we hit more and more packages, I'm pining for `doit`...

bollwyvl:
Also just got the first full local build of this through :tada: !

python_packages/jupyter_lsp/setup.py
```diff
@@ -13,7 +13,7 @@
     data_files=[
         (
             "etc/jupyter/jupyter_server_config.d",
-            ["py_src/jupyter_lsp/etc/jupyter-lsp-serverextension.json"],
+            ["/etc/jupyter-lsp-serverextension.json"],
```
bollwyvl, line 16: pretty sure it can't start with a `/`...

krassowski: :laughing:

krassowski:
The new jupyterlab-lsp package means that we need some work on conda side too - I will appreciate your help over there, I would be terribly slow learning how to do stuff properly on conda.

bollwyvl:
Don't worry, we got you. Also: the layout should (unfortunately) be `python_packages/jupyter{,lab}_lsp/src/jupyter{,lab}_lsp`...

krassowski:
Flag remained in setup.cfg, not sure what is going on:

```
python_packages/jupyter_lsp/src/non_blocking.py:12: error: Module 'ctypes' has no attribute 'WinError'
python_packages/jupyter_lsp/src/non_blocking.py:12: error: Module 'ctypes' has no attribute 'windll'
```

Removed flag jupyterlab on purpuse, I guess needs fixing because jupyter_lsp should not depend on lab:

```
python_packages/jupyter_lsp/src/types.py:244: error: Skipping analyzing 'jupyterlab': found module but no type hints or library stubs
```

bollwyvl:
Yeah, see earlier comment: you're now shadowing the built-in `types`, etc. if you installed those into your pyenv... you're probably going to have to rebuild it :crying_cat_face:

bollwyvl:
> `python_packages/jupyter_lsp/src/types.py:244: error: Skipping analyzing 'jupyterlab': found module but no type hints or library stubs`

That's probably for `pageConfig`?

krassowski:
Ok, I got what you mean with organization. But nah, the ctypes should not be shadowed in that case...

krassowski:
for node roots:

```python
    @default("node_roots")
    def _default_node_roots(self):
        """get the "usual suspects" for where `node_modules` may be found

        - where this was launch (usually the same as NotebookApp.notebook_dir)
        - the JupyterLab staging folder (if available)
        - wherever conda puts it
        - wherever some other conventions put it
        """

        # check where the server was started first
        roots = [pathlib.Path.cwd()]

        # try jupyterlab staging next
        try:
            from jupyterlab import commands

            roots += [pathlib.Path(commands.get_app_dir()) / "staging"]
        except ImportError:  # pragma: no cover
            pass

        # conda puts stuff in $PREFIX/lib on POSIX systems
        roots += [pathlib.Path(sys.prefix) / "lib"]

        # ... but right in %PREFIX% on nt
        roots += [pathlib.Path(sys.prefix)]

        return roots

```

krassowski:
I guess it is fine, because it is optional...

bollwyvl:
> for node roots:

Yeah, that's trash, and won't even generally be relevant anymore. Kill it with :fire: !

python_packages/jupyter_lsp/src/non_blocking.py
```diff
@@ -9,8 +9,8 @@
 
 if os.name == "nt":  # pragma: no cover
     import msvcrt
-    from ctypes import POINTER, WinError, byref, windll, wintypes
-    from ctypes.wintypes import BOOL, DWORD, HANDLE
+    from ctypes import POINTER, WinError, byref, windll, wintypes  # type: ignore
```
bollwyvl, line 12: :+1:

krassowski:
It looks (mostly) green again. Did you mean to send a PR against this branch rather than master? I would say safe to rebase now.

bollwyvl:
I wanted to kick CI a few times before re-targeting this branch...

krassowski:
Conda is fussy again `CondaHTTPError: HTTP 504 GATEWAY TIME-OUT for url <https://conda.anaconda.org/nodefaults/linux-64/current_repodata.json>`.

That's it from me for today.

bollwyvl:
Again, heroic work. I'll see about setting up some higher retries and timeouts and stuff with a `.github/.condarc` over on #454, hopefully you'll have nice stuff to look at tomorrow.

bollwyvl:
Py36 should be unblocked within the hour... If not, at least I can do
something directly about it now!

bollwyvl:
hooray green. checking out binder (added to description).

> decide whether to publish both npm and python package or only one of those 

I feel it's important to keep publishing to `npm` so people can extend things. Let's not make another (worse) walled garden on pip.

bollwyvl:
ugh, i forgot how _ugly_ it is if binder detects that you have r stuff, and goes off and does `apt` r, rstudio, shiny... I mean, it's all great stuff, but is not really germane to _our_ demo. I guess we change the name of our `environment.yml`, and run it our own selves in `postBuild`...

bollwyvl:
Binder looks good enough! Since we don't do a `lab build` :tada: it's pretty snappy (once it gets to our stuff), but does pop the dreaded `build recommended` because of offline-notebook. Yeah, well, that ain't part of our demo either, so i think we just uninstall it.

Because it _does_ run, I think we hold off any heavy work on it.

Similarly, since we've got working CI, we could start a `develop` branch or something, make that a CI-enabled target, merge this there, and start working in parellel on smaller tasks with more meaningful diffs, as this is getting to be a bit of a monster. I just don't think we want to be blowing up RTD and the README.md with Instructions From The Future...

krassowski:
Great to hear that! 3.0 is out and if the announcement gets out tomorrow then it would probably be fine to merge to master. I added link to RTD for 2.x in README; but indeed there are two things to resolve:
- publish on PyPI (should we publish rc first? may need versioning adjustments as npm uses different suffixes than PyPI)
- publish on conda-forge (does it need a PyPI package first?)

bollwyvl:
> publish on conda-forge 

Again, don't worry, let's just get something out we feel good about that the interested _can_ install, through some means other than standing up our dev env. Unless we think we're going to split the repo, we can keep running it out of the existing feedstock... it's already complicated as hell (with the language servers, etc.) best we don't make things that need multiple PRs.

> (does it need a PyPI package first?)

Yeah, we _do_ need a pypi upstream sdist. It's kind of a _thing_ to get a custom `rc` "label" (like a sub-channel). However, if an CF rc helps us be more confident, we can certainly do it. But it's really no big deal if conda folk have to wait a sec, as folk generally don't use the `rc`s much, anyway, if you _do_ set it up.

> link to RTD for 2.x in README

Perhaps we call that `stable`? There's a lotta stuff in the RTD admin UI, but I know it's possible. I think it can build individual tags as well (you have to "activate" them or something). We're just fighting confusion and support issues from folks that don't read Big Bold Notes.

Anyhow, I'm happy to help out in any way (in addition to condachores, of course).

krassowski:
So here is the release candidate 0 on PyPI to unblock the conda work and reserve name: https://pypi.org/project/jupyterlab-lsp/

krassowski:
And the announcment of 3.0 went live about an one hour ago. I would say no need to get rc on conda, it's just good to have rc on PyPI. Let me know if I can help with the conda work and lets merge this to master and iterate on smaller things then.

bollwyvl:
Oh, neat. Here's the process, I'll get on it: https://github.com/conda-forge/cfep/blob/master/cfep-05.md

bollwyvl:
Anyhow, started framing it up, but not going to push too much harder on it this morning. LMK when there are releases, and I'll get on it!

bollwyvl:
Also we want _some_ kind of test for the new package... just checking whether it's installed/the files exist/etc...

martinRenou:
I completely missed this PR in my notifications, thanks a lot for the work! Tell me if you need anything :)

krassowski:
Thank **you** for starting the effort well in advance.

krassowski:
RC is live: https://pypi.org/project/jupyter-lsp/1.0.0rc0/

bollwyvl:
Some notes:
- we should probably update the tagline
```
Multi-Language Server WebSocket proxy for Jupyter Notebook/Lab server
                                                          ^
                                                    Maybe just "Server"?
```

bollwyvl:
`jupyter-lsp >1.0.0` does not match `1.0.0rc0`, which fails `pip check` (and likely means that someone would _not_ get anything if they tried to `pip install`)

- we should add this to `integrity`
- we should add `pip check` to CI

bollwyvl:
- the `rc`s went out without tags
  - we need to decide how we're going to do tags in the future, e.g. jupyterlab-like
    - `v3.0.0` is
      - `release/jupyterlab-lsp/1.0.0`
      - `release/jupyter-lsp/2.0.0`
    - or give up, and do CalVer...

bollwyvl:
https://github.com/conda-forge/jupyter-lsp-feedstock/pull/8

bollwyvl:
Theoretically should work (within the hour):

[Binder](https://mybinder.org/v2/gist/bollwyvl/76a720a27be20aec815d171227dc28cd/HEAD?urlpath=lab%2Ftree%2Fenvironment.yml)

bollwyvl:
Well, it built, but the serverextension is not enabled, because binder is running under `notebook` + `nbclassic`, not straight-up `jupyter_server`. I don't think this is going to be a long-term problem, so I'm willing to roll with it.

krassowski:
Added integrity checks and removed rc tags. If it passes I will upload new version and merge.

Where would the `pip check` go?

krassowski:
Added `pip check` to `job.test.yml` though earlier I was thinking about keeping it in build (which would require to install packages there too I guess)

bollwyvl:
> Added pip check to job.test.yml

Yeah, sorry, mean to actually PR these things, got caught up in local election coverage :blush:.

Also, as we ran into a similar packaging issue over on `jupyterlab-classic` once it got to conda (this happens _all the time_), I PRd some stuff to do a relatively lightweight check over there... I can adapt it here.

Basic idea: do _just_ a run of the packages, as built, with _just_ `setup-python` (no node, conda) and do the CLI smoke tests, as they seem. Will get a PR up, but might not be able to finish it.

bollwyvl:
#456 smoke tests are green, so probably ready to merge in here. Slowest one was 6 minutes, so not going to bother with caching pip.

lresende:
Any eta for a working rc?

krassowski:
1-2 hours for PyPI release, some time more until conda packages catch up. I will do one final push and publish the new release once CI passes.

lresende:
Thank you @krassowski

krassowski:
@lresende the release is fully out now. Please see: https://github.com/krassowski/jupyterlab-lsp/releases/tag/v3.0.0 for update instructions

PR #454, opened 2021-01-03
[squash] Jupyterlab3 CI updates

bollwyvl:
## References

- hopefully can be cherry-picked (with squash) into #452
  - just this diff: https://github.com/krassowski/jupyterlab-lsp/compare/jupyterlab3...bollwyvl:jupyterlab3-ci 

### Packaging/metdata
- moves more files in-tree in packages
  - was getting some weird stuff when installed as-packaged
- updates `install.json` and `jupyterlab-lsp/package.json` to point at the new package name

### CI
- adds a `build` step to do the all the distribution builds, uploads for downstream jobs
  - nice side-effect: two commands can upload everything from CI-built assets :tada:
  - `+~4min` before any `atest` jobs start
- uses `--find-links` and `--no-index` to install `whl`s without knowing the version
  - removes this check from integrity
- skips js builds altogether on atest
  - windows: `-~3min`
- moves js unit tests to lint (since it already has to do a lot with js stuff)
- uses mambaforge
  - windows: `-~2min`
- tries to cache `node_modules`, falls back to `.yarn-packages`
  - windows: `-~4min`
- removes the `jedi` and `tectonic` cache
  - these weren't doing anything positive because we overload `$HOME`
    - bringing them back would require more research into how to specify these locations with env vars or config, but would very probably reduce some robot timeouts
- updates binder `postBuild` to (probably) work
  - haven't tested
     
> timings above are:
>   - vs the last run of #452
>   - with a hot cache

## User-facing changes

n/a

n/a

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


krassowski:
I was thinking about making release workflow as separate from test, with some fancy trigger.

bollwyvl:
> I was thinking about making release workflow as separate from test, 

I think the _actual_ uploading is a separate concern. With this approach, I'm just ensuring we are testing _exactly_ what we _would_ be shipping. The release step _could_ pull from this artifact, though I still am not super-excited about putting _my_ pypi/npm creds into CI, no matter the encryption, etc. Another approach is to create github "draft" releases from tags, and then 

> with some fancy trigger.

If you _do_ want to put creds in, a nice approach is posting a "draft" github release, which you can then Push The Button to promote it, and have that make the magic happen.

krassowski:
Right. Though it does not *that* bad - no need for full credentials, just API token which may be limited in scope to only one project, as per https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

bollwyvl:
Right, but I like those creds to only be available _to the release job_, and not to every random npm post-install script and github action.

bollwyvl:
Calling it for the day. I updated the pr description with all the stuff I've been fiddling with. This is probably ready to be merged into #452, if there's nothing overly-offensive happening, but haven't retargeted as:
- i'll happily fix anything you don't like 
- I don't know if re-targeting would kill the currently-running jobs which would have the most updated timings

Still no update on https://github.com/conda-forge/jupyter_server-feedstock/pull/34 for py36 builds, maybe the PR fairy will come over night :fairy: .

bollwyvl:
Timings look good... getting close to 10 minutes saved on the slowest windows excursion, and inching us back from the dreaded hour. 

Future versions of `setup-miniconda` will obviate a bunch of the hackery this introduces, as well.

Without doing more entropy plumbing with make/doit (e.g. pre-solved lock files), I don't think I can get it down much more... I guess the `build` _could_ use `setup-node` and `setup-python`, which could shave another  minute. Robot waiting `Until Keyword Suceeds` is definitely the long pole: I've found pabot to be _awesome_ on linux/osx (up to about 4 processes on CI, basically linear speedup, though the server gets stood up more) but windows can't seem to handle it, so wouldn't move the needle much.

Will re-target, and will be off open source stuff (but occasionally checking email) for the next $DAY_JOB worth of hours... I've no idea what started on fire over the holidays!

python_packages/jupyterlab_lsp/jupyterlab_lsp/install.json
```diff
@@ -1,5 +1,5 @@
 {
   "packageManager": "python",
-  "packageName": "jupyter_lsp",
-  "uninstallInstructions": "Use your Python package manager (pip, conda, etc.) to uninstall the package jupyter_lsp"
+  "packageName": "jupyterlab_lsp",
+  "uninstallInstructions": "Use your Python package manager (pip, conda, etc.) to uninstall the package jupyterlab_lsp"
```
krassowski, line 4: Good catch

packages/jupyterlab-lsp/package.json
```diff
@@ -130,7 +129,7 @@
     "discovery": {
       "server": {
         "base": {
-          "name": "jupyter-lsp"
+          "name": "jupyterlab_lsp"
```
krassowski, line 132: Well, if I publish another npm release and someone installs from npm, `jupyter-lsp` would suffice. Otherwise they would not see this anyway.

.github/workflows/job.test.yml
```diff
@@ -12,25 +13,188 @@ defaults:
...
+      - name: Publish builds
+        uses: actions/upload-artifact@v2
+        with:
+          name: jupyterlab-lsp dist ${{ github.run_number }}
+          path: ./dist
+
   acceptance:
     runs-on: ${{ matrix.os }}-latest
-    name: ${{ matrix.os }} py${{ matrix.python }} node${{ matrix.nodejs }}
+    name: atest ${{ matrix.os }} py${{ matrix.python }}
```
krassowski, line 196: Not a fan of `atest` abbreviation, but we already use it so it can stay I guess.

krassowski:
Great work, I like the idea with building once! I restarted the CI to make sure it works with 3.6.

packages/jupyterlab-lsp/package.json
bollwyvl: yeah, who knows. just thought of it while fixing the other one...

.github/workflows/job.test.yml
bollwyvl: we can call it whatever you like! but shorter is better so we can get _some_ info off the UI. `robot`, `browser`, etc.

bollwyvl:
omg so green :exploding_head:

.github/workflows/job.test.yml
bollwyvl: if they could be emoji, i'd be all for :robot:

PR #456, opened 2021-01-06
Add pip-only install/CLI smoke test

bollwyvl:
## References

- for https://github.com/krassowski/jupyterlab-lsp/pull/452#issuecomment-755289990
  - will re-target to #452 once green

## Code changes

Adds a `pip`-only install test to CI (doesn't even do a checkout). By making these rely on `lint` _and_ `build`, I'm hoping the acceptance jobs start first, so that this doesn't add to the bottom-line wall-clock time... even still, may want to cache pip, etc.

The existing `pip check` in the acceptance job is _still_ good to keep, as we want to know if our test dependencies have drifted or conflict with each other.

Ideally we'd _just_ install the `jupyterlab-lsp` tarball, to ensure the `install_requires` is configured properly, but I don't know how to make pip do that _and_ not pull in old versions of a locally-built dependency.

## User-facing changes

N/A, but should help non-dev installs go more smoothly.

N/A

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -20,6 +20,7 @@ env:
 
   PYTHONUNBUFFERED: 1
   ATEST_RETRIES: 3
+  PIP_DISABLE_PIP_VERSION_CHECK: 1
 
   # TODO: replace once mambaforge "just works" with setup-miniconda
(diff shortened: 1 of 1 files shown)


bollwyvl:
Ok, all the smoke tests are green, will re-target

PR #459, opened 2021-01-07
make websocket server to send ping frame to client

### LINKED ISSUE #458: websocket server do not send a ping frame to client, make websocket frequently lost connection throught a reverse proxy like nginx
## Description

Hi~:
Notebook websocket will send ping frame periodically, by default the interval is 30 sec.
https://github.com/jupyter/notebook/blob/1550410229e302ecf74d5bf82d5caba6e6296a64/notebook/base/zmqhandlers.py#L162

but lsp websocket will not. bcz the handler override the open method.
https://github.com/krassowski/jupyterlab-lsp/blob/c9ba03bd85cef7b8f923099205bf7b0d958f3892/python_packages/jupyter_lsp/jupyter_lsp/handlers.py#L25

From my view. I think it's very necessary to send a ping frame periodically.bcz in production environment,the request are ofen forwarded through a reverse proxy like nginx, and alway nginx is set with a read timeout rule.If there is no ping frame, the websocket connection may not transmit any data for a long time, and then trigger ngnix timeout rule, the ws connection will be force closed by nginx.
 
From nginx document:

> By default, the connection will be closed if the proxied server does not transmit any data within 60 seconds. This timeout can be increased with the proxy_read_timeout directive. Alternatively, the proxied server can be configured to periodically send WebSocket ping frames to reset the timeout and check if the connection is still alive. 

**the recommendation is the set websocket server to periodically send WebSocket ping frames~**

so how about read the notebook configuration and send the ping frame?

## Reproduce

use jupyterlab-lsp with nginx together

## Expected behavior

websocket server read the notebook configuration and send the ping frame like notebook websokcet handler
## Context

- Operating System and version: any
- Browser and version: any
- JupyterLab version: any

<pre>
≈
</pre>
</details>

<details><summary>Required: installed lab extensions</summary>
<pre>
≈
</pre>
</details>

<details><summary>Troubleshoot Output</summary>
<pre>
≈
</pre>
</details>

<details><summary>Command Line Output</summary>
<pre>≈
</pre>
</details>

<details><summary>Browser Output (recommended for all interface issues)</summary>
<pre>

</pre>
</details>

franckchen:
## References

try to fix #458 

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
Thank you for the PR! Our test mocks were not prepared to handle more advanced features of the websocket mixin, but I just improved them and added a test case to make sure it works. I will merge if the CI tests pass.

krassowski:
I wonder if we also need `pong` from the other side.

krassowski:
There is an interesting possibly pyls-related failure on Python 3.6 on Ubuntu, but not on Mac nor Win:

```
Element 'css:div.lsp-statusbar-item' did not get text 'Fully initialized' in 1 minute.

Also teardown failed:
Phrases found in /home/runner/work/jupyterlab-lsp/jupyterlab-lsp/atest/output/linux_36_1/lab.log: {'pyls_jsonrpc.endpoint - Failed to handle request': True}
```

krassowski:
(The close/reopen is to restart the CI)

franckchen:
> from the other side

No. I think we dont need care about pong. Web browser will automatically reply pong message, dont require developer to code about it. Browser even dont provide api for developer to send pong to server.I dont know if this is what you are worried about.

python_packages/jupyter_lsp/jupyter_lsp/tests/test_session.py
```diff
@@ -70,3 +70,30 @@ async def test_start_unknown(known_unknown_server, handlers, jsonrpc_init_msg):
 
     assert not manager.sessions.get(ws_handler.language_server)
     assert_status_set(handler, {"not_started"})
+
+
+@pytest.mark.asyncio
+async def test_ping(handlers):
+    """see https://github.com/krassowski/jupyterlab-lsp/issues/458"""
+    a_server = "pyls"
```
krassowski, line 78: I just do not think it is needed to run this check for all language servers (just wanted to have something to pass to `ws_handler.open()`) but did not want to create a mock either.

python_packages/jupyter_lsp/jupyter_lsp/handlers.py
```diff
@@ -26,6 +26,7 @@ def open(self, language_server):
         self.language_server = language_server
         self.manager.subscribe(self)
         self.log.debug("[{}] Opened a handler".format(self.language_server))
+        super().open()
```
bollwyvl, line 29: very good catch. might we even want to `super.open` first, before any overloaded behavior?

python_packages/jupyter_lsp/jupyter_lsp/tests/test_session.py
bollwyvl: hm, yeah, maybe we need a real pygls mock server at some point :(

python_packages/jupyter_lsp/jupyter_lsp/tests/test_session.py
```diff
@@ -70,3 +70,30 @@ async def test_start_unknown(known_unknown_server, handlers, jsonrpc_init_msg):
...
+
+    handler, ws_handler = handlers
+    manager = handler.manager
+
+    manager.initialize()
+
+    assert ws_handler.ping_interval > 0
```
bollwyvl, line 85: tests are looking good

python_packages/jupyter_lsp/jupyter_lsp/tests/test_session.py
krassowski: I am fine with bringing pygls as a mock dependency in the future.

PR #465, opened 2021-01-10
Add support for jupyterlab-classic

### LINKED ISSUE #460: Support jupyterlab-classic
## Elevator Pitch

jupyterlab-classic allows to use modern lab components with the classic notebook user experience. It is the perfect way to provide the LSP features for users who prefer the classic UI.

## Motivation

Improve the UX for users who do not enjoy lab layout.

## Design Ideas

It is already possible to install, but it does not work yet because of hard dependencies on:
- statusbar
- file editor

we need to make those optional for the package to work with the classic UI.

Issue in jupyterlab-classic: https://github.com/jtpio/jupyterlab-classic/issues/72

Discussion items:
- [ ] where to put the status indicator and how it should look like (aim to reuse the status bar item for simplicity)
- [ ] what to do with diagnostics panel:
   - hide it altogether?
   - open in a new browser tab as a main area widget (how about syncing then though)?
- [ ] how do we test classic? Should we have a robot suite with a minimal test of the basic features?

krassowski:
## References

Fixes #460.

In notebook:
- works:
  - hover
  - diagnostics
  - completion
  - signatures
  - highlights (slow? pyls issues?)
  - status item (shown on notebook toolbar)
- does not (yet) work:
  - rename of variables (or is it me not having the pyls plugin enabled?)
  - diagnostics panel (disabled by app name detection)
  - jump to definition (works for symlinked files only?)

In addition, in the file editor:
  - completion does not wok, the completer is disabled

## Code changes

- Switched some plugins from `requires` to `optional`.
- Generalized `labShell` to `shell`
- Added icons
- Moved status item to toolbar of notebook if statusbar not available
- Fixed icons highlighting in status item

## User-facing changes

The extension will now work with JupyterLab-classic

Screenshot from 2021-01-13 01-01-47

#### Fixed icon highlight on select:

| before | after |
|--------|-------|
| Screenshot from 2021-01-12 22-02-38  | Screenshot from 2021-01-12 22-00-47 |
| Screenshot from 2021-01-12 22-02-09 | Screenshot from 2021-01-12 22-01-50  |

None

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [x] changelog entry


packages/jupyterlab-lsp/src/adapter_manager.ts
```diff
@@ -26,12 +27,12 @@ export class WidgetAdapterManager implements ILSPAdapterManager {
...
   }
 
-  constructor(protected labShell: ILabShell) {
+  constructor(protected shell: ILabShell | IClassicShell) {
     this.adapterChanged = new Signal(this);
     this.adapterDisposed = new Signal(this);
     this.adapterTypeAdded = new Signal(this);
     this.adapterTypes = [];
-    labShell.currentChanged.connect(this.refreshAdapterFromCurrentWidget, this);
+    shell.currentChanged.connect(this.refreshAdapterFromCurrentWidget, this);
```
jtpio, line 35: Wondering whether `WidgetAdapterManager` could maybe take a more generic `JupyterFrontEnd.IShell`, and then check for either `ILabShell` or `IClassicShell` to be able to use the `currentChanged` signal.

Otherwise it could maybe default to `currentWidget`.

bollwyvl: I think, at best, we can keep special cases (in core) for JupyterLab.

If we find ourselves needing more stuff, it probably needs to be a separate extension, and we figure out the common extension point. But if it can be entirely generic, _mazel tov_.

bollwyvl:
> (statusbar item needs to be shown somewhere)

A sticky wicket, for sure, but again (as I generally run without, when getting to actually _use_ jupyterlab) it isn't too bad to use with without... it doesn't actually let you _change_ anything, right?

Anyhow: we could likely get something to work that embedded in the notebook toolbar. As usual, I worry about that area becoming a "trash bin" that starts scrolling because there are so many buttons with huge labels. 

So if we can get to `[Icon] [5+/-2 characters] v` we can probably something in the notebook bar, and probably also add to the editor bar. Add to settings, e.g. `"showStatusInActivityToolbar": ["Notebook", "Editor"]` or a couple switches, fine.

> Diagnostics

Yeah, just having to going to have to do without, i guess, as a dialog would make no sense.

bollwyvl:
So obviously, there's a lot of CSS that's relying on being part of the status bar, but i think it gets the idea across:

Screenshot from 2021-01-10 21-16-49

But I think the idea is sound, if we can get to something that:
- works in status bar
- works in toolbar
- preferably has good keyboard access, etc.
- is a bit smaller
we'll be in a good spot. I don't think that would block this PR, though. 

Also, anecdata: it was super smooth adding `jupyterlab-classic` to `binder/environment.yml` and seeing @krassowski's _almost_ just work with @jtpio's gives me hope for this whole federated extensions goosechase!

jtpio:
> Also, anecdata: it was super smooth adding `jupyterlab-classic` to `binder/environment.yml` and seeing @krassowski's _almost_ just work with @jtpio's gives me hope for this whole federated extensions goosechase!

Yes, that's what I love about this new extension system!

krassowski:
@bollwyvl do we want to have a dedicated acceptance test for this? How difficult it would be to start up `classic` for one suite instead of `lab`? Or should we merge it as an experimental feature for now and only worry about it if the idea actually takes of an people start using it?

krassowski:
I guess `[Icon]` without any text would be ok, as long as we actually change the icon instead of the colour itself (like a nod towards colourblind people). On contrary I prefer to keep the text in the statusbar (but this still can be a single widget, just initialized with a different option).

bollwyvl:
Re: ui, yeah. Again, wouldn't block on "solving" the ui problem, if getting
something in quick can get people trying it. Really need gutters probably
for diagnostics... Debugger forces them... But probably won't do anything
in classic lol. Good luck to whoever tries to shoehorn that into classic
(been there, no thanks).

Robot: once installed, classic will already be available on its own URL...
Shouldn't require any server setup changes, just a "Go To" to a fixture
notebook URL and an alternate "wait for splash/ready" to know when it's
ready (e.g some diagnostic shows) since we don't have status bar. Worst
case is Zen mode, as there is nowhere but the toolbar and the editors. And
settings.... Pffffbt.

jtpio:
> Really need gutters probably
for diagnostics... Debugger forces them... But probably won't do anything
in classic lol.

Right, in jupyterlab-classic the debugger won't be loaded and shouldn't show the gutters then.

bollwyvl:
I think Zen mode with (at least line number) gutters, but a lot of other things turned down, is actually a pretty pleasant place to be:

Screenshot from 2021-01-12 16-11-44

If that gutter was wider, and showed `circle, optionally with number` for warnings and `octagon, optionally with number` for errors, I might turn off the `squiggles` altogether, or only expose them on hover:

Screenshot from 2021-01-12 16-18-47

krassowski:
What a shame CodeMirror does not support gutters on the right. Or does it? It could save use potential conflicts with debugger (or just having a very wide gutter) .

krassowski:
Anyways, I added the status item to the notebook toolbar, hidden the text, and added two extra icons ('m not fully satisfied with the code-clock appearance, but we can change it later) to make distinctions more obvious. Completions in files need PR over in the juptyerlab-classic.

@jtpio it would be very nice if there was a way to get available areas (or lack of thereof) from app of `app.shell` so that the diagnostics command can know that it should not be enabled/visible.

packages/jupyterlab-lsp/src/adapter_manager.ts
krassowski: I think that interface checks on an object at runtime are not supported in typescript to check for `IClassicShell`?  I could check if `app.shell` is an instance of `ClassicShell` but then we would depend on classic and we would not want to force users to install it.

jtpio: Otherwise passing the `shell`, `labShell`, and `classicShell` as initialization options for the constructor:

```typescript
interface IOptions {
  shell: JupyterFrontEnd.IShell;
  labShell?: ILabShell;
  classicShell?: IClassicShell;
}
```

With `labShell` and / or `classicShell` possibly `null` (`optional` in the plugin definition)

jtpio:
> it would be very nice if there was a way to get available areas (or lack of thereof) from app of app.shell so that the diagnostics command can know that it should not be enabled/visible.

Maybe the `shell.widgets(area)` could throw an exception if an area doesn't exist?

At the moment it doesn't, but it looks like it should (I think lab does it):

https://github.com/jtpio/jupyterlab-classic/blob/57d7000390eebd7457235be2433d364808fc624b/packages/application/src/shell.ts#L171-L179

jtpio:
> Maybe the `shell.widgets(area)` could throw an exception if an area doesn't exist?

FYI @krassowski, I just released `jupyterlab-classic==0.1.2` with this change, and also exposing the `completer-extension:files` plugin for text editors:

https://github.com/jtpio/jupyterlab-classic/releases/tag/0.1.2

Which should fix a couple of items mentioned in https://github.com/jtpio/jupyterlab-classic/issues/72

packages/jupyterlab-lsp/package.json
```diff
@@ -65,6 +65,7 @@
   },
   "devDependencies": {
     "@babel/preset-env": "^7.4.3",
+    "@jupyterlab-classic/application": "^0.1.1",
```
jtpio, line 68: To grab the latest changes:

```suggestion
    "@jupyterlab-classic/application": "^0.1.2",
```

jtpio: To be followed by a `yarn.lock` update.

jtpio: Ah thanks, good catch :+1: 

It should be available now.

PR #474, opened 2021-01-15
Fix connection manager loosing track of notebooks

### LINKED ISSUE #469: Inconsistent state for the status bar indicator
## Description

The status bar indicator might sometimes report the `Waiting...` state, while it looks like it is connected and functional.

This seems to happen when switching between notebooks.

## Reproduce

1. Open the `Python.ipynb` example notebook
2. Check the first example
3. Open a new notebook
4. Switch between the two notebooks

waiting-statusbar

## Expected behavior

Instead of `Waiting...` the item would switch back to `Fully Initialized`.

## Context

- Operating System and version: Ubuntu 20.04
- Browser and version: Latest Chromium
- JupyterLab version: 3.0.4

krassowski:
which were not correctly scoped by path by switching away from `id_path` which was refactored to not contain the actual file path to `uri` which does contain the file path and `id_path`.

## References

Fixes #469

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## CHANGELOG
 
-### `@krassowski/jupyterlab-lsp 3.0.1` (unreleased)
+### `@krassowski/jupyterlab-lsp 3.1.0` (unreleased)
 
 - features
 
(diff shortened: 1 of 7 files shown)


jtpio:
Thanks!

PR #477, opened 2021-01-16
Fix truncated stream

### LINKED ISSUE #450: Large websocket responses get truncated and cannot be parsed
## Description

Sometimes a very long response from the language server gets truncated and the relevant action is not performed. This appears to be somewhat random, because subsequent attempts can sometimes succeed.

```R
Uncaught SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at i.readMessage (index.js:formatted:1)
    at index.js:formatted:1
    at WebSocket.e.onmessage (index.js:formatted:1)
readMessage @ index.js:formatted:1
(anonymous) @ index.js:formatted:1
e.onmessage @ index.js:formatted:1
```

The message truncates at less than 7k characters

Similar issue was reported in a different project using PHP backend here: https://github.com/beyondcode/dusk-dashboard/issues/20.

## Reproduce

1. Open an R notebook
2. Type `base::` (potentially press <kbd>Tab</kbd>, but should auto-invoke)
4. See lack of completion suggestions and error in console

## Expected behavior

1. We may want to try to recover truncated JSON messages by parsing the viable portion (e.g. for completion it would make sense to show something rather than nothing, especially for extra long namespaces such as `base` in R)
2. The error should not happen in the first place

## Context

- Operating System and version: Ubuntu 20.10
- Browser and version: Chrome
- JupyterLab version: 2.2.x

master on 2.1.2

krassowski:
## References

Fixes #450, enabling parsing of large responses from R server (e.g. completion on base) and any responses from javascript/typescript language servers

## Code changes

Turns out `BufferedIOBase.read()` does buffering... And we need to read in chunks until we consume all needed chunks.

## User-facing changes

Completion for R works better, completion for javascript works. May also help some Python packages with huge namespaces, e.g numpy. It will also improve perceived completion speed.

long_namespace_completion_works

## Chores

- [x] linted
- [x] tested - TODO a robot test for R base completion.
- [ ] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,11 @@
 
   - send periodic pings on websocket channels to maintain connection ([#459], thanks @franckchen)
   - R languageserver is no longer incorrectly shown as available when not installed ([#463])
+  - fix completion of very large namespaces (e.g. in R's base or in JavaScript) due to truncated message relay ([#477])
 
 [#459]: https://github.com/krassowski/jupyterlab-lsp/pull/459
 [#463]: https://github.com/krassowski/jupyterlab-lsp/pull/463
+[#477]: https://github.com/krassowski/jupyterlab-lsp/pull/477
 
 ### `@krassowski/jupyterlab-lsp 3.0.0` (2021-01-06)
 
--- a/atest/05_Features/Completion.robot
+++ b/atest/05_Features/Completion.robot
@@ -203,6 +203,14 @@ Completes Correctly With R Double And Triple Colon
     Wait Until Keyword Succeeds    40x    0.5s    File Editor Line Should Equal    3    datasets:::.packageName
     [Teardown]    Clean Up After Working With File    completion.R
 
+Completes Large Namespaces
+    [Setup]    Prepare File for Editing    R    completion    completion.R
+    Place Cursor In File Editor At    6    7
+    Wait Until Fully Initialized
+    Trigger Completer
+    Completer Should Suggest    abs    timeout=30s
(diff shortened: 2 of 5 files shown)


bollwyvl:
Awesome fix! I should probably figure out a python-side unit testing
approach for more than just "does the server turn on"... While the
acceptance tests are testing it, we'd probably get better insight
(potentially benchmarks) from something with at least a _few_ layers not in
the mix. Perhaps this is where the mock pygls server comes in, as we can
more carefully tune our "worst case" examples, e.g. malformed messages, etc.

krassowski:
It did not solve the issue on Windows which is slightly more complicated there.

On Windows we get `OSError: [Errno 22] Invalid argument` which seems to be coming from the non-blocking stream implementation (https://stackoverflow.com/questions/34504970/non-blocking-read-on-os-pipe-on-windows/34504971#34504971). Do we need the stream to be non-blocking? Should we rewrite it using `tornado.iostream`?

Also, the non-blocking behaviour has issues in Python: https://bugs.python.org/issue13322

krassowski:
Also why do we use the `BufferedIOBase` in the first place? Could it be causing the weird performance issue for Window users? Just bouncing ideas I had when looking at the code....

bollwyvl:
As long as we are running in the same event loop as the server, these
really do need to be non-blocking.

Now that we've dropped python 3.5 we might have some more options... I seem
to recall that being a fairly major roadblock. The standard lib stuff may
be enough... jupyter_server now has anyio as a dependency, which has some
higher-level abstractions we also might be able use.

Issue #272, opened 2020-05-19
Slow completion for large Python packages when using the default pyls


jhettler:
## Description

Autocompletion - especially of bigger packages like pandas is slow.

## Reproduce

1. Go to Jupyterlab
2. Write `import pandas as pd`
3. Write `pd.<tab>` 
4. It takes seconds in Jupyterlab, in classic notebook it is immediately autocompleted.

Is it usual behavior or there is something wrong configured?

## Expected behavior

Autocompletion finished nearly immediately :)

## Context

- Operating System and version:  _Linux b832f09e7bfa 5.4.0\-7626-generic #30\~1588169883\~20.04\~bbe668a\-Ubuntu SMP  x86_64 x86_64 x86_64 GNU/Linux_ - Docker
- Browser and version: _Chrome Version 81.0.4044.138 (Official Build) (64-bit)_
- JupyterLab version: _Jupyterlab version 2.1.0 (same behavior was with 2.1.2)_

<details><summary>Required: installed server extensions</summary>
<pre>
config dir: /home/jovyan/.jupyter
    jupyterlab_git  enabled 
config dir: /opt/conda/etc/jupyter
    jupyter_lsp  enabled 
    jupyterlab  enabled 
    jupyterlab_git  enabled 
    nbdime  enabled 
    - Validating...
      jupyterlab_git 0.20.0 OK
    - Validating...
      jupyter_lsp 0.8.0 OK
    - Validating...
      jupyterlab 2.1.0 OK
    - Validating...
      jupyterlab_git 0.20.0 OK
    - Validating...
      nbdime 2.0.0 OK
</pre>
</details>

<details><summary>Required: installed lab extensions</summary>
<pre>
   app dir: /opt/conda/share/jupyter/lab
        @aquirdturtle/collapsible_headings v2.2.0  enabled  OK
        @bokeh/jupyter_bokeh v2.0.0  enabled  OK
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        @jupyterlab/git v0.20.0  enabled  OK
        @krassowski/jupyterlab-lsp v1.0.0  enabled  OK
        jupyter-matplotlib v0.7.2  enabled  OK
        jupyterlab-plotly v1.5.4  enabled  OK
        jupyterlab-python-file v0.4.0  enabled  OK
        nbdime-jupyterlab v2.0.0  enabled  OK
        plotlywidget v1.5.4  enabled  OK
[...]

PR #487, opened 2021-01-23
Completer improvements: kernel type mapping and dynamic documentation

### LINKED ISSUE #333: Provide mapping between some kernel types and LSP types?
Currently the multiple sources of completion do not always play great together. For example, the kernel-provided instances:

Screenshot from 2020-09-03 19-21-33

which makes the completer icon field wider and the icon appear aligned to left rather than centered. As what matters is the UX, we could map some of the kernel completion types which do not correspond 1-1 to completion kinds to the closest reasonable match if such exists. For `instance` this would be `variable` or `value`.

This should be configurable as the user may be using a kernel we do not know of and therefore have different completion types provided.

krassowski:
## References

- provides mechanism for mapping between some kernel types and LSP types for icons selection, fixes #333
  - motivated by IJulia returning very different types than LSP types and IPython returning `<unknown>` if it cannot match a type
- implements `completionItem/resolve` to allow for:
  - display of documentation panel with R language server which does not send documentation in the first request but requires `completionItem/resolve`
  - faster completion (#272) once pyls or jedi-language-server implement `completionItem/resolve` too
  - [x] store original completionItem and resolve using exactly it?
  - [x] add test with R language server
  - [ ] open issues on pyls and jedi proposing to implement `completionItem/resolve`; create a PR in the repo that will accept such a contribution (help welcome!)
- makes use of newly exposed `Completer.IRenderer`, for implementation of active item tracking and implements makrdown rendering of documentation in `Completer.IRenderer`
- add support for deprecation via `CompletionItemTag.Deprecated`
  - [x] declare tag support
  - [ ] check if any supported server implements it and if yes, test it; checked that do not support: jedi, julia, pyls, R; maybe javascript/typescript server? 

Related upstream PR:
- https://github.com/jupyterlab/jupyterlab/pull/8930
- https://github.com/jupyterlab/jupyterlab/pull/9643
- https://github.com/jupyterlab/jupyterlab/pull/9663

## Code changes

Added `LazyCompletionItem` class implementing `CompletionHandler.ICompletionItem` but also doing much more to workaround the limited flexibility of upstream (see the self-reference trick to workaround https://github.com/jupyterlab/jupyterlab/pull/9643). The new object-oriented code reduced complexity of the completion handler making it considerable more pleasant and enabled implementation of `completionItem/resolve`.

## User-facing changes

- Documentation panel renders markdown if server provided markdown
  - only for `completionItem/resolve` retrieved docs for now
  - for all servers once https://github.com/jupyterlab/jupyterlab/pull/9663 gets merged and released
- Documentation panel works with R :tada: (and other language servers that use `completionItem/resolve`, e.g. javascript/typescript)

R-docs

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ## CHANGELOG
 
-### `@krassowski/jupyterlab-lsp 3.1.1` (unreleased)
+### `@krassowski/jupyterlab-lsp 3.2.0` (unreleased)
+
+- features:
+
+  - documentation panel in completer now works for R language too: implemented `completionItem/resolve` ([#487])
+  - kernel types returned by IPython and IJulia are now mapped to LSP types; you can customize the mappings in settings ([#487])
 
 - bug fixes:
 
@@ -14,6 +19,7 @@
   - `PythonModuleSpec` no longer raises exception when the server module does not exist ([#485])
 
 [#485]: https://github.com/krassowski/jupyterlab-lsp/pull/485
+[#487]: https://github.com/krassowski/jupyterlab-lsp/pull/487
 
 ### `@krassowski/jupyterlab-lsp 3.1.0` (2021-01-17)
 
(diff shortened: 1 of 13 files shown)


krassowski:
The lint failure seems to be due to the changes in the new mypy release (0.800). [changelog](http://mypy-lang.blogspot.com/2021/01/mypy-0800-released.html).

PR #488, opened 2021-01-24
Fix IO issues with long messages on windows

krassowski:
## References

Follow-up after #477 - the added acceptance tests revealed that the non-blocking mode on windows prevents messages longer than one chunk from being passed. After reading more on the Python bugs issue tracker and on the SO it appears that:
- non-blocking on Windows will not work with buffered
- buffered has bugs and is not expected to play well with non-blocking even on Unix

There are two way to deal with it:
1. remove the non-blocking mode - preferable assuming it has no side effects 
2. remove the buffered IO base (replace it with a non-buffered one)

Let's see how (1) works on CI across OSes

## Code changes

In current state drops non-blocking mode for IO streams when reading messages from language servers.

## User-facing changes

- Completion for very large namespaces works on Windows too.
- Potential performance improvement

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


krassowski:
I now see a note from @bollwyvl saying that we should probably keep non-blocking as long as we are running in the same event loop as server (https://github.com/krassowski/jupyterlab-lsp/pull/477#issuecomment-766253193). Well, still useful to give CI a spin IMO.

bollwyvl:
Here's the `anyio` abstraction I couldn't remember:

https://anyio.readthedocs.io/en/stable/streams.html

## pro
- _if_ they've solved the stdio streams issue, in-loop, on-windows, we would _really_ want to use their implementation instead of my cargo-culted mess
- things that work in `anyio` will work with other IO loop implementations... there are a number of possible directions, e.g. `trio`, `curio`, `uvloop`
  - the same is true of `ipykernel`, and _presumably_ `xeus-python`
  - also considering alternate pythons (e.g. `pypy3`) it might make sense to have a (tested) `jupyter-lsp[performance]` install profile that trades the simplicity of pure python dependencies for the performance of compiled ones
    - while `pypy3` is not _generally_ available on windows, JSON parsing and validation would be other natural places to eke out a bit more performance on that platform

## con
- starting to write anyio code (vs `tornado` or `asyncio`) would rapidly make backports untenable.

krassowski:
Disabling the buffering solve the issue. It was merged with #494. It might introduce a problem for a very specific case when the message length exceeds your memory size, but that would be an issue anyways on the frontend or in the actual server.

PR #494, opened 2021-01-25
Attempt to fix truncated Jedi LSP responses

dpriedel:
## References

#450

## Code changes

This patch attempts to fix the truncated Jedi LSP responses which
produce warning messages about content-length mismatch. It does so
by looping around and trying again upon an empty read. The theory
is that the content producer is just not ready with more data yet.

- add function parameter 'max_empty_parts_in_a_row' with a default
value of 50.  This will be decremented on each empty read.
- change the default value of the 'max_parts' parameter to 100
since we should get larger chunks on each successful read.

Either of these values could be adjusted based on experience.

### DIFF
--- a/python_packages/jupyter_lsp/jupyter_lsp/stdio.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/stdio.py
@@ -61,7 +61,7 @@ class LspStdIoReader(LspStdIoBase):
 
     @default("max_wait")
     def _default_max_wait(self):
-        return 0.5 if os.name == "nt" else self.min_wait * 2
+        return 0.1 if os.name == "nt" else self.min_wait * 2
 
     async def sleep(self):
         """Simple exponential backoff for sleeping"""
@@ -138,7 +138,8 @@ class LspStdIoReader(LspStdIoBase):
             raw = b"".join(raw_parts)
             if len(raw) != length:  # pragma: no cover
                 self.log.warning(
-                    f"Readout and content-length mismatch:" f" {len(raw)} vs {length}"
+                    f"Readout and content-length mismatch: {len(raw)} vs {length};"
+                    f"remaining empties: {max_empties}; remaining parts: {max_parts}"
                 )
 
         return raw
--- a/python_packages/jupyter_lsp/jupyter_lsp/stdio.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/stdio.py
@@ -98,9 +98,10 @@ class LspStdIoReader(LspStdIoBase):
                 await self.sleep()
 
     def _read_content(
-        self, length: int, max_parts=100, max_empty_parts_in_a_row=50
+        self, length: int, max_parts=100, max_empties=50
     ) -> Optional[bytes]:
-        """Read the full length of the message unless exceeding max_parts.
+        """Read the full length of the message unless exceeding max_parts or
+           max_empties empty reads occur.
 
         See https://github.com/krassowski/jupyterlab-lsp/issues/450
 
@@ -119,13 +120,11 @@ class LspStdIoReader(LspStdIoBase):
         raw_parts: List[bytes] = []
         received_size = 0
         while (
-            received_size < length
-            and len(raw_parts) < max_parts
-            and max_empty_parts_in_a_row > 0
+            received_size < length and len(raw_parts) < max_parts and max_empties > 0
         ):
             part = self.stream.read(length)
             if part is None:
-                max_empty_parts_in_a_row -= 1
+                max_empties -= 1
                 continue  # pragma: no cover
             received_size += len(part)
             raw_parts.append(part)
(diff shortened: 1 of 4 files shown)


python_packages/jupyter_lsp/jupyter_lsp/stdio.py
```diff
@@ -116,10 +116,11 @@ def _read_content(self, length: int, max_parts=1000) -> Optional[bytes]:
         raw = None
         raw_parts: List[bytes] = []
         received_size = 0
-        while received_size < length and len(raw_parts) < max_parts:
+        while received_size < length and len(raw_parts) < max_parts and max_empty_parts_in_a_row > 0:
             part = self.stream.read(length)
             if part is None:
-                break  # pragma: no cover
+                --max_empty_parts_in_a_row
```
krassowski, line 122: `-= 1`? `--` gives + in python

bollwyvl:
Thanks for the PR :heart:! This has been a bear, and we really appreciate the help.

Aside from the text thing above (please try `python -m scripts.lint from the repo`, see what it says), we definitely need to put this under a representative test, otherwise it's hard to keep it fixed.

We don't have a concrete example in the repo yet, but we should figure out a test that:

- spins up _at least two_ language servers, so we don't fool ourselves on the blocking aspects, especially on windows
  - they _can_ be the same implementation, just different spec names, i guess
- repeatedly generates a big enough response to demonstrate this condition 

@krassowski any thoughts? I'm most comfortable trying to test vs pyls, but I know you had some good gotchas with R...

krassowski:
I was thinking about just copying the large response from R server that I can locally reproduce. An example from @dpriedel of an action that triggers the problem with pyls would be very helpful too.

I am not sure if the buffer length always stays constant (DEFAULT_BUFFER_SIZE), this would need to be accounted for when designing a unit tests; we could simulate a delay using some sleep in a thread.

On a more general note, and as mentioned elsewhere the `BufferedIOBase` seems to be troublesome (https://bugs.python.org/issue13322) so we may need to:
 - experiment with `read1` instead of `read`
 - experiment with row (non-buffered) stream (`RawIOBase`) <-- I have hopes that this might have a positive performance impact but I might be wrong
 - use another async communication implementation

so the priority would be:
1) a test for large message on this PR
2) maybe a test handling buffered IO delays as addressed by the issue in this PR, but may need rewriting if we decide against using buffered IO in the future

bollwyvl:
Ha, my favorite stress test is `astropy.units.<TAB>` https://docs.astropy.org/en/stable/units/#module-astropy.units

python_packages/jupyter_lsp/jupyter_lsp/stdio.py
dpriedel: I don't understand your comment


python
Python 3.9.1 (default, Dec 13 2020, 11:55:53)
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> counter = 10
>>> while counter > 0:
...  print(counter)
...  counter -= 1
...
10
9
8
7
6
5
4
3
2
1
>>> print(counter)
0
>>>

krassowski: Yes, `x -= 1` is the right way to do this, but you seem to have accidentally used `--x` in your commit (which is a common muscle memory for people who program in Python and another C-like language)

dpriedel:
gah !!


sorry about that

dpriedel:
> Aside from the text thing above (please try `python -m scripts.lint from the repo`, see what it says),
> 


I tried to run that script but it seems to require several modules (robot...) that I don't have on my system.

dpriedel:
> Aside from the text thing above (please try `python -m scripts.lint from the repo`, see what it says)

I think the code change is clean now -- no more flake8 warnings.

I got the lint script to run on my system. Attached is the output showing zero errors on my system

I hope this is useful.

[lsp.log](https://github.com/krassowski/jupyterlab-lsp/files/5875541/lsp.log)

krassowski:
I got a nice reproduction on Windows run:

```
>       assert result == message * repeats
E       AssertionError: assert 'abababababababababababababababababababababababababababab' == 'ababababababababababababababababababababababababababababababababababababababababababababab'
E         - ababababababababababababababababababababababababababababababababababababababababababababab
E         ?                                                         ----------------------------------
E         + abababababababababababababababababababababababababababab

python_packages\jupyter_lsp\jupyter_lsp\tests\test_stdio.py:81: AssertionError
------------------------------ Captured log call ------------------------------
WARNING  traitlets:stdio.py:141 Readout and content-length mismatch: 58 vs 90
```

The problem is that it still happens. Another problem is that the tests have timeout issues.

python_packages/jupyter_lsp/jupyter_lsp/stdio.py
```diff
@@ -61,7 +61,7 @@ class LspStdIoReader(LspStdIoBase):
 
     @default("max_wait")
     def _default_max_wait(self):
-        return 2.0 if os.name == "nt" else self.min_wait
+        return 0.1 if os.name == "nt" else self.min_wait * 2
```
krassowski, line 64: Ok, so this is ridiculous right now (on purpose). @bollwyvl do you remember why Windows was set to `2.0`? I think this is not needed anymore. I think that we could just go with `0.2` or rather `self.min_wait * 4`?

krassowski:
Thank you @dpriedel! After adding tests and disabling buffering all now runs smoothly across all the operating systems.

PR #497, opened 2021-01-31
Release 3.3.0 and 1.1.2

krassowski:
Release time :tada:

Also enabling CI builds on 2.x.

### DIFF
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
+      - 2.x
   pull_request:
     branches:
       - master
+      - 2.x
   workflow_dispatch:
 
 defaults:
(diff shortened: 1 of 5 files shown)


bollwyvl:
Probably want to delete this branch before cutting the release...

krassowski:
Is it the tag with the same name as the branch that causes problem?

krassowski:
I see that this might be not convenient: https://softwareengineering.stackexchange.com/questions/230438/in-git-is-it-a-bad-idea-to-create-a-tag-with-the-same-name-as-a-deleted-branch. Is there a release branch naming scheme that you would prefer to avoid clashes?

bollwyvl:
Yeah, to git, it's all just `refs`... having almost-the-same-named branches and tags can lead to ... problems. I think we're well and down the _tags start with `v`_ line at this point... no need to change anything...

bollwyvl:
https://github.com/conda-forge/jupyter-lsp-feedstock/pull/15

PR #445, opened 2020-12-31
[wip] Fix robot PageConfig parsing, upgrading isort, pin jedi

bollwyvl:
Sorry this kinda ended up as a mess (should have been a one-character PR, ha), but i had to rebuild my env, and some things were going a bit wacky. This has me running all the way to failing some acceptance tests for other reasons, but figured it was worth getting out, if only for cherry-picking the good parts.

## References

- fixes #444

## Code changes

- fix robot parsing of JSON which might contain single `\` escapes
- upgrade isort
- pin jedi
- get isort and and all the fixings from `conda-forge` only with `nodefaults` instead of `defaults`

## User-facing changes

n/a

## Backwards-incompatible changes

n/a

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

PR #502, opened 2021-02-02
[wip] Unpin jedi

bollwyvl:
ipython 7.20 just landed (as did seleniumlibrary 5), let's see if everything's cool!

[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/unpin-jedi?urlpath=lab%2Ftree%2Fexamples%2FPython.ipynb)

## References

- reverts pin from #445

## Code changes

n/a

## User-facing changes

- removes `jupyter-offlinenotebook` from binder so it doesn't pop the build warning

n/a

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


bollwyvl:
Graah, as with https://github.com/jupyter/jupyter_client/pull/615, we want to unpin jed, _but only on py3.7+_...

binder/postBuild
```diff
@@ -18,6 +18,9 @@ python -m pip install -e . --ignore-installed --no-deps -vv
 jupyter labextension develop . --overwrite
 popd
 
+# Clean up incompatible extension
+jupyter labextension uninstall --no-build jupyter-offlinenotebook || echo "ok"
+
```
krassowski, line 23:
```suggestion

```

No longer needed as per https://github.com/jupyterhub/repo2docker/pull/996

bollwyvl:
Rad! While a useful feature, I still dislike how much space it takes up in
the toolbar, but it's one less thing to do.

krassowski:
So we have a `pip check` failing. Unpinning `jedi` solves part of the issue (with pyls), but there is still:

`autopep8 1.5.6 has requirement pycodestyle>=2.7.0, but you have pycodestyle 2.6.0.`

krassowski:
Though pyflakes 2.3 seems to be on conda-forge: https://github.com/conda-forge/pyflakes-feedstock ...

bollwyvl:
So... we're not getting very helpful feedback, but I think what we're probably seeing is getting a partial solved for a `pyls` that is so old it (wrongly) doesn't have a jedi pin. On 3.6, this is pulling in the three year old `0.11.1`:

```
python-language-server    0.11.1                   py36_0    conda-forge
```

So I don't think we can really unpin jedi until `python-ls` has a release...

krassowski:
Merging should solve #598.

PR #511, opened 2021-02-09
Fix install instructions in README

fsondej:
without ' ' the command fails with:
no matches found: python-language-server[all]

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Use of a python `virtualenv` or a conda env is also recommended.
    R ([languageserver](https://github.com/REditorSupport/languageserver)) servers:
 
    ```bash
-   pip install python-language-server[all]
+   pip install 'python-language-server[all]'
    R -e 'install.packages("languageserver")'
    ```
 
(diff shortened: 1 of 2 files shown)


krassowski:
Thank you! I wonder if this is due to a specific operating system or an older pip version - the examples on the pip website do not wrap the specifier with extras in quotes: https://packaging.python.org/tutorials/installing-packages/#installing-setuptools-extras. What OS/pip version do you use if I may ask?

If this is indeed a problem that affects many users we should also update it in the documentation:

https://github.com/krassowski/jupyterlab-lsp/blob/a823c987b4c6d14dc755a6976e8a77f1f6bac774/python_packages/jupyter_lsp/jupyter_lsp/specs/pyls.py#L16

PS. there is a soft fork of pyls in the works that improves the completion greatly and if you like the extension overall you might want to try it out - I will be happy to recieve feedback. Please see: https://github.com/krassowski/jupyterlab-lsp/issues/272#issuecomment-774806496 (and the following comment on disabling jedi for even better experience).

fsondej:
pip 20.2.4
Operating System: Manjaro Linux
Kernel Version: 5.9.16-1-MANJARO

I found this solution in https://github.com/palantir/python-language-server README which says to install it with ' '.

Thanks, I'll check this fork out!

krassowski:
Thanks! The comments [here](https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install) say that it is a good practice to wrap in single quote marks to avoid any shell expansion, so lets do it. I changed it in documentation too.

Issue #495, opened 2021-01-29
Completion improvements


krassowski:
## What are you trying to do?

I have some WIP improvements to share, opening this to track it, hopefully will be done this weekend but still have some other thing to finish up for my studies before I can do anything - writing it up quickly so I do not forget all the ideas.

1. The LSP completions already work when kernel is dead or not started, but do not work in other cases.

    - [ ] make the LSP completions work when kernel is slow to respond (add a timeout of 800ms, user-adjustable)
    - [ ] make the LSP completions work when kernel is busy (immediately or after the timeout as above, user decides)

2. Completion breaks when a notebook is renamed. This is related to recent fixes to completion after kernel restart.

    - [ ] fix connections after rename

3. `suppressInvokeIn` predates continuous hinting; I now see that we want another related option. Here is the rationale:

    - [ ] using continuous hinting, when writing function name in definition, e.g. `def my_func():` the hints show up at every keystroke if there is a match; but there are many irrelevant matches; this is not the case in some other editors which will only suggest `main` and `async` (so they offer context sensitive suggestions). Adding `"def"` to `suppressInvokeIn` suppresses auto-hinting when writing function name in Python/Julia/JavaScript (not in R though) - but this is not optimal as it should be the LSP server responsibility to decide. The problem is that the completions come from both the server and kernel and even if server will only suggest "main" and "async", the kernel may add a lot of useless noise. It would be good to have a list of tokens for which the kernel suggestions can be turned off by the user; it could by default include `def`.

4. Sorting completions

    - [ ] it would be great it we could subclass the `Completer` to change the ordering of completions items in ordering provided by LSP server; needs a PR upstream. Example: `def func():\nr<tab>` in some editors shows return first rather than `raise` as it is in most cases what the users want; the sorting information should be provided by the language server (but again - we do not currently use it).
    - [ ] If no sorting information is provided, we could consider having a type-based tie-resolution mechanism (optional, in user-defined order); e.g. having `return` and `reversed` with equal score 1 for `re<tab>` it might make sense to show the keyword `return` first rather than the function `reversed`.

PR #520, opened 2021-02-13
Sorted completions, improve completions marking and filtering and doc panel

krassowski:
- implements sorting of completions solving point 4 of #495 by taking use of `sortText` if provided (jets tested and indirectly integration-tested)
- makes use of `filterText` if provided (jest tested)
- allows user to choose whether in presence of ties (i.e. two suggestions having the same match score) the kernel suggestions should go first or last (the latter by default) (integration tested)
- ensures that a full copy of item object gets created during filtering so that the documentation can still be retrieved dynamically (adds integration test for that)
- always re-renders the docpanel so that the markdown is always rendered (while waiting for https://github.com/jupyterlab/jupyterlab/pull/9663 to be merged)
- allows to render labels that contain HTML tags (https://github.com/jupyterlab/jupyterlab/issues/9790)

## Code changes

- Use a custom completer model (`LSPCompleterModel`) and create a generic template `GenericCompleterModel` which will make the live much easier once upstreamed for 4.0 (https://github.com/jupyterlab/jupyterlab/issues/9763); at this point it is experimental API to be upstreamed so if anyone depending on our codebase looks at it - please do not rely on this API just yet.
- adds a second pass of `matchSumOfSquares` to allow for matching on `filterText` which does not have to be a substring of `label`; there is a price to pay for the second match calculation on numpy completions, but the total cost is over an order of magnitude lower of the current limiting factor of SVG rendering and item rendering in general.

## User-facing changes

TBD. So many nice things :)

- moved `LazyCompletionItem` to a file of its own (`item.ts`); better to make this change early
- added  `LSPCompleterModel`

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 ## CHANGELOG
 
+### `@krassowski/jupyterlab-lsp 3.4.0` (unreleased)
+
+- features:
+
+  - the priority of the completions from kernel can now be changed by switching new `kernelCompletionsFirst` setting ([#520])
+  - completer panel will now always render markdown documentation if available ([#520])
+    - the implementation re-renders the panel as it is the best we can do until [jupyterlab#9663](https://github.com/jupyterlab/jupyterlab/pull/9663) is merged
+  - the completer now uses `filterText` and `sortText` if available to better filter and sort completions ([#520])
+
+- bug fixes:
+  - completer documentation will now consistently show up after filtering the completion items ([#520])
+  - completions containing HTML-like syntax will be displayed properly (an upstream issue) ([#520])
+
+[#520]: https://github.com/krassowski/jupyterlab-lsp/pull/520
+
 ### `@krassowski/jupyterlab-lsp 3.3.1` (2020-02-07)
 
 - bug fixes:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -227,7 +227,7 @@ python scripts/atest.py --suite "05_Features.Completion"
 ##### Run a single test
 
 ```bash
-python scripts/atest.py --test "Works With Kernel Running"
+python scripts/atest.py --test "Works When Kernel Is Idle"
 ```
 
 ##### Run test with a tag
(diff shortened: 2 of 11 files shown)


bollwyvl:
Awesome stuff. The sorting of search results is hilariously hard, getting mired in the whole field-based boosting, etc. but this is a very good step.

bollwyvl:
It's probably worth a look through `@lumino/algorithm`, there are a number of (Big O-annotated) implementations that might be helpful, and look more like core.

krassowski:
Yes, the code related to sorting is taken from core and changes are meant to be up-streamed for 4.0. I also modified the pyls sorting on my fork, please feel welcome to chime in and comment it: https://github.com/krassowski/python-language-server/pull/1/files#diff-66c341f45ce292941d6a15c7e287ce5faf7f7e7d10f35bb5919722995bd982c9R9-R36

As pyls is in a weird state where the spyder folks who were maintaining it (even though it was maintaining in a literal sense of literal life support and not much development) no longer have merge rights (which might take months to rectify) we should think about some solution. One solution would be a community fork, but lets see how things play out.

While my fork is not yet ready for general use (no CI and infrastructure set up) and I do not like the legacy baggage from pyls (like Python 2), it might be possible to refactor it to something usable and keep it compatible with community extensions.

An alternative would be to default to jedi-langauge-server as it is being still maintained and developed; it is nice because of the pygls but not nice because of a strictly limited scope. If it had the plugins/entry points mechanism as pyls it would be so great! (See https://github.com/pappasam/jedi-language-server/issues/32).

A third alternative is using pyright. It is node-based (and MS) and controversial due to how pylance is being licensed.

Just mentioning this here, but it will probably deserve its own issue soon.

bollwyvl:
Yeah, it's hard to have nice things, great work you're doing.

This is probably _another JEP_, not just a discussion on our repo. I feel like a number of things _could_ happen:

- hard fork of pyls, i wouldn't hold my breath on the current copyright holders clearing anything up
  - despite the baggage, we know how it works, and the plugins work
  - would also require forking the plugins, or convincing them to migrate along with us
- get a robust plugin mechanism into pygls itself so that jedi-language-server wouldn't even know it was being extended :imp: 
  - the pygls pattern is a little weird, but actually pretty good for this
  - the pluggy route that pytest, etc. uses would actually be superb for this
- maybe with the above, create and maintain Yet Another Python Language Server
  - but demonstrate solid integration with the python kernels (and their derivative wrapper kernels), notebooks, etc.

I don't really see myself wanting to rely on any of the first-party MS stuff (especially if it's future-abandonware, and my feelings on requiring `node` to do python work are likely clear at this point), if anything they'll keep building their walled garden to get a slice of that sweet, sweet data science/ML mindshare and money.

PR #538, opened 2021-02-20
add extra conf.d config reader for servers, fix conf.d names

### LINKED ISSUE #535: Server detection configuration files are not passed to us from config directories
Placing LSP server config in jupyter_notebook/server_config.d does not seem to work. See #534 for details; using `jupyter_server_config.json` still works. The server apps detects the json files:

> [ServerApp] Paths used for configuration of jupyter_server_config:
>   etc/jupyter/jupyter_server_config.d/pyright.json

in several passes. If this is an upstream issue, it needs to be looked into; if this is something at our end, we need to fix/update the docs.

bollwyvl:
## References

- fixes #535

## Code changes

- [x] `LanguageServerManager.conf_d_language_servers` 
  - [x] adds an (unconfigurable) trait which re-uses the `ConfigManager` to look through all the `.d` folders to find `language_servers`
    - [ ] it might be "too late" to do much else (e.g. turn off autodetection, add node roots)... will have to consider how much we want to be configured in more places
  - [x] tests loading from all `jupyter*config.d` paths in `jupyter_config_path` (by env var)
- [x] fixes up name of files sent to `.d` folders for _our_ config as a `jpserverextension` or `nbserverextension`
- [x] adds CI for `serverextension` (a la `notebook`) being enabled
  - might not be _strictly_ necessary as binder has updated...

## User-facing changes

n/a

n/a

(though due to the _lack_ of magic from `data_files`, folks might have some crusty files laying around if they upgrade)

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


bollwyvl:
Seems like the fails are :snowflake: ...

krassowski (approved):

PR #539, opened 2021-02-21
Improve language servers configuration documentation

krassowski:
- [x] improve wording of the language server autodetection and on how to add a new one
- [x] fix some broken links
- [x] reference the scala instructions below the language servers listing
- [x] explain in more detail how configuration files approach differs from specs approach, and explain how to contribute using existing helper classes

## Chores

- [x] linted
- [ ] tested
- [x] documented
- [ ] changelog entry


CONTRIBUTING.md
```diff
@@ -322,44 +322,56 @@ python scripts/lint.py
...
+A spec is a Python callable (a function, or a class with `__call__` method) that accepts a single argument, the
+`LanguageServerManager` instance, and returns a dictionary of the form:
 
 ```python
 {
   "python-language-server": {            # the name of the implementation
-      "version": 1,                      # the version of the spec schema
+      "version":  SPEC_VERSION,          # the version of the spec schema (an integer)
       "argv": ["python", "-m", "pyls"],  # a list of command line arguments
       "languages": ["python"]            # a list of languages it supports
```
bollwyvl, line 346: we should add mime types, since they are _basically_ required for notebook support

CONTRIBUTING.md
```diff
@@ -322,44 +322,56 @@ python scripts/lint.py
...
-A spec is a python function that accepts a single argument, the
-`LanguageServerManager`, and returns a dictionary of the form:
+A spec is a Python callable (a function, or a class with `__call__` method) that accepts a single argument, the
+`LanguageServerManager` instance, and returns a dictionary of the form:
 
 ```python
 {
   "python-language-server": {            # the name of the implementation
-      "version": 1,                      # the version of the spec schema
+      "version":  SPEC_VERSION,          # the version of the spec schema (an integer)
```
bollwyvl, line 344: maybe a link to the schema _folder_ on github (since we might change the name of the _file_ in the future).

docs/Configuring.ipynb
```diff
@@ -271,8 +270,8 @@
     "```\n",
     "\n",
     "You are good to go now! Just start `jupyter lab` and create a notebook with\n",
-    "either the Spark or the Scala kernel and you should be able to see the metals\n",
-    "server initialised from the bottom left corner."
+    "either the Spark or the Scala kernel and the metals server should automatically\n",
```
bollwyvl, line 273: looking at the metals thinger... it lists the coursier instructions twice... sorry, still no progress on making that conda installable... there are lots of... opinions on the right way to do so.

docs/_static/css/custom.css
```diff
@@ -32,7 +32,9 @@ body div.nboutput.container div[class*='highlight'] pre {
 }
 
 .wy-nav-content {
-  max-width: 100%;
+  /* Increase the width of the content from the default 800px;
+  not using 100% as it makes reading sentences unbearable on 4K screens */
+  max-width: 900px;
```
bollwyvl, line 37: maybe we make `p` have a max width? the images and tables play a lot better when wide...

krassowski: That's an idea.. Maybe content max-width 1200px and paragraphs max-width 900px?

Also, maybe we just make the images clickable for easier detail viewing? OR maybe the two huge images that we have in https://jupyterlab-lsp--539.org.readthedocs.build/en/539/Architecture.html are just too huge anyways and we should work on a better layout?

CONTRIBUTING.md
krassowski: I was thinking about some CI dead link checker.

bollwyvl: `pytest-check-links` works well, indeed. I usually only run it against the generated HTML, since I don't really care what hoops you have to jump through on the source side to get to solid HTML.

Heck, we can even check spelling in HTML with `hunspell`... i usually test against `en-GB` and `en-US`, and the (inevitable) custom dictionary actually becomes a useful artifact in its own right.

krassowski: Interestingly we already have `pytest-check-links` in docs script. Not sure if we would get notified of any issue on CI ever though...

krassowski: Also, would it pick up the issue of missing anchor (`page.ipynb#section`) target?

PR #541, opened 2021-02-21
Release jupyter-lsp v1.1.4

krassowski:
Note: no frontend release this week.


bollwyvl:
Started:
https://github.com/conda-forge/jupyter-lsp-feedstock/pull/20

We should have _some_ kinda tag to denote that 4e6d164 was a release...

PR #544, opened 2021-02-23
Handle case where getCursor() returns null in highlighter

### LINKED ISSUE #543: Error thrown when creating a new cell with shift + enter
## Description

When creating a new cell by pressing shift+enter, an exception is thrown and printed to the JS console.

## Reproduce

1. Open a new notebook
2. Select the first blank cell and press shift+enter
3. A warning will be printed to the dev console
    ```
    codemirror_editor.js:185 LSP.WidgetAdapter: Editor not found in editor_to_source_line map
    ```
4. An un-caught exception is printed to the dev console
   ```
   TypeError: Cannot read property 'line' of null
    at VirtualDocument.get_editor_at_source_line (document.js:491)
    at Proxy.get_editor_at_root_line (codemirror_editor.js:276)
    at Proxy.getTokenAt (codemirror_editor.js:283)
    at Proxy.get_token_at (codemirror_editor.js:124)
    at HighlightsCM.onCursorActivity (highlights.js:87)
   ```
   
   There are other ways to reproduce this issue, they seem to happen when the focus changes. If there is no current editor available, then the cursor position is not valid.

## Expected behavior

   The first warning `Editor not found in editor_to_source_line map` should be printed, but no exception should be thrown.

## Context

This issue does not seem to be environment-dependent. I've tested on OSX and Linux, with Chrome and Chromium-based browsers. I've tested with jupyter-lsp from master, the latest stable release (v3.4.1) as well as v2.1.3.

brendan0powers:
## References

Fixes #543 

## Code changes

onCursorActivity currently assumes that getCursor() will throw an exception if it was unable to obtain a valid cursor position. This is not the case. There are a few instances where it will return either undefined or null. In particular, if there is not yet an editor available. This change checks root_position and ensures it has a valid value.

## User-facing changes

none

none

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/features/highlights.ts
+++ b/packages/jupyterlab-lsp/src/features/highlights.ts
@@ -145,6 +145,11 @@ export class HighlightsCM extends CodeMirrorIntegration {
       return;
     }
 
+    if (!root_position) {
+      this.console.warn('no root position available');
+      return;
+    }
+
     const token = this.virtual_editor.get_token_at(root_position);
 
     // if token has not changed, no need to update highlight, unless it is an empty token
--- a/packages/jupyterlab-lsp/src/features/highlights.ts
+++ b/packages/jupyterlab-lsp/src/features/highlights.ts
@@ -145,7 +145,7 @@ export class HighlightsCM extends CodeMirrorIntegration {
       return;
     }
 
-    if (!root_position) {
+    if (root_position == null) {
       this.console.warn('no root position available');
       return;
     }


packages/jupyterlab-lsp/src/features/highlights.ts
```diff
@@ -145,6 +145,11 @@ export class HighlightsCM extends CodeMirrorIntegration {
       return;
     }
 
+    if (!root_position) {
```
krassowski, line 148: Would the following work as well for you?

```suggestion
    if (root_position == null) {
```

If yes, then I would slightly prefer it for consistency with other places in codebase.

brendan0powers: That works for this particular issue, but getCursor may return undefined as well. For example: https://github.com/krassowski/jupyterlab-lsp/blob/d86925f4a1c51c5d0fe76c1d2be74b49ffb71927/packages/jupyterlab-lsp/src/virtual/codemirror_editor.ts#L66

krassowski: The magic of null is that it `null == undefined` (not that I propose == rather than ===)

krassowski: To elaborate `!position` could bite in the future, should we decide to represent positions as integer offsets (I do not really believe that it would happen, but it is used by some editors like that and it might happen).

krassowski: and then `!0` would give `true`

brendan0powers: Oh, yea, missed the == vs ===. That change seems fine.

brendan0powers: Thanks for the clarification!

krassowski (approved):

brendan0powers:
Would it be possible to back-port this to the 2.x series? I'm happy to help with the backport.

jtpio:
> Would it be possible to back-port this to the 2.x series? I'm happy to help with the backport.

Just opened https://github.com/krassowski/jupyterlab-lsp/pull/546 to backport the change.

bsyouness:
This seems to still be happening in other spots see #658.

Issue #236, opened 2020-03-30
Compatibility with Multiple Completer Extensions


edzkite:
## Elevator Pitch
Kite is planning on creating an extension for JupyterLab, as a fork of this extension. Ideally, if users have both this extension and Kite's extension installed, things should just work.

## Motivation
Based off my understanding of JupyterLab's `completer-extension` [package](https://github.com/jupyterlab/jupyterlab/blob/master/packages/completer-extension/src/index.ts) and JupyterLab-LSP's implementation, it's only possible to have one connector per notebook or file.

Creating this issue to understand if:
1. That assessment is accurate
2. If so, what it'd take to support multiple connectors, e.g. LSPConnector + KiteLSPConnector across extensions
3. Alternates to achieve said goal.

bollwyvl:
> fork of this extension. 

I for one am happy to find ways to add more composite behavior to this repo: we're just two maintainers and a handful of contributors, and we're still hustling to make what we have work predictably. On #76, we've laid out (and have slowly been moving towards) a repo structure and workflow to allow for this, and #184 points towards supporting more kinds of sources. I also certainly feel kernels themselves need to get in the game, and be able to interact with (or act in place of) language servers. 

> it's only possible to have one connector per notebook or file.

When these show up, it usually indicates a need to:
- carve out the functionality into something standalone (as an extension)
- make it look as much like core lab as possible
- propose it to lab core, ideally in a non-breaking way
- get it merged and released
- deprecate the standalone extension

So someplace, there needs to be a `MultiCompletionConnector` to which multiple completion sources get registered... a bit tired to dig further, but it seems like getting a `composite` added somewhere in the [`CompletionManager`](https://github.com/jupyterlab/jupyterlab/blob/7fb018558f9c812e3e2a3355fb6ab7c1a30486d8/packages/completer/src/tokens.ts#L18) spec would make the most sense, followed by some additions to the [extension](https://github.com/jupyterlab/jupyterlab/blob/7fb018558f9c812e3e2a3355fb6ab7c1a30486d8/packages/completer-extension/src/index.ts#L89). Then some kind of connector voodoo, and some rules around how quickly they come back: i'd wager if we get client-side stuff in, it's going to come back _real fast_, and it would be a shame to wait for a websocket roundtrip to a server process (or third party call) to slow _all_ of that down. So maybe there's a model missing, too.

krassowski:
[...]

PR #549, opened 2021-02-27
Source-aware completions and more completer settings

### LINKED ISSUE #492: Use completion item detail in completion renderer
Completion item provides `detail` property which includes language-specific detail information about the completion, e.g. the module from which a function was imported.

1) It should be displayed by default instead of type text if available (the type is already shown by the icon); We can modify the `<li>` element with the completion item after its creation, or modify how it is created here:

    https://github.com/krassowski/jupyterlab-lsp/blob/d8b670528605f4f736da08c965c42b82688d0981/packages/jupyterlab-lsp/src/features/completion/completion_handler.ts#L137

2) the user should be able to choose what they want to be displayed (maybe there options: detail, type, whichever available); this can happen later, but the access to settings in the renderer is already possible using the `this.options.integrator.settings`:

    https://github.com/krassowski/jupyterlab-lsp/blob/d8b670528605f4f736da08c965c42b82688d0981/packages/jupyterlab-lsp/src/features/completion/renderer.ts#L78

    to add a new setting one should modify the [schema/completion.json](https://github.com/krassowski/jupyterlab-lsp/blob/master/packages/jupyterlab-lsp/schema/completion.json) file and follow with `jlpm build:schema` from the repository root.

3) the resolve request needs to be updated to copy the detail from response, as is the documentation here:

    https://github.com/krassowski/jupyterlab-lsp/blob/d8b670528605f4f736da08c965c42b82688d0981/packages/jupyterlab-lsp/src/features/completion/renderer.ts#L25

krassowski:
## References

Small step towards #495 and #236. Fixes #492.

## Code changes

- [x] trim spaces in insertText when merging completions (was the source of duplicated completions as Python kernel was returning 'return' and 'return ')
- [x] generalised completion merging code so that plugin in kite or another will be easier in the future
- [x] extend `CompletionHandler.ICompletionItemsReply` to include source information (`ICompletionsReply`)
- [x] define `ICompletionsSource` for source metadata storage
- [x] extend `CompletionHandler.ICompletionItem` to include source info and others (`IExtendedCompletionItem`)
- [ ] source in `IExtendedCompletionItem` and `ICompletionsReply` can be null/undefined at different stages. How to handle this safely for future us?

## User-facing changes

- [x] detail is now shown instead of the type by default; if no detail is available, then type or source ("Kernel" for kernel completions) is shown instead.

detail-and-source-opt

- [x] exposed settings:
  - [x] `caseSensitive`: old behaviour retained by default
  - [x] `includePerfectMatches`: old behaviour retained by default
  - [x] `labelExtra`: new behaviour - display detail from LSP rather than the type in the completion label

new-completion-settings

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - features:
 
   - adds `%%bigquery` IPython cell magic support for BigQuery ([#553], thanks @julioyildo)
+  - completions filtering can be set to case-insensitive in settings ([#549])
+  - completions filtering can hide exact matches ([#549])
+  - the extra information displayed next to the completion label now can include 'detail' (usually module/package of origin), and can be customized in settings ([#549])
 
 - bug fixes:
 
@@ -18,6 +21,7 @@
 
 [#544]: https://github.com/krassowski/jupyterlab-lsp/pull/544
 [#547]: https://github.com/krassowski/jupyterlab-lsp/pull/547
+[#549]: https://github.com/krassowski/jupyterlab-lsp/pull/549
 [#553]: https://github.com/krassowski/jupyterlab-lsp/pull/553
 [#560]: https://github.com/krassowski/jupyterlab-lsp/pull/560
 [#562]: https://github.com/krassowski/jupyterlab-lsp/pull/562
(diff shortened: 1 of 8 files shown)


krassowski:
I am not 100% sure if the new behaviour with detail should be enabled as default. It seems to work fine, but there can be this awkward delay visible on the GIF when `ggplot2` and `geom_` functions are being typed, when the completion items are being resolved (i.e. the detail gets fetched from the server).

PR #557, opened 2021-03-18
localize strings for jupyterlab-lsp

JessicaBarh:
## Code changes

any text displayed on the UI is made localizable using the TranslationBundle and ITranslator from @jupyterlab/translation

## User-facing changes
User will be able to use jupyterlab-lsp extension in prefered language given that the language pack is installed.

Can also be viewed by using the language settings to enable string prefix for all localized strings:
image

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/components/statusbar.tsx
+++ b/packages/jupyterlab-lsp/src/components/statusbar.tsx
@@ -109,9 +109,8 @@ class CollapsibleList extends React.Component<
 }
 
 class LSPPopup extends VDomRenderer<LSPStatus.Model> {
-  constructor(model: LSPStatus.Model, trans: TranslationBundle) {
+  constructor(model: LSPStatus.Model) {
     super(model);
-    this.trans = trans;
     this.addClass('lsp-popover');
   }
   render() {
@@ -140,15 +139,15 @@ class LSPPopup extends VDomRenderer<LSPStatus.Model> {
 
           let status = '';
           if (connection?.isInitialized) {
-            status = this.trans.__('initialized');
+            status = this.model.trans.__('initialized');
           } else if (connection?.isConnected) {
-            status = this.trans.__('connected');
+            status = this.model.trans.__('connected');
           } else {
-            status = this.trans.__('not connected');
+            status = this.model.trans.__('not connected');
           }
 
           const icon =
-            status === this.trans.__('initialized')
+            status === this.model.trans.__('initialized')
               ? circleIcon
               : circleEmptyIcon;
 
@@ -197,13 +196,13 @@ class LSPPopup extends VDomRenderer<LSPStatus.Model> {
       <div className={'lsp-popover-content'}>
         <div className={'lsp-servers-menu'}>
           <h3 className={'lsp-servers-title'}>
-            {this.trans.__('LSP servers')}
+            {this.model.trans.__('LSP servers')}
           </h3>
           <div className={'lsp-servers-lists'}>
             {servers_available.length ? (
               <CollapsibleList
                 key={'available'}
-                title={this.trans.__('Available')}
+                title={this.model.trans.__('Available')}
                 list={servers_available}
                 startCollapsed={true}
               />
@@ -213,7 +212,7 @@ class LSPPopup extends VDomRenderer<LSPStatus.Model> {
             {running_servers.length ? (
               <CollapsibleList
                 key={'running'}
-                title={this.trans.__('Running')}
+                title={this.model.trans.__('Running')}
                 list={running_servers}
               />
             ) : (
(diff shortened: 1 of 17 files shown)


packages/jupyterlab-lsp/src/features/rename.ts
```diff
@@ -35,6 +40,7 @@ const COMMANDS: IFeatureCommand[] = [
       features
     }) => {
       const rename_feature = features.get(FEATURE_ID) as RenameCM;
+      rename_feature.trans = trans;
```
krassowski, line 43: Maybe it would be better to add `trans` to `IEditorIntegrationOptions` which are passed to `IFeatureEditorIntegration` (and its implementation in `CodeMirrorIntegration`) in:

https://github.com/krassowski/jupyterlab-lsp/blob/e9b115da75d36f3efd43872b0bbfbf10d610c20a/packages/jupyterlab-lsp/src/adapters/adapter.ts#L574-L582

krassowski (commented):
Fantastic work, thank you for starting this effort!

I saw the test are failing for rename, and had a look into how this could be potentially improved - please see my suggestion in the comment.

Please let us know if we can help in any way. Do not worry about tests failing on `pip check` - this is a known issue that I am going to address soon.

JessicaBarh:
Thanks @krassowski! I added your suggestion and looks to be working. Do you have any ideas for the diagnostics.spec.ts test errors?

packages/jupyterlab-lsp/src/features/diagnostics/diagnostics.ts
```diff
@@ -45,6 +46,8 @@ function escapeRegExp(string: string) {
   return string.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&');
 }
 
+let trans: TranslationBundle;
```
krassowski, line 49: Maybe move it to the `consturctor()` of `DiagnosticsPanel` (which needs to be created) and initialize with `nullTranslator`, like:

```js
export const diagnostics_panel = new DiagnosticsPanel(nullTranslator);
```

krassowski:
I gave it a go on [binder](https://mybinder.org/v2/gh/JessicaBarh/jupyterlab-lsp/add-i18n?urlpath=lab) and it is looking good. The only place where I found an opportunity for improvement was in the Severity column of diagnostics panel:

Screenshot from 2021-03-23 21-24-07

which is implemented here:

https://github.com/krassowski/jupyterlab-lsp/blob/f324c69db3685dbf3fac92d097cda2c8f0b57388/packages/jupyterlab-lsp/src/features/diagnostics/listing.tsx#L171

JessicaBarh:
> I gave it a go on [binder](https://mybinder.org/v2/gh/JessicaBarh/jupyterlab-lsp/add-i18n?urlpath=lab) and it is looking good. The only place where I found an opportunity for improvement was in the Severity column of diagnostics panel:
> 
> Screenshot from 2021-03-23 21-24-07
> 
> which is implemented here:
> 
> https://github.com/krassowski/jupyterlab-lsp/blob/f324c69db3685dbf3fac92d097cda2c8f0b57388/packages/jupyterlab-lsp/src/features/diagnostics/listing.tsx#L171

Perfect, thanks for finding that @krassowski. I've just added that as well as the 'panel columns' menu that was also missing.

packages/jupyterlab-lsp/src/components/statusbar.tsx
```diff
@@ -138,14 +139,17 @@ class LSPPopup extends VDomRenderer<LSPStatus.Model> {
...
           } else {
-            status = 'not connected';
+            status = this.model.trans.__('not connected');
           }
 
-          const icon = status === 'initialized' ? circleIcon : circleEmptyIcon;
+          const icon =
+            status === this.model.trans.__('initialized')
+              ? circleIcon
+              : circleEmptyIcon;
```
krassowski, line 152: I wonder if it would be better to delay the translation until the line 161 (was 157):

```
              <span className={'lsp-document-status'}>
                {status}
```

bollwyvl:
Apropos of nothing (aside from a separate pr): the jupyter_server api is
changing for translations. We should take a look at any server-built
strings that aren't data (e.g. specs) and see if we want to add _i18n
anywhere.

One thing I added over on the crazy kernel branch (which I swear I'll warm
up soon, because ls-ws) was a "jupyter lsp troubleshoot" CLI that could do
some troubleshooting... It had some non-generated some strings.

krassowski:
Thank you @JessicaBarh!

JessicaBarh:
Sorry @krassowski haven't been able to get back on that the past week, thanks for finishing that!

PR #493, opened 2021-01-24
[wip] Add Robot Framework language server

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/add-robot-server?urlpath=lab%2Ftree%2Fatest%2Fexamples%2Fexample.robot)

## References

- fixes #332 with `robotframework-lsp`, which is generally available on `pypi` and `conda-forge`
- also adding
  - `jupyterlab_robotmode` for robot syntax highlighting
  - `robotkernel` as it has a number of nice bells and whistles for browser-based testing
    - `xeus-robot` is _also_ very cool, as it works with the debugger, which we need to start considering
      - but I think we want a separate PR that adds all the `xeus` stuff in one go, as they are most likely to conflict with each other
        - e.g. `xeus-python`, `xeus-sql`, `xeus-robot`, etc.

## Code changes

- [x] add `robotframework-lsp` in a number of places (_not_ lint, though!)
  - [ ] if we find ourselves wanting to use robot notebooks, we can figure out an `nblint` plugin or something 
- [x] the robotframework 4.0 release is coming up... there might be some version adventures between releases, have pinned for the time being to avoid surprises.
- [ ] validate that [spec](https://github.com/robocorp/robotframework-lsp/blob/master/robotframework-ls/src/robotframework_ls/ext/jupyter_lsp.py) that `robotframework-lsp` ships :tada:  works  
  - [ ] we're better off upstreaming a PR than claiming support for one with tweaks
- [x] add editor robot tests
- [ ] add notebook robot tests

## User-facing changes

- Users should see nicer integration with robot files and notebooks
  - will demo on our Binder for looking at/running _our_ tests, which is cute
    - browser screenshots _can_ be made to work
      - might need to add `apt.txt` (or stuff it, and use a `Dockerfile`)... binder's getting a bit unwieldy

## Backwards-incompatible changes

- N/A

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry

PR #558, opened 2021-03-20
CI updates: action updates, caching, some pins

bollwyvl:
## References

- near-term alternative to #502

## Code changes

- pin some more pyls-jedi-ipython things
  - these can be relaxed on py3.7+ once `python-ls` lets us upgrade to jedi 0.18
  - maybe just drop 3.6 at that point? not supported by numpy, ipython, robot-ls #493
- defer installing node/julia language servers until more "cheap" tests are done
  - extension checking, etc
- cache more things
  -  no numbers yet, but should cut a couple minutes off windows
- upgrade some actions
  - it's kinda icky being on `setup-miniconda@master`, but i personally vouch for it :grimacing: 
  - `cache@v2` should _actually_ make the cache work on windows because gnu tar

## User-facing changes

- n/a

- n/a

## Chores

- [x] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


bollwyvl:
Aside from my two whitespace lint issues (gah), https://github.com/krassowski/jupyterlab-lsp/runs/2153513491?check_suite_focus=true is coming up :heavy_check_mark:

krassowski:
Thank you! Ideally we would unpin, but I agree it is a mess now and we need a new pyls release.

> it's kinda icky being on setup-miniconda@master, but i personally vouch for it grimacing

What about using a hash `setup-miniconda@77b16ed746da28724c61e1f1ad23395a4b695ef5`?

bollwyvl:
✔️ looks good on master. Maybe I'll just... Do the release on
setup-miniconda 😂

Issue #551, opened 2021-03-03
Add support for graph query languages


bollwyvl:
## Elevator Pitch

Add language server support for key graph query and data languages, such as SPARQL, Turtle, and GraphQL

## Motivation

The most prominent language servers in this space just got a refresh, and it's probably worth trying them again.

## Design Ideas

- Add specs for the [stardog language server](https://github.com/stardog-union/stardog-language-servers). Probably just the open standards (with or without quotes) mentioned above, rather than the stardog-specific languages.

- add at least ipython magic/syntax support for the unambiguous languages: `%%sparql`, `%%graphql`, `%%ttl` (but probably not `turtle`, which we don't want to [trample](https://en.wikipedia.org/wiki/Trampled_by_Turtles)).

PR #560, opened 2021-03-20
Fix regular-expression based extractors offset

### LINKED ISSUE #559: Regexp foreign extractors do not return proper ranges if extracted value is in prefix of matched pattern
## Description

Investigating #555, I found a related but distinct issue in regular expression-based foreign extractors.

Screenshot from 2021-03-20 12-50-45

The naive way of computing the index for the capture group is currently using `indexOf`, which finds `h` in %%pyt**h**on, rather than the second `h` in next line. A better way to get index for capture group which is used to extract the code is needed.

The wrong index leads to wrong offset and the foreign document not being correctly aligned and not detected at proper cursor position.

## Reproduce

1. Add magic cell with `%%python\nh`
2. Try to tab-complete after `h`
3. See completions for all symbols, not those matching `h`

## Expected behavior

Only relevant completions are shown.

## Context

- JupyterLab version: 3.0.9
- JupyterLab-LSP version: 3.4.1

krassowski:
## References

Fixes #559

## Code changes

- splits `RegExpForeignCodeExtractor`'s `extract_to_foreign` into `foreign_capture_groups` and `foreign_replacer` (in a backward compatible way, with a deprecation notice for 4.0)
- corrects computation of offset (and thus ranges) for extracted documents by relying on `foreign_capture_groups` to determine start index; the old way was ok-ish for some extractors, but not a good general solution

## User-facing changes

Completion and other features will work more consistently in cell magics and line magics (and other extractors).

None, but deprecation notices added for `RegExpForeignCodeExtractor.extract_to_foreign` and `RegExpForeignCodeExtractor.keep_in_host`.

## Chores

- [x] linted
- [x] tested
- [x] documented
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,12 @@
 
   - prevents throwing a highlights error when adding new cell with <kbd>Shift</kbd> + <kbd>Enter</kbd> ([#544])
   - fixes IPython `pinfo` and `pinfo2` (`?` and `??`) for identifiers containing `s` ([#547])
+  - fixes incorrect behaviour of LSP features in some IPython magics with single line of content ([#560])
 
 [#544]: https://github.com/krassowski/jupyterlab-lsp/pull/544
 [#547]: https://github.com/krassowski/jupyterlab-lsp/pull/547
 [#553]: https://github.com/krassowski/jupyterlab-lsp/pull/553
+[#560]: https://github.com/krassowski/jupyterlab-lsp/pull/560
 
 ### `jupyter-lsp 1.1.4` (2020-02-21)
 
--- a/atest/05_Features/Completion.robot
+++ b/atest/05_Features/Completion.robot
@@ -296,6 +296,28 @@ Shows Documentation With CompletionItem Resolve
     Completer Should Include Documentation    the default method of the
     [Teardown]    Clean Up After Working With File    completion.R
 
+Shows Only Relevant Suggestions In Known Magics
+    # https://github.com/krassowski/jupyterlab-lsp/issues/559
+    # h<tab>
+    Enter Cell Editor    20    line=2
+    Trigger Completer
+    Completer Should Suggest    help
+    Completer Should Not Suggest  from
+    Completer Should Suggest    hash
+
+Completes In R Magics
+    # Proper completion in R magics needs to be tested as:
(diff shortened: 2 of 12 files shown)


bollwyvl:
So we were trying our hand at an out-of-tree extension (locally) for #551... when the full built application is set up, the `RegExpForeignCodeExtractor` and a few of the interfaces would need to be hoisted all the way up to `jupyterlab-lsp/index.ts`. For reference, at present we're basically just exporting `plugins`. This ends up making us ship (and use) our own copy of jlsp by referencing stuff in `lib/extractors/regexp`, which isn't too fun.

I feel like trying it out for a very string-in-string-out kind API (before we get to rich widgets, etc.) would be an interesting experiment. Will make a separate ticket!

krassowski:
Wait, does it mean that I can freely break the API because it was never actually public? Is this possible to like `@ts-ignore` lack of the export, or is it rather unlikely that anyone used it?

krassowski:
I guess that one could have used `/lib` is this correct?

krassowski:
I came to a conclusion that forcing a single group is not a good idea, and ideally if having multiple adjacent groups in output is good enough, we should allow to do this without the need for writing a javascript `replacer` function. In particular this would allow for:
- better handling of intertwined foreign-host fragments using the indices of the capture groups (in the future), and
- specifying more of the regexp-based extractors as pure JSON

I ended up updating `foreign_capture_group: number` to `foreign_capture_groups: number[]`

bollwyvl:
> I can freely break the API because it was never actually public

No, not really, I'd say, but hoisting it all the way to `index.ts` is kind of a _statement_.

> I came to a conclusion that forcing a single group is not a good idea

So in unrelated stuff, there _is_ a JSON spec for some related stuff... generally under the term problem matchers. I lost the good spec document (will find) but here's a python one:

https://github.com/actions/setup-python/blob/main/.github/python.json

The good parts:
- it's JSON
- there are many examples in the wild
 
The bad parts:
- github actions and vscode have a _slightly_ different interface definition (always) and it's not machine readable

The future
- it would be rad to have those to be able to jump-to-definition in e.g. outputs
- the pattern seems expandable to _all kinds of things_, including these transclusions
- i'd _love_ to be able to have a `load_ext` be able to create an output that changed how the client worked

PR #562, opened 2021-03-20
Propose Extractor API (#561, Plan A)

### LINKED ISSUE #561: Expose extractor API (as separate package?)
## What are you trying to do?

> from https://github.com/krassowski/jupyterlab-lsp/pull/560#issuecomment-803457273
>>
>> So we were trying our hand at an out-of-tree %%sparql extractor (locally) for #551... when the full built application is set up, the RegExpForeignCodeExtractor and a few of the interfaces would need to be hoisted all the way up to jupyterlab-lsp/index.ts. 

#### Plan A**

- in `packages/jupyterlab-lsp/tokens.ts`, as described in original issue, just hoist a few classes, e.g. 
  - `export { IForeignCodeExtractor } from './extractors/types'`  
  - `export { RegExpForeignCodeExtractor } from './extractors/regexp'`

#### Plan B

- extract  these tokens (and their parent modules) into a new package
- in `packages/transclusions/index.ts` and elsewhere, import these tokens

## How is it done today, and what are the limits of current practice?

>> For reference, at present we're basically just exporting plugins. This ends up making us ship (and use) our own copy of jlsp by referencing stuff in lib/extractors/regexp, which isn't too fun.

## What is new in your approach and why do you think it will be successful?

>> I feel like trying it out for a very string-in-string-out kind API (before we get to rich widgets, etc.) would be an interesting experiment. Will make a separate ticket!

- originally in #76, we had long discussed publishing additional portions of the underlying architecture 

## Who cares? If you are successful, what difference will it make?

- extension authors will easily be able to create simple, potentially single `.ts` file plugins to add new transclusion rules
- Jupyterlab-lsp maintainers might not have to review PRs for every special magic

## What are the risks?

- if shipped as part of the public API of `jupyterlab-lsp`, we'd have to be pretty sure to get it _right_ the first time
- if made as a separate package, we would have to handle the semver range
  - either start it at the current release version (and bump the parent version), 
    - or restart it at `0.1.0` or something
      - this _may_ make it easier to maintain in backport

## ~~How much will it cost?~~

Free as in speech :laughing: .

## How long will it take?

- **Plan A**
  - a few hours of work, get on the next release train 
- **Plan B** 
  - a fair amount of boilerplate `json` management to get connected to `lerna`
  - updating documentation
  - future costs will include publishing additional packages

## What are the mid-term and final “exams” to check for success?

Write and test an example, a la #163.

bollwyvl:
## References

- fixes #561
- alternative to `#???`

## Code changes

- [x] in `jupyterlab-lsp/index.ts`, expose some minimal functional APIs from
  - [x] `jupyterlab-lsp/api/extractor.ts`
  - [x] `jupyterlab-lsp/api/overrides.ts`
- [x] adds `_example-extractor` to test the API 
- [ ] busy work while doing the above
  - [x] fixes a couple linter warnings
  - [x] normalize `rimraf` dependency across `package.json`s
  - [x] updates some flags for `robotframework 4.0` 
  - [x] add `--preserveWatchOutput` to `watch`

## User-facing changes

- n/a

- n/a
  - authors of extensions for the new versions would be able to easily add extractor extensions, without shipping copies of the files (by deep referencing `lib`)

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [x] changelog entry


krassowski:
Sorry for the mess in the commits!

bollwyvl:
I added a minimal package with a simple test to help keep us honest :grimacing:

packages/_example-extractor/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "author": "JupyterLab-LSP Development Team",
+  "files": [
+    "lib/**/*.{js,ts}"
+  ],
+  "dependencies": {
+    "@krassowski/jupyterlab-lsp": "^3.4"
```
krassowski, line 13:
```suggestion
    "@krassowski/jupyterlab-lsp": "^3.5"
```
Though the versions need to be bumped elsewhere for it to compile I think

bollwyvl: yeah, seems out of scope for this pr...

bollwyvl:
still doin' some npm solvin'...

packages/_example-extractor/package.json
krassowski: sure, can do it later, just noted to remember!

bollwyvl:
Trying to roll back, but concerned the depedency chain of `lsp-ws` might be getting un-upgradeable...

packages/_example-extractor/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  ILSPCodeExtractorsManager,
+  RegExpForeignCodeExtractor
+} from '@krassowski/jupyterlab-lsp';
+
+const NS = '@krassowski/jupyterlab-lsp-example-extractor';
+
+export const extractor = new RegExpForeignCodeExtractor({
+  language: 'foo',
+  pattern: '^%%(foo)( .*?)?\n([^]*)',
+  extract_to_foreign: '$3',
```
krassowski, line 16:
```suggestion
  foreign_capture_groups: [3],
```

bollwyvl: 93e7d4d

bollwyvl:
Tests are looking good locally, no further changes planned...

bollwyvl:
Seeing this locally, too:
```
'css:div.lsp-statusbar-item' not found.
```

bollwyvl:
Ah, there's some bad entropy...

bollwyvl:
Current plan (testing locally, _before_ pushing):

- revert `packages/jupyterlab-lsp/src/tokens.ts`
- add a `packages/jupyterlab-lsp/src/api/index.ts` that gets exported in `packages/jupyterlab-lspsrc//index.ts`
- add `api/extractor.ts` and `api/overrides.ts`, export these in the `api/index.ts`
- hopefully not get circular import problems!

bollwyvl:
Interactive testing looking good locally.

bollwyvl:
Cool, some stuff is coming up :heavy_check_mark:

bollwyvl:
aside from OSX 3.8 which is crapping out on `master`, all looks good!

krassowski (approved):

Issue #274, opened 2020-05-27
test-drive report


ivanov:
Hi there,

I tried out JupyterLab-LSP and wanted to share my experience. You can find it here:
https://github.com/ivanov/jlsp-test-drive/blob/master/README.md

The short version is many things did not work for me, and most of them resulted in a notebook server console log error message that looked something like this:

```
Bad message (TypeError('not all arguments converted during string formatting')): {'name': 'LabApp', 'msg': "s couldn't write message: %s", 'args': (<LspStdIoWriter(parent=<LanguageServerSession(language_server=pyls, argv=['/home/pi/.virtualenvs/jlsp/bin/pyls'])>)>, 'Content-Length: 218\r\n\r\n{"jsonrpc":"2.0","id":276,"method":"textDocument/references","params":{"context":{"includeDeclaration":true},"textDocument":{"uri":"file:///home/pi/code/pid/wpid/lsp/highlight.py"},"position":{"line":1,"character":2}}}'), 'levelname': 'ERROR', 'levelno': 40, 'pathname': '/home/pi/.virtualenvs/jlsp/lib/python3.8/site-packages/jupyter_lsp/stdio.py', 'filename': 'stdio.py', 'module': 'stdio', 'exc_info': (<class 'BrokenPipeError'>, BrokenPipeError(32, 'Broken pipe'), <traceback object at 0x7f3658674240>), 'exc_text': None, 'stack_info': None, 'lineno': 163, 'funcName': 'write', 'created': 1590606815.4147756, 'msecs': 414.7756099700928, 'relativeCreated': 1358175.628900528, 'thread': 139871523256128, 'threadName': 'MainThread', 'processName': 'MainProcess', 'process': 84827}
    Traceback (most recent call last):
      File "/home/pi/.virtualenvs/jlsp/lib/python3.8/site-packages/jupyter_lsp/stdio.py", line 161, in write
        await convert_yielded(self._write_one(response.encode("utf-8")))
      File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
        result = self.fn(*self.args, **self.kwargs)
      File "/home/pi/.virtualenvs/jlsp/lib/python3.8/site-packages/jupyter_lsp/stdio.py", line 169, in _write_one
        self.stream.write(message)
    BrokenPipeError: [Errno 32] Broken pipe
```

I used Python 3.8.3 on Debian unstable with 
```
jupyter-lsp==0.8.0
[...]

Issue #20, opened 2019-09-05
Unit test connection and connection restoration after failure


krassowski:
Follow up on #4. Two major scenarios to test:
- LSP server not available when notebook starting, but available afterwards
- connection to the LSP server temporarily lost

PR #563, opened 2021-03-21
[WIP] Ensure recovery from BrokenPipeError

krassowski:
## References

Fixes part of #20, and maybe #274. Well, it does not work yet (if I remember correctly), but the tests might come in handy.

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry


cinuor:
```python
  183 class LspStdIoWriter(LspStdIoBase):
  184     """Language Server stdio Writer"""
  185
  186     async def write(self) -> None:
  187         """Write to a Language Server until it closes"""
  188         while not self.stream.closed:
  189             message = await self.queue.get()
  190             try:
  191                 body = message.encode("utf-8")
  192                 response = "Content-Length: {}\r\n\r\n{}".format(len(body), message)
  193                 await convert_yielded(self._write_one(response.encode("utf-8")))
  194             except BrokenPipeError:
  195                 #  self.queue.task_done()
  196                 # propagate broken pipe errors
  197                 raise
  198             except Exception:  # pragma: no cover
  199                 # catch other (hopefully mild) exceptions
  200                 self.log.exception("%s couldn't write message: %s", self, response)
  201             finally:
  202                 self.queue.task_done()
  203
  204     @run_on_executor
  205     def _write_one(self, message) -> None:
  206         self.stream.write(message)
  207         self.stream.flush()
```

Hi, I delete the code at line 195 in `stdio.py`, and it works

1. ` self.queue.task_done()`  is called twice, and the BrokenPipeError will be covered by ValueError("task_done() called too many times").
2. The function `_write_lsp` does not catch BrokenPipeError，then the process will not be initialize again.

PR #565, opened 2021-03-22
Release v3.5.0

krassowski:
New release :taco:


bollwyvl:
~~Bah, windows, and we didn't get logs :sob: Let me know if there's anything (short of debugging that, i guess) i can help with!~~

Hooray!

krassowski:
I believe it is just a hiccup that happens on some runners. I checked the previous run on master and it was fine. If it repeats on master after the merge there will be logs this time :)

PR #576, opened 2021-04-04
Use ICodeMirror, more type imports, order plugin for fixer

### LINKED ISSUE #575: Private CodeMirror shipped, used in a number of places instead of The CodeMirror
## Description

If a downstream extension gets a handle on...

- on `@jupyterlab/codemirror#ICodeMirror.CodeMirror` and adds things (e.g. modes),  
- our `IExtractorManager` and adds an extractor that references that new mode

...these will not be found by the Extractor API, and syntax highlighting doesn't get updated properly.

## Reproduce

> this is kinda hard to demonstrate right now, working to get our thing open sourced

- Add a codemirror mode
- add an extractor that references the mode
- make a document that gets picked up by the extractor
- continue to see the host syntax highlighting 

## Expected behavior

- see guest language syntax highlighting 

## Context

- Operating System and version:
- Browser and version: Firefox 78 LTS
- JupyterLab version: 3.0.12

By (laborious) inspection, one can see that the distributed js bundle indeed contains a second copy of CodeMirror. Again (laboriously) one can track down `get_mode`, and see that the private codemirror is used (current release calls it `In().findModeByMIME` in obfuscated-js-speak).

## Design Ideas

The `ICodeMirror.CodeMirror` token is first available inside of the `activate` step of a lab plugin. We probably just need a static API somewhere that lets us access it (and its typings, which include `meta` and a few other plugins) from other files, or put it on `lab_integration`. Handing it down in every signature sounds... awful.

If we're just using statics, it's probably going to be fine: we'll be able to `import type CodeMirror` in many places, and then reference the static. If we're subclassing `CodeMirror` classes, it's gonna get real messy... to work around this issue, we actually create a dynamic subclass of `CMSyntaxHighlighting` and overload the private method :blush:, then update the factory :scream: .

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/gh-575-avoid-private-cm?urlpath=lab)

## References

- fixes #575

## Code changes

- [x] uses `import type` for most/all `import 'codemirror'`
  - _should_ help the compiler/webpack know we don't want another codemirror
- [x] uses `ICodeMirror.CodeMirror` for the `CMSyntaxHighlighting.get_mode`
  - this looked like the only case of actually calling a "dirty" static method
  - however, we may want to hoist `codeMirror` to the baseline `lab_integration`...
- [x] add the `import-plugin`  plugin for eslint to see if everything's good
  - this is _way_ more aggresssive than the default, but at least _can_ fix most of the issues (let it run a few times with tweaks)
  - we lot some comments to sorting
  - collapsed some duplicated imports
  - doesn't like our `export class` and `export namespace` pun: i never did much, either, but here we are
  - apparently we _have_ been making lots of cyclical imports... maybe they're just types, but yeah. ~200 errors
    - had to disable for now
  - would like some way to say, _only allow importing `codemirror` with `import type`_, haven't found yet
  - does require building with `tsc` (and all the schema stuff) _before_ linting... 
    - [x] updated `jlpm bootstrap` as such
- [x] add a test to verify we aren't shipping codemirror
  - i am not exactly sure how to do this... when it does a `dev` build, the `node_modules__codemirror__codemirror_js` shows up... but perhaps there's a way to look at a production `remoteEntry` and find magic strings...

## User-facing changes

- n/a

- n/a

## Chores

- [x] linted 
- [x] tested 
- [x] documented 
- [x] changelog entry

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
     "vscode-css-languageserver-bin": "^1.4.0",
     "vscode-html-languageserver-bin": "^1.4.0",
     "vscode-json-languageserver-bin": "^1.0.1",
-    "yaml-language-server": "^0.12.0",
-    "vscode-json-languageservice": "^3.9.1 <3.10.0"
+    "vscode-json-languageservice": "^3.9.1 <3.10.0",
+    "yaml-language-server": "^0.12.0"
   },
   "husky": {
     "hooks": {}
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   },
   "private": true,
   "scripts": {
-    "bootstrap": "jlpm --no-optional --prefer-offline && lerna bootstrap && jlpm lint && jlpm clean && jlpm build",
+    "bootstrap": "jlpm --no-optional --prefer-offline && lerna bootstrap && jlpm clean && jlpm build && jlpm lint",
     "build": "jlpm build:schema && jlpm build:meta && jlpm build:ws && jlpm build:labextension",
     "build:schema": "lerna run build:schema --stream",
     "build:meta": "lerna run build --stream --scope @krassowski/jupyterlab-lsp-metapackage",
--- a/scripts/distcheck.py
+++ b/scripts/distcheck.py
@@ -1,0 +1,73 @@
+"""sanity checks for webpack vs python packaging"""
+import subprocess
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).parent.parent
+
+# known packages that cause problems if vendored in webpack
+BAD_CHUNK_PATTERNS = {
+    (
+        "codemirror_codemirror",
+        "codemirror_lib",
+    ): """
+    Please ensure is CodeMirror is imported by type, e.g.
+
+        import type * as CodeMirror from 'codemirror'
+
+    see https://github.com/krassowski/jupyterlab-lsp/issues/575
+    """
+}
+
+# canary file created by a debug build
+BUILD_LOG = "build_log.json"
+
+# where the packages are
+PY_PACKAGES = ROOT / "python_packages"
+
+# just one, for now
+LABEXTENSION_PATHS = [
+    "jupyterlab_lsp/jupyterlab_lsp/labextensions/@krassowski/jupyterlab-lsp"
+]
+
+
+def check_webpack():
+    """verify certain packages would not be bundled
+
+    this it to avoid distributing private copies of modules
+    """
+    for path in LABEXTENSION_PATHS:
+        build_log = PY_PACKAGES / path / BUILD_LOG
+        print("checking for", build_log)
+        if not build_log.exists():
+            print(build_log, "missing, doing verbose debug build")
+            subprocess.check_call(["jlpm", "lerna", "run", "build:labextension:dev"])
+            break
+
+    bad = []
+    good = 0
+
+    for path in LABEXTENSION_PATHS:
+        for asset in (PY_PACKAGES / path / "static").rglob("*.js"):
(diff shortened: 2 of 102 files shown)


bollwyvl:
A heads-up: I am unlikely to finish this work in the next few days, unless everything basically works the first time...

bollwyvl:
I guess re: size: _not_ shipping codemirror probably saves ~150kb, so we could start tracking the wheel/sdist size, and fail builds that increase it above _a certain size_... I don't see any particular tools out there for doing so, but would be easy enough to add to our test suite, if we don't mind hoisting the hardcoded value to... somewhere. Maybe just expect it in an environment variable, and ignore if not set.

bollwyvl:
Well, looks like some stuff is passing, so it's probably not _entirely_ broken... once I can pull the built artifacts, I'll try to put together a CI check for the artifact size.

bollwyvl:
This critical bit looks good locally (finally):

Screenshot from 2021-04-04 14-21-04

bollwyvl:
Also the sdist artifact size appears about right: 310kb (from pypi) vs 251.5kb (just built).

Unfortunately, the `remoteEntry` neither has an explicit reference to `'codemirror'`... still trying to dig up how we might be _sure_ it's correct... `package.json` also not very helpful...

package.json
```diff
@@ -25,8 +25,8 @@
     "vscode-css-languageserver-bin": "^1.4.0",
     "vscode-html-languageserver-bin": "^1.4.0",
     "vscode-json-languageserver-bin": "^1.0.1",
-    "yaml-language-server": "^0.12.0",
-    "vscode-json-languageservice": "^3.9.1 <3.10.0"
+    "vscode-json-languageservice": "^3.9.1 <3.10.0",
```
krassowski, line 28: Do you remember why we have the pin for `<3.10.0`? It looks like we could update to [3.11.0 or 4.0](https://github.com/microsoft/vscode-json-languageservice/blob/main/CHANGELOG.md)

bollwyvl: alas, i do not... it's one of the language servers that might not get updated very often...

krassowski: Will try bumping it then, to make sure we don't stay behind.

bollwyvl: ah, it was yaml. they are up on 0.17 now, so _presumably_ work with the newer service.

Of course, it is irrelevant to our JSON language server, because it's a hella old static dump of... some snapshot or another. Anyhow, would prefer to not move too many more rocks on this PR... very scared of finding more things.

scripts/distcheck.py
```diff
@@ -0,0 +1,73 @@
...
+        "codemirror_codemirror",
+        "codemirror_lib",
+    ): """
+    Please ensure is CodeMirror is imported by type, e.g.
+
+        import type * as CodeMirror from 'codemirror'
+
+    see https://github.com/krassowski/jupyterlab-lsp/issues/575
+    """
+}
```
bollwyvl, line 20: Test of _not_ having `codemirror_codemirror` or `codemirror_lib` in our dev build is the best I could come up with, seems to be working:

https://github.com/krassowski/jupyterlab-lsp/pull/576/checks?check_run_id=2265592002#step:15:50

bollwyvl: Welp, guess i'll fix that typo and call this done...

bollwyvl:
Yep, tried some more looking, and short of upstreaming something to the eslint plugin, don't see how we can enforce the codemirror issue. The post-dev-build test will have to be good enough for now.

krassowski (approved):

Issue #579, opened 2021-04-16
Jupyter-lsp as a strange cp1252 problem on Windows 10


stonebig:
## Description
on windows 10 / Python-3.8.9 / "native pip install" (not anaconda)
with jupyterlab-3.0.14 and jupyter_lsp-3.0.5

opening https://github.com/winpython/winpython_afterdoc/blob/master/docs/Winpython_checker.ipynb 

on the cell having a specifically Utf-8 character not ok on cp1252 , there is  bug
![image](https://user-images.githubusercontent.com/4312421/115080172-9bf1d400-9f02-11eb-9a39-c07c4ae5adc5.png)

![image](https://user-images.githubusercontent.com/4312421/115078455-eaea3a00-9eff-11eb-8f4a-8f732989d9f9.png)

searching for the character via notepad++
![image](https://user-images.githubusercontent.com/4312421/115078544-0bb28f80-9f00-11eb-9f1d-2e3b4c23ed6f.png)

What I don't catch, is that all shall be python-3 and saved under utf-8, so what  "cp1252" is doing there ?

## Context

- Operating System and version: Windows 10 (french latest)
- Browser and version: Chrome 90 (latest)
- JupyterLab version: 3.0.14 (latest)

<details><summary>Required: installed server extensions</summary>
<pre>
Paste the output from running `jupyter server extension list` (JupyterLab >= 3)
or `jupyter serverextension list` (JupyterLab < 3) from the command line here.
You may want to sanitize the paths in the output.
.....python-3.8.9.amd64\lib\site-packages\jupyter_server\transutils.py:13: FutureWarning: The alias `_()` will be deprecated. Use `_i18n()` instead.
  warnings.warn(warn_msg, FutureWarning)
Config dir: ....\python-3.8.9.amd64\etc\jupyter
    dask_labextension enabled
    - Validating dask_labextension...
      dask_labextension 5.0.1 ok
    jupyter_lsp enabled
    - Validating jupyter_lsp...
      jupyter_lsp 1.1.4 ok
    jupyter_server_proxy enabled
    - Validating jupyter_server_proxy...
      jupyter_server_proxy  ok
    jupyter_server_mathjax enabled
    - Validating jupyter_server_mathjax...
      jupyter_server_mathjax  ok
    jupyterlab enabled
    - Validating jupyterlab...
      jupyterlab 3.0.14 ok
    jupyterlab_git enabled
    - Validating jupyterlab_git...
[...]

PR #580, opened 2021-04-16
patch jupyter_lsp-1.1.4 for 'native' Windows users

stonebig:
on Windows, even if cp1252, a Notebook must be written in utf-8.
Anaconda on Windows is not 'native', so doesn't have the problem.

shall solve https://github.com/krassowski/jupyterlab-lsp/issues/579

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [x] changelog entry


bollwyvl (approved):
LGTM! thanks @stonebig !

PR #582, opened 2021-04-24
Add Python LSP Server

maresb:
I didn't have time to test this or to fill out this pull request properly. :cry: Sorry!!!

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [x] changelog entry

PR #584, opened 2021-04-25
Switch to python-lsp-server for testing

krassowski:
## References

Follow up on #582 fixing Python module name, actually enabling it in tests and changing tests to use it rather than the old Palantir's version.

## Code changes

Updated references of `pyls` to `pylsp`.

## User-facing changes

The readme will now recommend installation of pylsp instead of pyls.

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


python_packages/jupyter_lsp/jupyter_lsp/tests/conftest.py
```diff
@@ -20,6 +20,7 @@
     "dockerfile-language-server-nodejs",
     "javascript-typescript-langserver",
     "pyls",
```
bollwyvl, line 22: so we can take this out now?

krassowski: I guess so. Otherwise we would need to have two different environments, and one with legacy jedi and IPython which is IMO nto worth it.

bollwyvl:
Yeah, and it's probably time to drop py3.6 as well as it's ipython won't
work with newer jedi... Would simplify my robot pr, too!

krassowski:
Well, the tests work for now - I guess it it install an older version on 3.6. We should certainly drop 3.6 later this year if it is slowing us down.

Issue #440, opened 2020-12-30
Feature: Support Multiple Autocompleters


mlucool:
## Elevator Pitch

While LSP has a lot of great features, sometimes I don't want to use its autocompleter. Other autocompleters offer _different_ features that are better for certain workflows. LSP should let me choose my default autocompleter and change that per notebook.

## Motivation

Consider this code:
```
def foo(a):
    if a:
        return A()
    return B()
x = foo(True)
# New cell
x.<tab>
```

IPykernel's autocomplete knows about runtime properties of objects, which I don't believe LSP does. This means I'll get autocomplete results for both A and B in the above example, but ipykernel will only show A. Another autocompleter is [kite](https://github.com/kiteco/jupyterlab-kite) which uses ML to predict the best autocomplete results based on context (including more than just the function). Finally, LSP's autocomplete is (typically) fast and can be used without a running kernel. 

All this is to say is that none of these are better than the rest, just different. Depending on my use case or personal preference, I'd like to be able to use other autocompleters without losing the rest of the features of LSP.

## Design Ideas

Add a global setting that can store the "default" autocompleter then add a command to switch it per notebook. It's not clear if this feature should be requested on lab itself and LSP implement it or if this is all LSP specific....

krassowski:
Thanks, this is a good idea. We already combine results from both the kernel and the LSP but this is indeed problematic because either one of those can be slow and we need to wait for both to complete to display the result. Giving user the choice is the right thing. Enabling Kite is also a good idea. It was previously brought up in #236; since then we split the codebase making it more developer friendly to tackle those issues.

mlucool:
https://github.com/krassowski/jupyterlab-lsp/pull/586 solves this. Closing.

PR #586, opened 2021-04-27
Allow to disable LSP or Kernel completions independently.

Carreau:
Some user love LSP for many of its features, but prefer to keep
completions via the kernel.

This adds two boolean flags in the configuration to not send completions
request to LSP and not send completions requests to the kernel.

## References

#440  - I'm not sure how to toggle it on a per-notebook basis; I can likely add local state to LSPConnector which I guess is persisted on a per notebook basis but would welcome guidance on the best place to store that.

## Code changes

- Add 2 boolean to the setting. 
- turns those two booleans into properties on the LSPConnector, 
- when false skip requesting completions for the corresponding source.

## User-facing changes

None by default. User may disable one or two completer sources.

None

## Chores

- [x] linted 
- [ ] tested - manually for now I can try to add an automatic test. 
- [x] documented  
- [x] changelog entry 

Note: I haven't touched typescript and frontend in years,  so I'm likely doing some things wrong;

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,8 @@ from the Language Server (in notebook).
 If the kernel is too slow to respond promptly only the Language Server suggestions will be shown (default threshold: 0.6s).
 You can configure the completer to not attempt to fetch the kernel completions if the kernel is busy (skipping the 0.6s timeout).
 
-You can deactivate the kernel suggestions by adding `"kenel"` to the `disableCompletionsFrom` in the `completion` section
-of _Advanced Settings_. Alternatively if you _only_ want kernel completions you can add `"languageServer"` to the same
+You can deactivate the kernel suggestions by adding `"Kernel"` to the `disableCompletionsFrom` in the `completion` section
+of _Advanced Settings_. Alternatively if you _only_ want kernel completions you can add `"LSP"` to the same
 setting; Or add both if you like to code in hardcore mode and get no completions, or if another provider has been added.
 
 ### Rename
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ from the Language Server (in notebook).
 If the kernel is too slow to respond promptly only the Language Server suggestions will be shown (default threshold: 0.6s).
 You can configure the completer to not attempt to fetch the kernel completions if the kernel is busy (skipping the 0.6s timeout).
 
+You can deactivate the kernel suggestions by setting the `useKernelCompletions` to `false` in the `completion` section
+of advanced settings. Alternatively if you _only_ want kernel completions you can set `useLspCompletions` to `false`.
+Or both if you like to code in hardcore mode and get no completions.
+
 ### Rename
 
 Rename variables, functions and more, in both: notebooks and the file editor.
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ from the Language Server (in notebook).
 If the kernel is too slow to respond promptly only the Language Server suggestions will be shown (default threshold: 0.6s).
 You can configure the completer to not attempt to fetch the kernel completions if the kernel is busy (skipping the 0.6s timeout).
 
-You can deactivate the kernel suggestions by setting the `useKernelCompletions` to `false` in the `completion` section
-of advanced settings. Alternatively if you _only_ want kernel completions you can set `useLspCompletions` to `false`.
-Or both if you like to code in hardcore mode and get no completions.
+You can deactivate the kernel suggestions by adding `"kenel"` to the `disableCompletionsFrom` in the `completion` section
+of _Advanced Settings_. Alternatively if you _only_ want kernel completions you can add `"languageServer"` to the same
+setting; Or add both if you like to code in hardcore mode and get no completions, or if another provider has been added.
 
 ### Rename
 
--- a/packages/jupyterlab-lsp/schema/completion.json
+++ b/packages/jupyterlab-lsp/schema/completion.json
@@ -42,7 +42,7 @@
       "description": "The time to wait for the kernel completions suggestions in milliseconds. Set to 0 to disable kernel completions, or to -1 to wait indefinitely (not recommended)."
     },
     "disableCompletionsFrom": {
-      "description": "The sources from which to exclude completion from. Possible values include 'kernel', 'languageServer'.",
+      "description": "The sources from which to exclude completion from. Possible values include 'Kernel', 'LSP'.",
       "type": "array",
       "default": [],
       "uniqueItems": true
--- a/packages/jupyterlab-lsp/schema/completion.json
+++ b/packages/jupyterlab-lsp/schema/completion.json
@@ -41,17 +41,11 @@
       "type": "number",
       "description": "The time to wait for the kernel completions suggestions in milliseconds. Set to 0 to disable kernel completions, or to -1 to wait indefinitely (not recommended)."
     },
-    "useKernelCompletions": {
-      "title": "Request Kernel Completion",
-      "type": "boolean",
-      "default": true,
-      "description": "Whether to send completions request to the kernel."
-    },
-    "useLspCompletions": {
-      "title": "Request LSP Completion",
-      "type": "boolean",
-      "default": true,
-      "description": "Whether to send completions request to the lsp server."
+    "disableCompletionsFrom": {
+      "description": "The sources from which to exclude completion from. Possible values include 'kernel', 'languageServer'.",
+      "type": "array",
+      "default": [],
+      "uniqueItems": true
     },
     "waitForBusyKernel": {
       "title": "Wait for kernel if busy",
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -66,6 +66,10 @@ from the Language Server (in notebook).
 If the kernel is too slow to respond promptly only the Language Server suggestions will be shown (default threshold: 0.6s).
 You can configure the completer to not attempt to fetch the kernel completions if the kernel is busy (skipping the 0.6s timeout).
 
+You can deactivate the kernel suggestions by setting the `useKernelCompletions` to `false` in the `completion` section
+of advanced settings. Alternatively if you _only_ want kernel completions you can set `useLspCompletions` to `false`.
+Or both if you like to code in hardcore mode and get no completions.
```
bollwyvl, line 71: @krassowski would have to weigh in on this: if both of ours are turned off, is there _some other place_ they might come from?

README.md
```diff
@@ -66,6 +66,10 @@ from the Language Server (in notebook).
 If the kernel is too slow to respond promptly only the Language Server suggestions will be shown (default threshold: 0.6s).
 You can configure the completer to not attempt to fetch the kernel completions if the kernel is busy (skipping the 0.6s timeout).
 
+You can deactivate the kernel suggestions by setting the `useKernelCompletions` to `false` in the `completion` section
+of advanced settings. Alternatively if you _only_ want kernel completions you can set `useLspCompletions` to `false`.
```
bollwyvl, line 70: Nitpick `_Advanced Settings_`.

packages/jupyterlab-lsp/src/features/completion/completion_handler.ts
```diff
@@ -271,12 +283,16 @@ export class LSPConnector
...
             lsp_promise.catch(p => p)
-          ]).then(([kernel, lsp]) =>
-            this.merge_replies(
-              [this.transform_reply(kernel), lsp],
-              this._editor
-            )
-          );
+          ]).then(([kernel, lsp]) => {
+            let replies = [];
+            if (this.use_kernel_completions) {
```
bollwyvl, line 288: perhaps just check if these are non-null, as it seems... not unreasonable to expect a setting to have effect on an in-flight request.

```ts
if (kernel != null)
```

packages/jupyterlab-lsp/schema/completion.json
```diff
@@ -41,6 +41,18 @@
...
+      "title": "Request Kernel Completion",
+      "type": "boolean",
+      "default": true,
+      "description": "Whether to send completions request to the kernel."
+    },
+    "useLspCompletions": {
+      "title": "Request LSP Completion",
+      "type": "boolean",
+      "default": true,
+      "description": "Whether to send completions request to the lsp server."
```
bollwyvl, line 54: Yes, we have LSP in the top-level description, and the name of the project, sure, but perhaps `useLangugeServerCompletions`. And certainly `to the language server`

bollwyvl: Actually, as i expand this big-ol-file to look below... we already have `kernelCompletionsFirst`. I wonder if these could be condensed into something like the below, and put a `4.0` deprecation warning on KCF:
```yaml
completionSources:
  description: The sources from which to fetch completions. Order determines tiebreakers. An empty list will disable completions provided by this plugin.
  type: array
  default: [languageServer, kernel]
  uniqueItems: true
  items:
    enum: [languageServer, kernel]
```

Thoughts @krassowski

bollwyvl:
Re: notebook. Gah. Yeah. Been thinking about this for a long time.

Having implemented this in a number of places, I'd love it if per-notebook settings just worked across _every_ plugin in a `jupyterlab-settings` key that followed the structure of `overrides.json` (no JSON5, please), but that might be beyond us to be able to declare as a thing.

Alternately, we've talked about having per-(glob|fnmatch|regex) (probably the latter) settings which could be configured per-plugin.

At any rate, I don't think we tackle that on _this PR_, but should probably hoist it (at least) to an issue on _this repo_, if not to core.

bollwyvl:
Commented here: https://github.com/jupyterlab/jupyterlab/issues/8148#issuecomment-827250995

bollwyvl:
Woo green!

krassowski:
Thank you @Carreau. I wonder if having this as a single setting with an array of completion sources to disable (e.g. `disableCompletionsFrom`) would make more sense given that we may want to make it possible for other extensions (Kite, snippets) to integrate well with our completions mechanism (and ideally move the motion of having multiple completion sources to the core).

packages/jupyterlab-lsp/schema/completion.json
krassowski: The `completionSources` would work for me to but I would want to avoid enum to allow for completions from other sources too. In the end the plan is that it will be all removed when JupyterLab 4.0 is released and takes care of that for us (this is when we PR a multi-source completion mechanism which will prevent extension clashes).

README.md
krassowski: If am not mistaken we are taking an absolute control of completer and nothing else can provide completions just yet, but this will change _very soon_ allowing Kite and Snippets to provide custom completion items to our endpoint (and afterwards we will move it to JupyterLab for 4.0).

By _very soon_ I mean not earlier than end of May (sadly won't have time to on it earlier).

packages/jupyterlab-lsp/schema/completion.json
Carreau: I'm fine with anything you like.
I also thought that bool could at some point in the future allow to generate checkboxes, and easier to toggle via an action, but up to you.

Carreau:
> I wonder if having this as a single setting with an array of completion sources to disable (e.g. `disableCompletionsFrom`) would make sense given that we will want to make it possible for other extensions (Kite, snippets) to integrate well with our completions mechanism (and ideally move the motion of having multiple completion sources to the JupyterLab core).

That was my initial thought as well, either an include and exclude list. as I pointed above that would be fine with me; though it might be a bit harder to provide actions to enable disable, I mean that's only a couple of lines of code so I can switch to that.

Carreau:
Updated to take comments into accounts.

README.md
```diff
@@ -66,6 +66,10 @@ from the Language Server (in notebook).
 If the kernel is too slow to respond promptly only the Language Server suggestions will be shown (default threshold: 0.6s).
 You can configure the completer to not attempt to fetch the kernel completions if the kernel is busy (skipping the 0.6s timeout).
 
+You can deactivate the kernel suggestions by adding `"kenel"` to the `disableCompletionsFrom` in the `completion` section
```
krassowski, line 69:
```suggestion
You can deactivate the kernel suggestions by adding `"kernel"` to the `disableCompletionsFrom` in the `completion` section
```

krassowski:
Thank you. The one final thing I would like your feedback on is naming of the options. I already introduced a concept of "CompletionSource" to make it easy to add new sources and to display the name of the source in the interface to make user aware why they do not see e.g. documentation for a specific completion item. These are currently named "Kernel" and "LSP":

https://github.com/krassowski/jupyterlab-lsp/blob/39010530eba400bffc56282709343e9fcf8bc778/packages/jupyterlab-lsp/src/features/completion/completion_handler.ts#L398-L402

https://github.com/krassowski/jupyterlab-lsp/blob/39010530eba400bffc56282709343e9fcf8bc778/packages/jupyterlab-lsp/src/features/completion/completion_handler.ts#L451-L455

Should we alight the source "name" that is displayed with the identifier in settings? I am not saying that they need to be identical. Just curious if you have thoughts whether using "languageServer" (identifier) + "Language Server" (name in UI) is actually better  than using "lsp" (idnetifier) and "LSP" (name in the UI). Both would work for me.

bollwyvl:
I'm :+1: to `languageServer` and `Language Server`. Better still would be if we could get the actual name of the server/kernel in the UI. At the end of it, they're both being strictly transformed into LSP messages.

Further, if the jiggery-pokery around the `transorm_response` could make its way _inside_ the kernel thingy, so they both returned LSP `CompletionResponse` (or whatever it's called), that would further simplify the code, and remove (some) instances of special casing.

Carreau:
I've change the options "LSP" and "Kernel" for the options values;  I like "LSP",  it's short and hard to misspell – which is good especially if you have to type it in the settings.

I don't have particular preferences otherwise.

@bollwyvl I agree with the refactor, but would prefer to do it separately, as I'm not yet that comfortable with typescript. I was thinking that a dict of names/promises or `Promise.resolve(null)` could be an easy generic way of having multiple providers and unconditionally merge (even if there is only one provider) would be cleaner.

bollwyvl:
LGTM

krassowski (approved):
Thank you!

PR #587, opened 2021-04-29
add Pyright langserver spec

yuntan:
## References
related to #534

## Code changes
Added Pyright langserver spec.

## User-facing changes
User can use Pyright with JupyterLab.

## Backwards-incompatible changes
Nothing.

## Chores

- [x] linted 
- [x] tested 
- [x] documented 
- [x] changelog entry 

I wonder how to add default configuration for language server. In this case, setting `python.analysis.useLibraryCodeForTypes` to  `true` by default might helpful for users.

PR #588, opened 2021-04-29
Add priority setting for servers handling the same language

krassowski:
## References

Add initial priority setting as discussed in #587

## Code changes

- replace `getServerId` with `getMatchingServers` which returns names of all matching servers, sorted by priority (even though we only use the first one now, it will come in handy later to have an ordered list of those)
- add some useful console.warn and console.errors messages (which might end up being shown in the GUI later)
- fix types in few places, use types imported from LSP rather than manually re-typed ones
- make `TServerKeys = TLanguageServerId` to avoid too broad types (one was a string, the other was a literal union; now both use the literal union and one could be removed but keeping it for now because they have distinct meaning)

## Chores

- [x] linted 
- [ ] tested 
- [x] documented 
- [x] changelog entry


packages/jupyterlab-lsp/src/components/statusbar.tsx
```diff
@@ -501,19 +516,17 @@ export namespace LSPStatus {
...
         );
-        if (servers.length > 1) {
-          console.warn('More than one server per language for', language);
-        }
-        if (servers.length === 0) {
+        if (server_ids.length === 0) {
           continue;
         }
-        let server = servers[0];
+        // For now only use the server with the highest priority
```
krassowski, line 525: Should probably use the connection + server Identifier here. Also should we restart all connections if preferred language server changed our just add a note that JupyterLab reload is needed for changes to go into effect? Probably the latter

bollwyvl: i think wherever possible, settings changes should take effect immediately, since have established the pattern... can see how it would be a tad onerous

atest/03_Notebook.robot
```diff
@@ -47,7 +47,7 @@ Moving Cells Around
 Foreign Extractors
     ${file} =    Set Variable    Foreign extractors.ipynb
     Configure JupyterLab Plugin
-    ...    {"language_servers": {"texlab": {"serverSettings": {"latex.lint.onChange": true}}, "bash-langauge-server": {"bashIde.highlightParsingErrors": true}}}
+    ...    {"language_servers": {"texlab": {"serverSettings": {"latex.lint.onChange": true}}, "bash-langauge-server": {"bashIde.highlightParsingErrors": true}}, "pylsp": {"priority": 1000}}
```
bollwyvl, line 50: getting long... maybe we hoist to a fixture file?

docs/Configuring.ipynb
```diff
@@ -66,7 +66,8 @@
     "    `mkdir \"new directory\"` should be split into `[\"mkdir\", \"new directory\"]`;\n",
     "    If you have Python installed, you can use `shlex.split(\"your command\")` to\n",
     "    get such an array.\n",
-    "- the `languages` which the language server will respond to, and\n",
+    "- the `languages` which the language server will respond to (language names\n",
+    "  should be given in lowercase), and\n",
```
bollwyvl, line 70: why demand lowerecase? i mean, doesn't bother me, but don't see the point of resriction yet, though may be revealed as i read more...

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -16,6 +16,12 @@
           "description": "Configuration to be sent to language server over LSP when initialized: see the specific language server's documentation for more",
           "type": "object",
           "default": {}
+        },
+        "priority": {
+          "title": "Priority of the server",
+          "description": "When multiple servers match specific document/language, the server with higher priority will be used",
```
bollwyvl, line 22: `highest` perhaps

packages/jupyterlab-lsp/schema/plugin.json
```diff
@@ -16,6 +16,12 @@
...
+          "type": "number",
+          "default": 50
```
bollwyvl, line 24: maybe require it's positive?

packages/jupyterlab-lsp/src/editor_integration/testutils.ts
```diff
@@ -240,7 +242,8 @@ function FeatureSupport<TBase extends TestEnvironmentConstructor>(Base: TBase) {
       return new LSPConnection({
         languageId: this.document_options.language,
         serverUri: 'ws://localhost:8080',
-        rootUri: 'file:///unit-test'
+        rootUri: 'file:///unit-test',
+        serverIdentifier: 'pylsp'
```
bollwyvl, line 246: now in two places, maybe hoist to `DEFAULT_SERVER_ID`?

python_packages/jupyter_lsp/jupyter_lsp/schema/schema.json
```diff
@@ -34,7 +34,7 @@
       "type": "string"
     },
     "language-list": {
-      "description": "languages supported by this Language Server",
+      "description": "languages supported by this Language Server; lowercase names",
```
bollwyvl, line 37: yeah, have made it all the way through the current changes, and don't see where cased-ness is going to bite us. if we're going to say it, it should _do_ it, and have `"pattern": "[a-z_\\-\\.\\d]+"` or some such rot.

bollwyvl:
On the main looks good. I don't remember where, but I _thought_ there was an issue where someone wanted the "full deal" on python source and another server on notebooks, but don't remember where. If we're not going to do it now, certainly don't need to talk about it in the code anywhere!

packages/jupyterlab-lsp/src/manager.ts
```diff
@@ -38,16 +48,44 @@ export class LanguageServerManager implements ILanguageServerManager {
...
+          'Cannot match server by language: language not available; ensure that kernel and specs provide language and MIME type'
+        );
       }
     }
-    return null;
+
+    return matchingSessionsKeys.sort((a, b) => {
+      const a_priority = config[a]?.priority ?? 50;
+      const b_priority = config[b]?.priority ?? 50;
+      if (a_priority == b_priority) {
```
bollwyvl, line 81: hoist default weight, tie breaker of name, hoist to method so we're not allocating new functions every time?

```ts
compareServerPriority(a: TServerId, b: TServerId) {
return (config[a]?.priority || DEFAULT_PRIORITY) - (config_b[b]?.priority || DEFAULT_PRIORITY) 
  || a.localeCompare(b);
}
```

also,

packages/jupyterlab-lsp/schema/plugin.json
bollwyvl: yeah, for sorting, non-zero, positive is much better.

atest/03_Notebook.robot
krassowski: Actually, could we have `Configure JupyterLab Plugin` merge the new settings with the default priorities (unless custom priorities are given)? This should simplify tests that have custom config.

krassowski: Or just move `"pylsp": {"priority": 1000}` to a fixture file. I think we used to have a json file for frontend settings but cannot see it.

Issue #595, opened 2021-05-10
Incompatibility with JupyterLab core npm packages 3.0.10


bsyouness:
## Description

We use a custom version of JupyterLab which depends on the following dependencies that I upgraded:
<img width="967" alt="Screen Shot 2021-05-10 at 6 07 52 PM" src="https://user-images.githubusercontent.com/2256598/117736407-5ac0ad00-b1ac-11eb-9ef5-e9560608e0b5.png">

Doing that broke the LSP extension in that the check  `if not uri.startswith(virtual_documents_uri)` in "virtual_documents_shadow.py" at line 141 now fails. This is due to an additional "/" that gets inserted in the file URI as per this example:
```
uri file://c:/XXX/.virtual_documents/1%20Migration%20Guide.ipynb doesn't start with file:///c:/XXX/.virtual_documents
```

## Reproduce

Bump up to the latest version of JupyterLab.

krassowski:
Thank you for the heads up! Would you like to open a PR to fix this (and bump the versions?).

bsyouness:
Sure yeah!

krassowski:
It would be very welcome :) It seems to be related to url-parse 1.5.1 update and I think that is causing the current failures..

krassowski:
@bollwyvl do you have an opinion on this one?

bollwyvl:
Aw, man, I sure love windows paths. LOVE'm. It's unlikely i'll have any bandwidth to work on this this week, but one of them has to be right. Looks like [the python tests](https://github.com/krassowski/jupyterlab-lsp/blob/master/python_packages/jupyter_lsp/jupyter_lsp/tests/test_paths.py#L35) expect three slashes. This jives with my understanding of the `//` being the protocol separator, and then the `<nothing here>/` being the _authority_.

krassowski:
https://github.com/unshiftio/url-parse/issues/203 looks like an upstream bug report for this issue.

bollwyvl:
Whew, saved. So if anything, we need to specify _not_ 1.5.1, regex ddos be
damned?

krassowski:
I think the troublesome fix went into 1.5.0. It might be difficult to revert to an earlier version since it got a) a security fix label  b) already merged and released by JupyterLab core. I was thinking about an upstream fix, or making our server side comparison be more lenient (as a special case for the lack/presence of triple /).

krassowski:
[...]

PR #599, opened 2021-05-16
Workaround file URI matching issue for now

krassowski:
## References

See #595. This is not the ultimate nor preferred solution, but let's have CI (and extension) working for now. Hopefully this can be rolled back before the next release (conditional on url-parse getting a new release with merged fix and JupyterLab 3.0.16 picking up the update).

## Chores

- [x] linted 
- [x] tested 
- [ ] documented 
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ### (unreleased)
 
 - Add ability to deactivate Kernel completions or LSP completion through the settings.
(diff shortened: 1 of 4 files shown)


bsyouness (approved):
Looks good, thank you and sorry I couldn't help more! I would maybe add a comment for where the mac stuff is happening and another comment to remove this when the fix rolls out upstream?

PR #604, opened 2021-05-20
[WIP] Implement code snippet completer

jahn96:
## References
Starts working on implementing code snippet completion.
This pull request is related to the issue #208

## Code changes

- [x] Create a package called `code-snippet-completion`
- [x] Add @krassowski/completion-manager to dev-dependencies in package.json
- [x] Add ICompletionProviderManager as optional token in activate function of your extension
- [ ] Need to fix the error on `Plugin '@krassowski/jupyterlab-lsp:completion' failed to activate` and `No provider for: @krassowski/completion-manager:ICompletionProviderManager.` - could be related to the issue of setting up the environment.
- [ ] Need to create a `codeSnippetCompletionProvider` class that implements ICompletionProviderManager
- [ ] Need to register the class

## User-facing changes

User will be able to see the autocomplete options for code snippets they have saved.

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


krassowski:
Oh, so you thought about including it in the distribution of jupyterlab-lsp? I thought that you were going to include it in the snippets extension directly instead.

kpinnipa:
We were thinking it might be better to include it in the distribution of jupyterlab-lsp, do you think it would be better to include it in the snippets extension directly?

krassowski:
> But, now I think that it would be actually better to work in the snippets extension directly because it would be easier to fetch the saved snippets and so on.

Exactly, it will be just easier for you. Plus I once the ICompletionProvider lands in JupyterLab core (again, probably for 4.0) your users will be able to have completion with snippets without having to have this LSP extension :)

krassowski:
But of course, if it is convenient for you to start working/experimenting here on this PR - absolutely feel free to do so :)

krassowski:
Hi, I am going to close this one as #600 on which one it is based was superseded by the work upstream. Please see https://github.com/jupyter-lsp/jupyterlab-lsp/pull/600#issuecomment-1060030197 for references. Hopefully we can revisit it after JupyterLab 4.0 release :)

Issue #500, opened 2021-02-01
Hoist CodeMirror Mark Widgets to a Layer-based model


bollwyvl:
## Elevator Pitch

Hoist the various kinds of mark widgets (lines, token, gutter) to an evented model. Allow for separate owners by organizing sets of marks as layers.

## Motivation

Our current design assumes that _one_ agent (usually our diagnostic feature) controls all the marks on an editor instance. At the end of the day, the implementation would _still_ need to do the same work to ensure that marks are up-to-date, but if we want to support multiple sources of information (#184) without going crazy, we're going to need a better way to organize them. 

## Design Ideas

- `CodeEditor`
  - `IMarksModel`
    - layers: `IMarkLayer[]`
      - `title`
      - `description`
      - `rank`
      - `IMark[]`
        - `LineMark`
        - `TokenMark`
        - `GutterMark`

With this, we could also offer UI (notionally as a sidebar) for hiding/reordering the mark layers.

Presumably, this could be hoisted to Lab core's `CodeEditor`... or better still a JEP-level definition of "what is a text editor", with a full JSON Schema model of all of these pieces, such that this could be integrated into a terminal-based editing experience (e.g. `ipython`) and would be better describable by assistive technologies.

bollwyvl:
Looking forward, much of this is supported by the CodeMirror 6 [DecorationSet](https://codemirror.net/6/docs/ref/#view.DecorationSet).

Another happy thing I saw was the [legacy-modes](https://github.com/codemirror/legacy-modes/tree/main/mode) repo, which should provide not only many of the modes needed to get up to feature parity, but also the blueprint for porting other novel modes.

It'll be a haul, and CM6 is still in beta, and is not going to support LSP directly, but the architecture looks a lot more like something we can build on without having to get fancy. We'll lose direct control via CSS of basically everything, but I think given what we've learned on this project, the path to supporting mobile, screen readers, and LSP features are all pointing at CM6 as the right path forward.

PR #606, opened 2021-05-23
Implement console log window messages, update lsp-ws-connection

### LINKED ISSUE #601: Support window/logMessage
## What are you trying to do?

Some servers provide useful messages to the client. We should show these in the built-in JupyterLab console or in the browser console at the very minimum.

```
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"No pyproject.toml file found."}
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":2,"message":"stubPath typings is not a valid directory."}}
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"Assuming Python platform Windows"}}
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"Searching for source files"}}
{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":3,"message":"No source files found."}}
```

https://microsoft.github.io/language-server-protocol/specification#window_logMessage

The idea would be to use ILogger from https://github.com/jupyterlab/jupyterlab/tree/master/packages/logconsole

krassowski:
## References

Fixes #601 and updates lsp-ws-connection:

## Code changes

- use `import type` if possible
- update:
   - "vscode-jsonrpc": "^6.0.0",
   - "vscode-languageserver-protocol": "^3.16.0",
   - "vscode-languageserver-types": "^3.16.0",
   - "vscode-ws-jsonrpc": "0.2.0"
- remove unused code from lsp-ws-connection
- deprecate the `emit()` and getter based API for lsp-ws-connection
- add new API with well-typed:

```
  public serverNotifications: ServerNotifications;
  public clientNotifications: ClientNotifications;
  public clientRequests: ClientRequests;
  public serverRequests: ServerRequests;
```

## User-facing changes

Screenshot from 2021-05-23 17-43-24

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [x] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/connection.ts
+++ b/packages/jupyterlab-lsp/src/connection.ts
@@ -11,6 +11,10 @@ import {
   IPosition,
   LspWsConnection
 } from 'lsp-ws-connection';
+import {
+  registerServerCapability,
+  unregisterServerCapability
+} from 'lsp-ws-connection/lib/server-capability-registration';
 import type * as rpc from 'vscode-jsonrpc';
 import type * as lsp from 'vscode-languageserver-protocol';
 import type { MessageConnection } from 'vscode-ws-jsonrpc';
@@ -196,7 +200,7 @@ class ServerRequestHandler<
     protected emitter: LSPConnection
   ) {
     // on request accepts "thenable"
-    this.connection.onRequest(method, this.handle);
+    this.connection.onRequest(method, this.handle.bind(this));
     this._handler = null;
   }
 
@@ -319,12 +323,21 @@ export class LSPConnection extends LspWsConnection {
     }
   }
 
+  sendInitialize() {
+    super.sendInitialize();
+  }
+
   protected onServerInitialized(params: lsp.InitializeResult) {
     super.onServerInitialized(params);
     while (this.documentsToOpen.length) {
       this.sendOpen(this.documentsToOpen.pop());
     }
+    // TODO: move to send Initialize after disabling overwrites in ws-connection
+    //  or maybe move the code there? How to handle logging without bringing in lumino signals?
+    this.afterInitialized();
+  }
 
+  protected afterInitialized() {
     for (const method of Object.values(
       Method.ServerNotification
     ) as (keyof ServerNotifications)[]) {
@@ -348,6 +361,40 @@ export class LSPConnection extends LspWsConnection {
     );
     this.serverRequests = this.constructServerRequestHandler<ServerRequests>(
       Method.ServerRequests
+    );
+
+    this.serverRequests['client/registerCapability'].setHandler(
+      async (params: lsp.RegistrationParams) => {
+        params.registrations.forEach(
+          (capabilityRegistration: lsp.Registration) => {
+            try {
+              this.serverCapabilities = registerServerCapability(
+                this.serverCapabilities,
+                capabilityRegistration
+              );
+            } catch (err) {
+              console.error(err);
+            }
+          }
+        );
+
+        // TODO log event
+      }
+    );
+
+    this.serverRequests['client/unregisterCapability'].setHandler(
+      async (params: lsp.UnregistrationParams) => {
+        params.unregisterations.forEach(
+          (capabilityUnregistration: lsp.Unregistration) => {
+            this.serverCapabilities = unregisterServerCapability(
+              this.serverCapabilities,
+              capabilityUnregistration
+            );
+          }
+        );
+
+        // TODO log event
+      }
     );
   }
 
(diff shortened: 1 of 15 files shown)


packages/jupyterlab-lsp/src/connection.ts
```diff
@@ -2,28 +2,63 @@
...
+        logTrace: new Signal(this),
+        setTrace: params => {
+          this.connection.sendNotification('$/setTrace', params);
+        }
+      },
+      window: {
+        showMessage: new Signal(this),
+        logMessage: new Signal(this)
+      }
+    };
```
krassowski, line 61: CC @bollwyvl . I know this is a different pattern from the existing one, but trying to experiment on doing it better. I also know the dream would be to generate typings from a JSON file; the handlers for signals could be also generated from JSON, but not sure if those would work for all handlers + some signals might be performance-critical and maybe using direct callbacks rather than lumino Signal could be better.

packages/jupyterlab-lsp/src/connection.ts
```diff
@@ -2,28 +2,301 @@
...
+  [Method.ClientRequest.DOCUMENT_HIGHLIGHT]: lsp.DocumentHighlight[];
+  [Method.ClientRequest.DOCUMENT_SYMBOL]: lsp.DocumentSymbol[];
+  [Method.ClientRequest.HOVER]: lsp.Hover;
+  [Method.ClientRequest.IMPLEMENTATION]: AnyLocation;
+  [Method.ClientRequest.INITIALIZE]: lsp.InitializeResult;
+  [Method.ClientRequest.REFERENCES]: Location[];
+  [Method.ClientRequest.RENAME]: lsp.WorkspaceEdit;
+  [Method.ClientRequest.SIGNATURE_HELP]: lsp.SignatureHelp;
+  [Method.ClientRequest.TYPE_DEFINITION]: AnyLocation;
+}
```
krassowski, line 128: Cherry-picked from #278 but Methods now grouped into enums for run-time generation of signals and handlers, CC @bollwyvl

bollwyvl: Cool! Someday we'll be able to close that one out! Over in jupyter_server meeting, we talked about some approaches for moving forward for improving reuse of specs as docs (yay) and _simple_ type packages (no ORMs, future-imperiled pydantic stuff, etc).

I don't know how this would work out here, as these are sometimes a good deal more complicated.

krassowski: The idea is to have strong separation between server/client, notification/request as otherwise those get mixed up. Also, this implementation does not require manual typing anymore. Eventually I would like to define an interface of ILSPCompletion (as you did in #278) and make everything else an implementation detail. I hope that this would allow to smoothly add the kernel approach from #278 too (as an alternative implementation of the interface).

packages/jupyterlab-lsp/src/connection.ts
bollwyvl: Welp, even lower in the stack is the lumino Message bus. Nobody talks about it much, but it is _extremely_ robust, and what makes all the core widget stuff very smooth.

But performance-critical is as performance-critical-is-measured... we'd need some no-fooling benchmarks before stepping too far off something that at least quacks likes other parts of the stack.

packages/jupyterlab-lsp/src/connection.ts
bollwyvl: Yeah, well... the upstream LSP spec hides that particular piece of information (whether a message is in a request or a notification) behind some emoji. In the various attempts at building schema for it, I had to parse that business and build actual models of the exchange sequences: not so much fun.

In the jupyter_server discussion, the idea was to get to where we can incrementally define robust, declarative, language-agnostic specs (JSON Schema, OpenAPI, AsyncAPI, kaitaistruct) for different parts of the architecture, from which many language "bindings" could be built, as close to the standard library of that language as possible. Unfortunately, the place I called out initially was the ContentsManager API... which would drag along nbformat, and a subset of kernelspec. From there, the kernel session REST, and eventually the kernel interchange itself, could be documented... that would either be the baseline for the LSP Comm approach, or at least give us enough ground to stand on to document our "special" websocket.

Semi-related, if kernels were better able to describe, and report, their special syntaxes, and ideally a syntax highlighting approach... well, that sure would make a composite language document more understandable, and a lot easier than ginning up custom typescript extensions.

More concretely here... I think _really_ supporting multiple sources (of any of these things) will end up needing a more explicit document/mark model, a la #500... in the case of a completion, documentation requests, etc. these would be extremely short-lived, but one should see whatever is available as fast as possible, but know that some other sources _might_ be coming, and be able to configure the hell out of them, e.g.

```
In []: df.<TAB>
       |----------------------------------------|
       | apply                            [lsp] |
       | some_col                         [jkm] |
       |----------------------------------------|
       | Sources [lsp] 300ms  [jkm] 100ms [gear]| 
```

Issue #282, opened 2020-06-14
How to implement LSP for a multi-language kernel (SoS)?


BoPeng:
Many thanks for your great work on language server support. I have just tried `jupyterlab-lsp`, which works great for `Python` and `R`, but unfortunately does not work for a multi-language [kernel SoS](https://github.com/vatlab/sos-notebook) that I have developed.

The idea behind of SoS is that it is a superkernel that sits between frontend and other kernels (see [this illustration](https://vatlab.github.io/sos-docs/) for details). It allows the use of multiple kernels in one notebook (through [sos-notebook](https://github.com/vatlab/sos-notebook) for classic jupyter and [jupyterlab-sos](https://github.com/vatlab/jupyterlab-sos) for jupyterlab), and allows [data exchange  among live kernels](https://vatlab.github.io/sos-docs/doc/user_guide/exchange_variable.html).

The reason why `jupyterlab-lsp` does not work with SoS is simple: it does not know what language SoS is. If we are to solve this problem, there needs to be some way for SoS to notify `jupyterlab-lsp` the language used for each cell. I can work at both the frontend and backend (e.g. write a language server for SoS), but I am not sure if cell-level language support is at all possible with `jupyterlab-lsp`. I would appreciate any insight from the developers if and how this can be done. Thanks.

bollwyvl:
Yeah, the Language Server Protocol Specification doesn't say anything about multi-language documents, so we're kinda shooting in the dark here. Further, basically 0 language servers care about Jupyter's JSON format, or any special syntax kernel authors have added on top of their host language(s).

Our detection is currently based on existing Jupyter approaches like file extension sniffing, contents manager introspection, or in the case of notebooks, the kernel and or notebook metadata. If everything is just `sos`, we can't do much for you, nor do we offer many hooks into this system at this point.

Presently, we _do_ handle a small number of transclusions on the front-end, on a kernel-by-kernel basis, and it's rather deeply embedded inside the code. #191 discusses some approaches on how to normalize this, as a set of regular expressions + templates, or maybe some portable grammar and declarative transformation rules. If that's adopted, whether it's handled on the server side or the client side, there will be some hooks to extend it, ideally without having to rebuild the client (ha).

[...]

Issue #4, opened 2019-08-20
Ensure that we reconnect to the websocket connection to LSP if it is lost


krassowski:
A workaround, for now, is to simply refresh the JupyterLab in the browser (for example pressing F5).

PR #636, opened 2021-06-29
Tcp mode

FlyingSamson:
## References
I could not find any explicit issue that would be solved by this PR. However, the requirement for TCP support has been mentioned in #282 and #184 (The letter also contains some comments regarding design decisions)

## Code changes
- As discussed in #184 this PR firstly provides the switch from AsyncIO to AnyIO for the `LanguageServerSession` class and the `LspStreamBase`, `LspStreamReader` and `LspStreamWriter` classes (formerly `LspStdIoBase`, `LspStdIoReader` and `LspStdIoWriter`)
- Second it does provide the option to connect to TCP Language Servers (tested with [Wolfram-Language-Server](https://github.com/kenkangxgwe/lsp-wl/tree/v0.3.0), I'm happy to also provide my specification for it in another PR).
- As there are now different ways to connect to a LS, the specification given in `python_packages/jupyter_lsp/jupyter_lsp/schema/schema.json` was extended by a `mode` property which may take on values `stdio` or `tcp`, `stdio` being the default if no `mode` is specified, so as to not break existing language server specifications. If `mode==tcp` further properties `port` and `host` become relevant to specify the host and port where the language server is running and listening for queries. If `null` is passed as `host`, localhost is assumed for the connection, and if `null` is passed as `port` the server is launched and connected to at a random free port. Yet TODO: Test server running on remote host.

There should be no backward-incompatible changes as the default values for the newly introduced `mode` property is `stdio` and thus no changes for the (currently solely used) language servers communicating over stdio is required.

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


python_packages/jupyter_lsp/jupyter_lsp/connection.py
```diff
@@ -93,6 +96,9 @@ async def read(self) -> None:
                     self.wake()
 
                 IOLoop.current().add_callback(self.queue.put_nowait, message)
+            except (anyio.ClosedResourceError, anyio.EndOfStream):
+                # stream was closed -> terminate
+                break
```
bollwyvl, line 101: Perhaps a `self.log.debug` of something here, unless this is _entirely_ expected behavior... still working through the text.

python_packages/jupyter_lsp/jupyter_lsp/connection.py
```diff
@@ -171,32 +177,44 @@ async def read_one(self) -> Text:
...
+    async def _readline(self) -> Text:
         """Read a line (or immediately return None)"""
         try:
-            return self.stream.readline().decode("utf-8").strip()
-        except OSError:  # pragma: no cover
+            # use same max_bytes as is default for receive for now. It seems there is no way of getting
+            # the bytes read until max_bytes is reached, so we cannot iterate the receive_until call
+            # with smaller max_bytes values
+            async with anyio.move_on_after(0.2) as moa:
+                line = await self.stream.receive_until(b'\r\n', 65536)
```
bollwyvl, line 187: perhaps hoist magic number to constant

python_packages/jupyter_lsp/jupyter_lsp/connection.py
```diff
@@ -171,32 +177,44 @@ async def read_one(self) -> Text:
...
         """Write to a Language Server until it closes"""
-        while not self.stream.closed:
+        while True:
             message = await self.queue.get()
             try:
-                body = message.encode("utf-8")
-                response = "Content-Length: {}\r\n\r\n{}".format(len(body), message)
-                await convert_yielded(self._write_one(response.encode("utf-8")))
-            except Exception:  # pragma: no cover
+                nBytes = len(message.encode("utf-8"))
```
bollwyvl, line 208: prefer snake case `n_bytes`

python_packages/jupyter_lsp/jupyter_lsp/schema/schema.json
```diff
@@ -139,6 +139,36 @@
           "description": "list of MIME types supported by the language server",
           "title": "MIME Types"
         },
+	"mode": {
+	  "description": "connection mode used, e.g. stdio (default), tcp",
+	  "title": "Mode",
+	  "type": "string",
+	  "enum": ["stdio", "tcp"]
```
bollwyvl, line 146: perhaps add `"default": "stdio"`: won't _do_ anything, but is more explicit

python_packages/jupyter_lsp/jupyter_lsp/utils.py
```diff
@@ -0,0 +1,16 @@
...
+import socket
+
+
+def get_unused_port():
+    """Get an unused port by trying to listen to any random port.
+
+    Probably could introduce race conditions if inside a tight loop.
+    """
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.bind(('localhost', 0))
```
bollwyvl, line 12: we should probably make this (and wherever this came from `127.0.0.1` because docker :whale:)

bollwyvl:
Thank you so much for this! :heart_eyes: Despite all the :x:, this is a really good step forward.

I guess I'd like to see `Session` split into two implementations so there is just the one `if mode == "tcp"` check at the manager level. It is probably reasonable to either:

- have the (mostly) battle-tested one be the base implementation, and have the new TCP one subclass it
- make a new base class with a few easy-to-grok inner implementations in `TCPSession` and `StdioSession`

I wouldn't go too far overboard with `abc`'s etc.

I'm only a casual user of `anyio 2` (despite being tangentially to blame for its introduction into the jupyter stack), so i'll have to do a fair amount study to be able to help evaluate the specifics of this fully, but it does look like a good step forward. I'd _really_ love we if we could get rid of explicit executors, and/or replace our gross, cargo-culted non-blocking win32 code with anyio constructs, but that is a whole separate ball of wax. 

It looks like we will be able to test this fairly robustly with `python-lsp`, which supports `--tcp`, and indeed would need to crawl back up the coverage scale to feel good about it.

Giving it a pass through `jlpm lint` and `python scripts/lint.py` would also help clean things up a hair.

Solider on, and let us know if you get hard stuck anywhere!

FlyingSamson:
Thank you so much for taking the time to review my changes and your input! I hope I will find the time in the close future to fix everything.

Regarding anyio: I am not overly proud of the way I managed to get this working with anyio, especially the intersection where non async code (but still running within an existing event loop from jupyter lab) needs to call async code seems much to complicated for my taste. Even more so since the anyio change where `start_blocking_portal` now required a context manager which does not seem to play overly well with the object oriented world. So if you should find any better way to handle things, please let me know.

I will also try and split the Session class in two.

FlyingSamson:
While trying to get things running for externally running lsp-servers, I came across following problem (or so I think):

It seems that the current implementation never asks the server to shut down properly, i.e. no sequence `shutdown`&rarr; Server, `ack`&rarr;Client, `exit`&rarr; Server, seems to take place. I presume that was because each session was creating its own process running the required LS anyway, and after the work was done that process was just terminated. However, now that we would potentially like to support asking externally running servers for completions, we would have to close the connection more gracefully with accordance to the language server protocol, no?

So my question is: Am I right about this part of the protocol currently not being implemented? If so, I presume one would have to extend the `ClientRequest` and `ClientNotification ` enums in `connection.ts`. But I have no clue as to how I would proceed from there as I have zero knowledge of javascript programming. Do you have some pointers how I would go from there?

krassowski:
> So my question is: Am I right about this part of the protocol currently not being implemented?

Yes. I am not sure _when_ we want to ask servers to close, but at the minimum we need to do so at JupyterLab shutdown, which is also the safest option (but not trivial to detect). It is tricky because user closing the window does not mean that the session ends - they may be just switching to another browser or refreshing the window. So we would probably need to have a hook in the python code using `atexit`. Shutting the language server down when all documents for given language are closed is an option, but this requires investigation of how that would work for multi-user deployments.

Having said that, there is also a "shut down" option in JupyterLab's "File" menu, though I never use it (I always exit by terminating the server and closing the window) as I think many users do, hence the remark about the need of having `atexit` hook.

> If so, I presume one would have to extend the ClientRequest and ClientNotification enums in connection.ts

Yes, if we want to emit those from within the frontend (which as described above is complicated and could only reliably happen if and only if the "Shut down" button was pressed in JupyterLab, as any other server disconnect might indicate a temporary network failure, etc...). The next step would be to update `IClientNotifyParams` and `IClientRequestParams`, but again I am not convinced whether we want to have this handled by the frontend, or rather by the python extension instead.

We also need to investigate how this would interact with multi-user deployments such as JupyterHub and Real Time Collaboration. In short I do not think that we should implement this in this PR because there are too many unknowns; you are right that we should look into it and thank you for starting this discussion. I think we should create a new issue out of this to allow others to chime in with ideas before deciding on implementation.

FlyingSamson:
All right, thanks for clarifying. I will focus on implementing the TCP support for self launched processes for now then and open a separate issue to discuss on how to proceed from this stage, when this PR is done.

krassowski:
Picked up conflicts with schema. Sadly GitHub Actions do not kick in until conflicts are resolved. I also added a note about `stop_extension` hook in my previous comment.

FlyingSamson:
> Picked up conflicts with schema. Sadly GitHub Actions do not kick in until conflicts are resolved. I also added a note about `stop_extension` hook in my previous comment.

All right, so I can just pull master and merge it into the tcp_mode branch, thereby resolving the conflict, and then push again, correct?

bollwyvl:
> pull master and merge it into the tcp_mode branch,

or do it through the web ui if you're feeling brave...

python_packages/jupyter_lsp/jupyter_lsp/connection.py
FlyingSamson: By constant you mean a `traitlets.Int` member variable which is configurable, as in the last changes?

python_packages/jupyter_lsp/jupyter_lsp/connection.py
FlyingSamson: Done

python_packages/jupyter_lsp/jupyter_lsp/schema/schema.json
FlyingSamson: Done

python_packages/jupyter_lsp/jupyter_lsp/connection.py
FlyingSamson: Never mind my last comment. We should (and now do) stop any read or write operations by running them inside a cancelscope which can be canceled before closing the streams. So yes this part is not really expected behavior. I added according debug messages here and within the `write` function, as suggested.

python_packages/jupyter_lsp/jupyter_lsp/utils.py
FlyingSamson: Would you mind elaborating on why using `localhost` would/could cause trouble from within docker and how it would behave differently than `127.0.0.1`? I'm rather new to docker and therefore eager to learn such stuff.

bollwyvl:
Binder gives you a virtual network, and it acts a little funny. In
CI/binder/hub, they run even funnier, and as a user of such a service, one
has even less control of them. We've had to do a fair amount of upstream
work to ensure that test suites (e.g. pygls) can be run as-is... Usually
the localhost/127 is all that's needed, but sometimes hard coded ports also
need some techniques e.g. environment variables, random port discovery.

So yeah, outside of tests, anything that takes an address should at least
_accept_ overriding the host/port, and pick robust, loudly announced
default behavior if unspecified.

python_packages/jupyter_lsp/jupyter_lsp/connection.py
```diff
@@ -58,18 +57,34 @@ class LspStdIoReader(LspStdIoBase):
...
     def _default_max_wait(self):
         return 0.1 if os.name == "nt" else self.min_wait * 2
 
     async def sleep(self):
         """Simple exponential backoff for sleeping"""
-        if self.stream.closed:  # pragma: no cover
+        if self.stream._closed:  # pragma: no cover
             return
         self.next_wait = min(self.next_wait * 2, self.max_wait)
         try:
```
FlyingSamson, line 86: Do we really need that `try ... except ...` part here? As far as I can see, neither the previous `asyncio.sleep` nor the new `anyio.sleep` throw.

python_packages/jupyter_lsp/jupyter_lsp/connection.py
```diff
@@ -58,18 +57,34 @@ class LspStdIoReader(LspStdIoBase):
...
+        self.log.debug("%s closed", self)
 
     @default("max_wait")
     def _default_max_wait(self):
         return 0.1 if os.name == "nt" else self.min_wait * 2
 
     async def sleep(self):
         """Simple exponential backoff for sleeping"""
-        if self.stream.closed:  # pragma: no cover
+        if self.stream._closed:  # pragma: no cover
```
FlyingSamson, line 83: Is this side effect really necessary? I think we only ever call this method more or less immediately after a read operation was performed, at which time the stream was alive then. That is, hardly any time will have gone by between then and the time we entered this function which will as first action recheck that the stream is alive, which isn't even a concern in the remainder of that function.

Just asking because it seems rather complicated to produce future-proof code that will check wether the stream is alive. The used `_closed` property is undocumented and might go away in future versions of anyio. Another option would be a `receive_exactly(0)` within a `try IncompleteRead except ...`, however the border case of 0 bytes to read also seems risky, as `receive_exaclty` might or might not decide in future version whether 0 bytes can safely read by a closed steam (namely returning empty `bytes`-object) or not (namely throwing `IncompleteRead`

python_packages/jupyter_lsp/jupyter_lsp/connection.py
```diff
@@ -171,32 +188,60 @@ async def read_one(self) -> Text:
...
+            # way of getting the bytes read until max_bytes is reached, so we cannot
+            # iterate the receive_until call with smaller max_bytes values
+            async with anyio.move_on_after(0.2):
+                line = await self.stream.receive_until(b"\r\n", self.receive_max_bytes)
+                return line.decode("utf-8").strip()
+        except anyio.IncompleteRead:
+            # resource has been closed before the requested bytes could be retrieved
+            # -> signal recource closed
+            raise anyio.ClosedResourceError
+        except anyio.DelimiterNotFound:  # pragma: no cover
```
FlyingSamson, line 204: Regarding the next review round: I added `# pragma: no cover` in some places that either should never be called in the way the functions are used within the rest of the code, or seemed rather hard to test for. I tried to orient myself at places where this was already used, but was not sure whether there is any policy saying at which point I'm allowed to use this and at which point I'm not. So if you think any of those are inappropriate and think I should add a test for them just say a word (ideally with an idea on how to test for that part).

python_packages/jupyter_lsp/jupyter_lsp/session.py
```diff
@@ -75,48 +101,81 @@ def to_json(self):
...
+
+    async def run(self):
+        """run this session in a cancel scope and clean everything up on cancellation
+
+        the event `self.started` will be set when everything is set up and the session
+        will be ready for communication
+        """
+        async with CancelScope() as scope:
+            self.cancelscope = scope
+            await self.initialize()
```
FlyingSamson, line 132: At this point the old code did call `stop` (now `cleanup`). However I did not see why this would be necessary as at any one point where `start` (now `run`) is called the session can only be in `STOPPED` or `NOT_STARTED` state. If you think otherwise, please let me now.

Maybe in that case it would make sense to add something like
```
if self.status not in [SessionStatus.STOPPED, SessionStatus.NOT_STARTED]:  # pragma: no cover
    throw RuntimeError("A session for this language server is already running")
```

krassowski:
Thank you for good questions! I will aim to give some answers during the weekend.

I wanted to explain why I cancelled the CI runs: there was a new JupyterLab (3.1) release which required some changes to the Robot tests on CI and those were going to fail slowly taking CI time, so I needed to stop them. It is all fixed now so once you merge master or rebase the CI here should work fine.

python_packages/jupyter_lsp/jupyter_lsp/connection.py
krassowski: Agreed, it does not seem useful here.

python_packages/jupyter_lsp/jupyter_lsp/connection.py
krassowski: I am not sure why it is here. I imagine it could be useful if it was down where the sleep is called. It would be fine to remove it and wrap the entire `sleep()` call which is inside `_read_content()` in a try-except instead, just to be safe.

python_packages/jupyter_lsp/jupyter_lsp/connection.py
krassowski: That's absolutely fine!

krassowski (commented):
Thank you for this! Sorry it took me so long to get back. There seems to be one test failing on Windows: `python_packages/jupyter_lsp/jupyter_lsp/tests/test_session.py::test_stop[terminate immediately]` but otherwise the changes look good to me!

FlyingSamson:
I'm still struggling to get the stop procedure to work on windows. According to the anyio docs `Process.kill` is an alias of `Process.terminate` on Windows, which in turn calls the Win32 API function `TerminateProcess()`. However, the latter, opposed to the behavior of `Process.kill` on Posix systems, seems to be asynchronous, i.e., it will return immediately and does not ensure that the process is dead before returning. Does somebody know how I could achieve the same behavior on Windows? 

If that indeed is not possible, how should I handle this on Windows?   
Should I just raise a warning in `session.stop_process` if the OS is Windows instead of the call to `Process.kill` on Posix systems?  
How am I supposed to test this? The current test works on Posix, so I presume that is fine. Should I just give Windows some extra time (not sure how much more time Windows will require, as I don't have a Windows machine to test this locally, maybe 10 seconds for a start) and check afterwards if the process has gone away? This would than at least test that the process will die at some point, if not at the time that we force it to on other systems.    
If indeed it takes that long for Windows to shut down the process, should the `timeout` parameter of `session.stop_process` be OS dependent, to keep it short on Posix systems.

Sorry for the myriad of questions. I'm thankful for any advice :)

krassowski:
> I'm still struggling to get the stop procedure to work on windows.

If no one has the ability to test/support Windows properly we can merge the best solutions you find appropriate and release this saying that Windows support is provisional and "help wanted".

> Should I just give Windows some extra time (not sure how much more time Windows will require, as I don't have a Windows machine to test this locally, maybe 10 seconds for a start) and check afterwards if the process has gone away?

Sounds good to me.

>  If indeed it takes that long for Windows to shut down the process, should the timeout parameter of session.stop_process be OS dependent, to keep it short on Posix systems.

yes

FlyingSamson:
I think the code should be ready for another review round (I don't think that the smoke installation error was caused by the made changes).

To fully take advantage of the structured concurrency offered by anyio, I think more work needs to be done from the jupyterlab extension API side, as the current implementation using Tornado Websockets, while supporting coroutines for `open()` and `on_message()`, does not seem to be made for structured concurrency. E.g., the doc of `open()` states that no messages will be processed until `open()` has returned, thus preventing any structured concurrency happening within without using another event loop in a separate thread.

python_packages/jupyter_lsp/jupyter_lsp/session.py
```diff
@@ -48,7 +81,14 @@ class LanguageServerSession(LoggingConfigurable):
     last_handler_message_at = Instance(datetime, allow_none=True)
     last_server_message_at = Instance(datetime, allow_none=True)
 
-    _tasks = None
+    stop_timeout = Float(
+        5,
+        help="timeout after which a process will be terminated forcefully",
```
bollwyvl, line 86: perhaps include units? in the description, or the name?

python_packages/jupyter_lsp/jupyter_lsp/session.py
```diff
@@ -48,7 +81,14 @@ class LanguageServerSession(LoggingConfigurable):
...
+    ).tag(config=True)
+    queue_size = Float(
+        math.inf,
```
bollwyvl, line 89: what about `-1` instead? `inf` is hard to encode in JSON

krassowski:
Looking at the issue of freezes everywhere except for Ubuntu py3.10 on; there is no sane way to debug it on CI/with pytest, but locally I can reproduce it. Starting lab I see:

```python
Exception while listening Cannot add child handler, the child watcher does not have a loop attached
      File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 162, in run
        await self.initialize()
      File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 180, in initialize
        await self.init_process()
      File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 328, in init_process
        await self.start_process(self.spec["argv"])
      File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 230, in start_process
        env=self.substitute_env(self.spec.get("env", {}), os.environ),
      File ".pyenv/versions/3.7.15/envs/lsp-lab3.x-py3.7/lib/python3.7/site-packages/anyio/_core/_subprocesses.py", line 135, in open_process
        start_new_session=start_new_session,
      File ".pyenv/versions/3.7.15/envs/lsp-lab3.x-py3.7/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 1112, in open_process
        start_new_session=start_new_session,
      File ".pyenv/versions/3.7.15/lib/python3.7/asyncio/subprocess.py", line 217, in create_subprocess_exec
        stderr=stderr, **kwds)
      File ".pyenv/versions/3.7.15/lib/python3.7/asyncio/base_events.py", line 1544, in subprocess_exec
        bufsize, **kwargs)
      File ".pyenv/versions/3.7.15/lib/python3.7/asyncio/unix_events.py", line 193, in _make_subprocess_transport
        self._child_watcher_callback, transp)
      File ".pyenv/versions/3.7.15/lib/python3.7/asyncio/unix_events.py", line 941, in add_child_handler
        "Cannot add child handler, "
    RuntimeError: Cannot add child handler, the child watcher does not have a loop attached
```

Coming from `self.process = await anyio.open_process()` call.

krassowski:
This appears to be https://bugs.python.org/issue35621 which was only fixed in Python 3.8 as the workaround from https://bugs.python.org/msg370355:

```python
import asyncio
from .threaded_child_watcher import ThreadedChildWatcher   # where ThreadedChildWatcher was copied from CPython
asyncio.set_child_watcher(ThreadedChildWatcher())
```

does work. This is for Unix only, no idea how to fix this on Windows, and the workaround may have side-effects on tornado.

krassowski:
I guess the question becomes: do we really need/want `start_process` to by `async`?

krassowski:
So things are getting interesting with https://github.com/krassowski/jupyterlab-lsp/pull/4/commits/12bfbd42b6085f290affd1f6375049d5bc27a229:
- Windows py3.10 works on CI (which gives me hope) but throws `ValueError: I/O operation on closed pipe` from time to time in tests which is [a known problem](https://stackoverflow.com/questions/63905578/valueerrori-o-operation-on-closed-pipe-when-using-asyncio-with-subprocess) and apparently can be worked around by calling private `process._transport.close()`. Probably worth finding a bpo or filing an issue with cPython.
- Ubuntu py3.7 starts up locally but hangs when running tests (both locally and on CI). When manually aborting the tests we can see that it raises:

```python
  File "/3.7.15/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/3.7.15/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/lsp-lab3.x-py3.7/lib/python3.7/site-packages/anyio/_core/_eventloop.py", line 70, in run
    return asynclib.run(func, *args, **backend_options)
  File "/lsp-lab3.x-py3.7/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 292, in run
    return native_run(wrapper(), debug=debug)
  File "/3.7.15/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/3.7.15/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "/lsp-lab3.x-py3.7/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 287, in wrapper
    return await func(*args)
  File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 179, in run
    await self.cleanup()
  File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 205, in cleanup
    await self.stop_process(self.stop_timeout)
  File "jupyterlab-lsp/python_packages/jupyter_lsp/jupyter_lsp/session.py", line 254, in stop_process
    self.process.terminate()
  File "/lsp-lab3.x-py3.7/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 1053, in terminate
    self._process.terminate()
  File "/3.7.15/lib/python3.7/asyncio/subprocess.py", line 131, in terminate
    self._transport.terminate()
  File "/3.7.15/lib/python3.7/asyncio/base_subprocess.py", line 150, in terminate
    self._check_proc()
  File "/3.7.15/lib/python3.7/asyncio/base_subprocess.py", line 143, in _check_proc
    raise ProcessLookupError()
ProcessLookupError
```

PR #658, opened 2021-08-09
Handle case where position is null in document.ts

bsyouness:
## References

This fixes issue #543 which #544 fixed partially.

## Code changes

Add a check that `pos` isn't `null` before trying to access its properties.

## Chores

- [x] linted 
- [x] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ __pycache__/
 *.so
 
 # Distribution / packaging
+python_packages
 .Python
 build/
 develop-eggs/
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ __pycache__/
 *.so
 
 # Distribution / packaging
-python_packages
 .Python
 build/
 develop-eggs/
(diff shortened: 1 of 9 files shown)


krassowski:
Thank you! The tests are failing - we do not have a `console` on object on `VirtualDocument`. Using plain console would be fine here for now, but if you want to add our custom console wrapper as a protected member of `VirtualDocument` (should be passed via `VirtualDocument.IOptions`) please feel free to do so.

bsyouness:
@krassowski I did as per your suggestion and added a custom console as a member of `VirtualDocument`.

.gitignore
```diff
@@ -7,6 +7,7 @@ __pycache__/
 *.so
 
 # Distribution / packaging
+python_packages
```
krassowski, line 10: There are quite important files in there ;) Maybe we could ignore `build` directory within but not the source code.

```suggestion
```

bsyouness:
Should I be concerned about/fix the partial test failures?

krassowski:
No, no worries I will take care of those.

krassowski:
I opened an issue about this upstream https://github.com/conda-forge/tzlocal-feedstock/issues/17.

bsyouness:
Thank you!

krassowski:
Thank you!

PR #671, opened 2021-08-30
Make signature persist, highlight active parameter, show on top

### LINKED ISSUE #619: How to configure so that only one signature suggestion shows
It's fine to automatically give a signature suggestions when I type a function.

However, in my case, two signature suggestion dialogs are displayed as follows:

signature

Actually, the bottom one shows when I type the function name `sum`, while the top one shows when I finished typing the parentheses. If I type slowly, the two dialogs show in turn; if i type fast, then the two dialogs shows simultaneously, which really bothers my typing.

I install `jupyterlab-lsp` according to the installation instruction in the readme.md, choosing `pylsp` as the language server.

My question is how to configure so that only one signagure suggestion is displayed. Thanks!

krassowski:
## References

Fixes #619.

## Code changes

- Updated CodeMirror types to include full signature of `runMode()`
- properly handle various different ways the signature details can be presented by the server
- make the tooltip optionally persistent and make it persistent for signatures

## User-facing changes

- The signature shows the active parameter if server returns such information (pyls/pylsp does, R-language server does not)
- The signature is displayed above the current line
- The signature is persistent and will be only hidden if:
   - [x] server tells us to do so by returning `null` (which has different meaning from `undefined`!)
   - [x] user presses ESC
   - [x] user types pre-defined closing characters, currently `)` and `;`
        - [ ] which is user-customizable
   - [ ] user moves the cursor prior to the range of the signature
   - [ ] user moves to a different cell

Screenshot from 2021-08-30 16-19-34

None

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


krassowski:
@chenyulue would you like to test this one on the binder and provide feedback on the new signature behaviour?

[Binder](https://mybinder.org/v2/gh/krassowski/jupyterlab-lsp/improve-signature?urlpath=lab)

PR #711, opened 2021-11-28
Add per-language binder demo links

krassowski:
This will allow for faster start-up for anyone who wants to try it out. Ideally the binder repos would get a tour and an example notebook catered to each language which are good "first issues".


krassowski:
```
Handles Server Extension Failure                                      | FAIL |
Parent suite setup failed:
Keyword 'Wait For Splash' failed after retrying 3 times. The last error was: WebDriverException: Message: Reached error page: about:neterror?e=connectionFailure&u=http%3A//localhost%3A63263/%40est/lab%3Freset%26token%3DEPEXw7oF&c=UTF-8&d=Firefox%20can%E2%80%99t%20establish%20a%20connection%20to%20the%20server%20at%20localhost%3A63263.
```

seems to be failing

Issue #669, opened 2021-08-26
Jump to definition: No jump targets found


jabbera:
## Description

I don't know when this started, but I can't get jump to definition working anymore. Whenever I use it I get: No Jump Targets Found

## Reproduce

![image](https://user-images.githubusercontent.com/301208/130882182-ae902b93-2a80-41bf-9890-193a460543e0.png)

## Expected behavior

It would jump to definition.

## Context

It's not clear to me if it's the same issue as: https://github.com/krassowski/jupyterlab-lsp/issues/466 

I do use / as my "root folder".

- Operating System and version: docker-stacks:minimal-notebook:latest (with some customization)
- Browser and version: Chrome 92.0.4615.159
- JupyterLab version: 3.1.7

<details><summary>Required: installed server extensions</summary>
<pre>
Config option `kernel_spec_manager_class` not recognized by `ListServerExtensionsApp`.
Config dir: /mnt/z/research_hub/.jupyter

/opt/conda/lib/python3.9/site-packages/jupyter_server/transutils.py:13: FutureWarning: The alias `_()` will be deprecated. Use `_i18n()` instead.
  warnings.warn(warn_msg, FutureWarning)
/opt/conda/lib/python3.9/site-packages/jupyter_server/transutils.py:13: FutureWarning: The alias `_()` will be deprecated. Use `_i18n()` instead.
  warnings.warn(warn_msg, FutureWarning)
Config dir: /opt/conda/etc/jupyter
    gmo_jupyter_share_link enabled
    - Validating gmo_jupyter_share_link...
      gmo_jupyter_share_link 0.3.6 OK
    jupyter_lsp enabled
    - Validating jupyter_lsp...
      jupyter_lsp 1.4.1 OK
    jupyter_server_proxy enabled
    - Validating jupyter_server_proxy...
      jupyter_server_proxy  OK
    jupyter_resource_usage enabled
    - Validating jupyter_resource_usage...
      jupyter_resource_usage  OK
    jupyter_server_mathjax enabled
    - Validating jupyter_server_mathjax...
      jupyter_server_mathjax  OK
    jupyterlab enabled
    - Validating jupyterlab...
      jupyterlab 3.1.7 OK
    jupyterlab_git enabled
    - Validating jupyterlab_git...
      jupyterlab_git 0.32.2 OK
    jupyterlab_spellchecker enabled
    - Validating jupyterlab_spellchecker...
[...]

PR #712, opened 2021-11-28
fix renaming of file when jumping code

jepcor97:
This PR works around the issue #669 as described [here](https://github.com/jupyter-lsp/jupyterlab-lsp/issues/669#issuecomment-981158541), by just removing the renaming of the tab, until there is a way to do it without actually renaming the files and breaking the envrioment:

## References
Solves issue #669

## Code changes
```
$ git diff fix_jumping..upstream/master
diff --git a/packages/code-jumpers/src/jumpers/jumper.ts b/packages/code-jumpers/src/jumpers/jumper.ts
index 4237c7d..714af70 100644
--- a/packages/code-jumpers/src/jumpers/jumper.ts
+++ b/packages/code-jumpers/src/jumpers/jumper.ts
@@ -91,7 +91,7 @@ export abstract class CodeJumper {
 
   private protectFromAccidentalEditing(document_widget: IDocumentWidget) {
     let editor_widget = document_widget as IDocumentWidget<FileEditor>;
-    editor_widget.title.label = editor_widget.title.label + '';
+    editor_widget.title.label = editor_widget.title.label + ' (external)';
     let editor = editor_widget.content.editor;
     let disposable = editor.addKeydownHandler(
       (editor: IEditor, event: KeyboardEvent) => {
```

### DIFF
--- a/packages/code-jumpers/src/jumpers/jumper.ts
+++ b/packages/code-jumpers/src/jumpers/jumper.ts
@@ -91,7 +91,7 @@ export abstract class CodeJumper {
 
   private protectFromAccidentalEditing(document_widget: IDocumentWidget) {
     let editor_widget = document_widget as IDocumentWidget<FileEditor>;
-    editor_widget.title.label = editor_widget.title.label + ' (external)';
+    editor_widget.title.label = editor_widget.title.label + '';
     let editor = editor_widget.content.editor;
     let disposable = editor.addKeydownHandler(
       (editor: IEditor, event: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


packages/code-jumpers/src/jumpers/jumper.ts
```diff
@@ -91,7 +91,7 @@ export abstract class CodeJumper {
 
   private protectFromAccidentalEditing(document_widget: IDocumentWidget) {
     let editor_widget = document_widget as IDocumentWidget<FileEditor>;
-    editor_widget.title.label = editor_widget.title.label + ' (external)';
+    editor_widget.title.label = editor_widget.title.label + '';
```
krassowski, line 94: I would suggest to replace it with a comment altogether:

```suggestion
// We used to adjust `editor_widget.title.label` here but an upstream
// bug (https://github.com/jupyterlab/jupyterlab/issues/10856) prevents
// us from doing so anymore.
```

jepcor97: Yes, better that way

PR #713, opened 2021-11-29
Fix server extension UI test: migrate config to `ServerApp` as needed

krassowski:
An attempt to fix issues on CI seen when configuring/reconnecting to JupyterServer https://github.com/jupyter-lsp/jupyterlab-lsp/pull/711#issuecomment-981161263

I cannot reproduce it locally.


krassowski:
It looks like we are back on green (for the most part).

PR #719, opened 2021-12-11
Prepare jupyterlab-lsp 3.9.2 and jupyter-lsp 1.5.1 release

krassowski:
- We don't have jupyter-releaser integration in just yet, but it is still acceptable to watch the progress of this PR's GitHub Action listening to The Final Countdown (https://github.com/jupyter-server/jupyter_releaser/issues/202).


krassowski:
It looks like Windows runners have problems with uploading artifacts (error 400) and MacOS even after pinning is using Julia 1.7 (why?)...

PR #723, opened 2021-12-12
Update Pythons in CI

### LINKED ISSUE #718: Drop Python 3.6, start testing a newer version, restore MacOS to latest
Python 3.6 is nearing EOL, and PyPy 3.6 got excluded from MacOS 11 on GitHub Actions. We should drop testing on 3.6 as it starts to cause issues. We should also restore `macos-latest` in the matrix (currently pinned to 10.15).

Cross-referencing:
- https://github.com/jupyter-lsp/jupyterlab-lsp/pull/717#issuecomment-991717395
- https://github.com/jupyterlab/jupyterlab/issues/11645
- https://github.com/python-lsp/python-lsp-server/pull/130

Edit: it looks that PyPy smoke is now also failing on Windows.
Python 3.6 EOL: 23 Dec 2021

bollwyvl:
[Binder](https://mybinder.org/v2/gh/bollwyvl/jupyterlab-lsp/gh-718-ci-pythons)

## References

- fixes #718
- prerequisite to rebooting #493

## Code changes

- deps: updates `python_requires` to 3.7
- CI: normalizes CPython versions under test to 3.7 and 3.10
  - ~~might need to roll back top to 3.9 for solveability~~
  - add  `r` 4.* and 3.* to matrix
  - share `node_modules` cache between `build` and `lint`
- docs: updates readthedocs to use new environment configuration
  - uses docker image pre-loaded with mambaforge

## User-facing changes

- n/a

- users on python 3.6 will be unable to upgrade to future versions

## Chores

- [x] linted
- [x] tested
- [ ] documented
- [ ] changelog entry


bollwyvl:
Looks like `3.10` is a no-go for.... [reasons](https://github.com/jupyter-lsp/jupyterlab-lsp/runs/4498395479?check_suite_focus=true#step:5:624):

```
package r-recommended-3.3.2-r3.3.2_0 requires r-base 3.3.2, but none of the providers can be installed
```

krassowski:
Is probably safe to witch R to 4.0 maybe even 4.1. Would that help?

bollwyvl:
>  R to 4.0 maybe even 4.1

Sure' maybe worth trying...

bollwyvl:
All of the acceptance tests have run to completion at least once, so I don't see why this wouldn't work, now. Still checking out binder, but otherwise fairly happy with the changes.

krassowski (approved):
LGTM, thank you!

bollwyvl:
Whoops, on master CI:
```
 Invalid workflow file

The workflow is not valid. .github/workflows/job.test.yml (Line: 16, Col: 10): Unexpected value ''
```

krassowski:
Huh? https://github.community/t/breaking-change-to-concurrency-group-syntax/215604

bollwyvl:
Hotfixed on e83d8e6, but still need to see if that's the right thing to have done. Basically, want to only be running one set of jobs per PR...

krassowski:
So that issue I linked to was resolved: https://github.com/actions/runner/issues/1532

PR #724, opened 2021-12-12
Add Robot Framework language server (again)

### LINKED ISSUE #332: Support/test a Robot Framework language server
## Elevator Pitch

Adding support for autodetecting a robot framework language server

## Motivation

- [robotframework-lsp](https://pypi.org/project/robotframework-lsp) has shipped a couple solid versions, and has a product built around it
  - the [conda-forge](https://github.com/conda-forge/staged-recipes/pull/12548) process has been started
- we have lots of examples, and this would increase our ability to dogfood diverse languages in building this code

## Design Ideas

- add to environment
- add spec
- add tests, etc.
- add robotframemework [syntax mode](https://github.com/robots-from-jupyter/robotkernel/tree/master/src/jupyterlab_robotmode)
  - it's part of `robotkernel`, which we _could_ choose to add to the development environment
  - i'm pretty much the original author of that syntax mode, and owe an update for some of the new syntax  (some for loop stuff isn't handled correctly), but it basically works, today

bollwyvl:
## References

- fixes #332
- supersedes #493

## Code changes

- refreshes robot dependencies
- adds new robot stuff
  - robotkernel
  - robotframework-lsp
  - robotframework-robocop (not configured yet)
  - robotframework-pabot (interested to see the intersection of this with other tools)
  - jupyterlab-robotmode

## User-facing changes

- vistors to the binder will see more robot-related features

- n/a

## Chores

- [ ] linted
- [ ] tested
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 ## Changelog
 
+### `jupyter-lsp 1.6.0` (unreleased)
+
+- compatibility:
+  - raises the minimum supported Python to 3.7 as 3.6 reaches its end-of-life ([#723])
+- maintenance and upkeep:
+  - adds Python 3.10, PyPy 3.7, and R 4.x to the automated tests ([#723])
+
+[#768]: https://github.com/jupyter-lsp/jupyterlab-lsp/pull/723
+
 ### `@krassowski/jupyterlab-lsp 3.9.2` (2021-12-12)
 
 - bug fixes:
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -33,3 +33,8 @@ dependencies:
   - tectonic
   - texlab
   - chktex
+  # robotframework
+  - jupyterlab_robotmode
+  - robotframework-lsp
+  - robotframework-robocop
+  - robotkernel
--- a/docs/rtd.yml
+++ b/docs/rtd.yml
@@ -16,6 +16,7 @@ dependencies:
   - python >=3.7,<3.11.0a0
   - python-graphviz
   - python-lsp-server
+  - robotframework-lsp
   - sphinx
   - sphinx-autodoc-typehints
   - sphinx-book-theme
--- a/requirements/atest.txt
+++ b/requirements/atest.txt
@@ -2,4 +2,7 @@
 -r ./lab.txt
 bs4
 robotframework >=4
+robotframework-lsp
+robotframework-pabot
 robotframework-seleniumlibrary
+robotkernel
--- a/requirements/atest.yml
+++ b/requirements/atest.yml
@@ -5,7 +5,10 @@ channels:
   - nodefaults
 
 dependencies:
-  - robotframework-seleniumlibrary
-  - robotframework >=4
   - firefox
   - geckodriver
+  - jupyterlab_robotmode
+  - robotframework >=4
+  - robotframework-pabot
+  - robotframework-seleniumlibrary
+  - robotkernel
--- a/requirements/github-actions.yml
+++ b/requirements/github-actions.yml
@@ -10,9 +10,15 @@ dependencies:
   - pip
   - nodejs {nodejs}
   # for python language server (and development)
+  - python-lsp-server
   - flake8 >=3.5
-  # temporarily pin autopep8
   - autopep8 <1.6.0
+  # robotframework for testing and language server
+  - jupyterlab_robotmode
+  - robotframework >=4
+  - robotframework-lsp
+  - robotframework-robocop
+  - robotkernel
   # for R language server and kernel
   - r {r}
   - r-irkernel
@@ -34,6 +40,5 @@ dependencies:
   - bs4
   - firefox
   - geckodriver
-  - robotframework >=4
+  - robotframework-pabot
   - robotframework-seleniumlibrary
-  - python-lsp-server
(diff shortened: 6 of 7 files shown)


bollwyvl:
Still need to test in `Configuration`: once `robocop` is turned on, underlines don't show up, even though things appear in the diagnostic panel...

bollwyvl:
> underlines don't show up

Ah, all of the diagnostics I've seen have zero width! Really surprised me!

I guess some options (not to be fixed on this issue):
- gutter (we've wanted that anyway, was hoping to wait for CM6)
- offer expanding zero-width diagnostics to the full line

krassowski:
> offer expanding zero-width diagnostics to the full line

or to 1 character? I would like this regardless of gutters.

bollwyvl:
Sure, perhaps expandZeroWidthDiagnostics with a default of 0, preserving
existing behavior, while 1 would be one character, and -1 could go to the
end of the line.

krassowski:
If that becomes a setting, I happy for it to have a non-zero default.

krassowski:
Is it still WIP, or can we merge?

bollwyvl:
I'll merge upstream and take another look, but probably good to go, no
substantive further changes planned...

bollwyvl:
I've
- removed the WIP 
- hoisted the diagnostic idea over to https://github.com/jupyter-lsp/jupyterlab-lsp/issues/743

If this passes, it's probably ready to go.

binder/postBuild
```diff
@@ -6,18 +6,23 @@ jlpm bootstrap
...
+    jupyter labextension develop . --overwrite
 popd
+
 # List extensions
 jupyter serverextension list
 jupyter server extension list
 jupyter labextension list
+
+if [ "$NB_USER" = "jovyan" ]; then
+    cp binder/overrides.json $NB_PYTHON_PREFIX/share/jupyter/lab/settings/
```
krassowski, line 27: we need to mkdir here:

```
+ cp binder/overrides.json /srv/conda/envs/notebook/share/jupyter/lab/settings/
cp: cannot create regular file '/srv/conda/envs/notebook/share/jupyter/lab/settings/': Not a directory
```

binder/overrides.json
```diff
@@ -0,0 +1,21 @@
+{
+  "@krassowski/jupyterlab-lsp/plugin": {
```
krassowski, line 2: probably needs `:`
```suggestion
  "@krassowski/jupyterlab-lsp:plugin": {
```

binder/postBuild
krassowski, line 27:
```suggestion
    mkdir -p $NB_PYTHON_PREFIX/share/jupyter/lab/settings/
    cp binder/overrides.json $NB_PYTHON_PREFIX/share/jupyter/lab/settings/
```

bollwyvl:
Thanks for the binder assist, trying right now. I put the user check in because i test (the rest of) `postBuild` as part of my local development flow, but having too much fancy stuff configured makes it harder to test... are there any other things we'd want to pre-configure, i guess? Maybe `checktex` or whatever cuteness it is?

bollwyvl:
Ugh, the "magic" `rstudio` stuff on binder is brutal. It looks like our initial solve pulls in 4.1, but then the hard pins bump it back down to 4.0, and then `pip install`s a bunch of weird stuff for `shiny` from github tarballs. Gross. I'll have to defer to `r` users whether this is all worth it, or if we should remove the `r` stuff from the "main" binder environment, and install just our `r`-related stuff (without rstudio and shiny server i guess)... conda-forge has made overtures of collaboration with the folk that run that ecosystem and they pretty much are defending their distribution mechanism.

Additionally: it looks like _something_ we're pulling in is forcing us down to jupyterlab 3.1.x, which is good for testing, i guess, but poor for picking up more recent features and bug fixes. It also means that the retrolab we're getting is fairly outdated at this point... I guess that's academic, as we'll want to start testing with `lab4` and `nb7` soon enough on another branch, and should keep what we've got working today working as long as we can.

binder/overrides.json
```diff
@@ -1,5 +1,5 @@
 {
-  "@krassowski/jupyterlab-lsp/plugin": {
+  "@krassowski/jupyterlab-lsp:plugin": {
```
bollwyvl, line 2: good catch

bollwyvl:
Ok, had a go at it on binder. Everything seems _mostly_ working except:
- third-party keyword libraries (e.g. `SeleniumLibrary`) doesn't seem to resolve, which _doesn't_ happen locally, and potentially something we can investigate at a later date
- completion (both from the kernel and the lsp) seems to be busted

Note that i can't reproduce these issues locally. We could put completion under test, but once again, at some point there's only so much we can control.

conda-lock for reference: [jupyter-lsp-binder.conda.lock.txt](https://github.com/jupyter-lsp/jupyterlab-lsp/files/7799264/jupyter-lsp-binder.conda.lock.txt)

krassowski:
> completion (both from the kernel and the lsp) seems to be busted

works for me on binder; the problem might be that it is case sensitive by default; we might want to allow servers specs to override the defaults (and have a default option for `caseSensitive` set to "auto" which means let server decide). Similarly for SQL we would want case-insensitive completion.

krassowski (approved):
LGTM, thanks!

PR #726, opened 2021-12-17
Fix identifier for pluralized strings

fcollonval:
## References

Follow-up of https://github.com/jupyter-lsp/jupyterlab-lsp/pull/700 due to bug encounter when generating the POT file (see that [issue](https://github.com/jupyterlab/language-packs/issues/198))

## Code changes

Add fake context for string converted to plural but without changing their id to create a new unique distinct reference for them.

## User-facing changes

N/A

N/A

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


krassowski (approved):
Thanks!

PR #727, opened 2021-12-19
Prepare jupyterlab-lsp 3.9.3 release

krassowski:
Minor patch for https://github.com/jupyterlab/language-packs/issues/198 and dev/testing dependencies refresh.


bollwyvl:
@krassowski sorry, was noticing this on my phone while running around, but didn't have a time to chime in: the bump of the minimum python did land on both `jupyterlab-lsp` (which got a minor bump) and on `jupyter-lsp` which didn't even get that .

I guess on `conda-forge`, I guess I'll just not bump the `jupyter-lsp` build number...

krassowski:
Good catch; I updated release notes to mention this. I mean EOL is this week. `jupyter-lsp` artifacts were not uploaded (cannot upload new one with the same version) so no need to worry about it I think. `jupyterlab-lsp` indeed will complain on Python 3.6 if someone tries to install 3.9.3.

bollwyvl:
Yeah, no harm no foul, i think.

PR #730, opened 2021-12-21
Request language server info in bug report template

krassowski:
Request language server info in bug report template


bollwyvl:
This is a good step, for sure... i feel like some future version of our language server spec should include a way to find that with a user-oriented CLI, e.g. `jupyter lsp versions`, etc. Unfortunately, not all language servers support such a thing from their command line interface, and I don't believe there is a message we could use to determine it from a running one...

PR #731, opened 2021-12-25
Check for `node_modules` in `~/.local/lib` or other user-defined prefix

krassowski:
## References

Supports global installations into local directory. I would not recommend this for most servers, but I think it is a valid use case.

## Code changes

Adds an extra default `node_modules` root.

## User-facing changes

Servers installed globally into `~/.local/` (i.e. using `npm install -g server-name` after `npm config set prefix '~/.local/'`) will be detected automatically.

None

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


bollwyvl:
Hm... Can the (as-configured) value be retrieved from the
(as-found-on-PATH) npm cli? Seems like that might be more robust as a
last-ditch effort, rather the prescribing some fixed value. My concern
being that this is a highly order-dependent set of commands with
significant side effects, which, as we have seen, may well not be done in
the order we'd expect.

krassowski:
Yes. We can use `npm prefix -g` to get the current prefix and append "lib" to that.

krassowski:
It looks that Windows failure could be relevant.

python_packages/jupyter_lsp/jupyter_lsp/types.py
```diff
@@ -257,6 +258,15 @@ def _default_node_roots(self):
         # ... but right in %PREFIX% on nt
         roots += [pathlib.Path(sys.prefix)]
 
+        # check for custom npm prefix
+        if shutil.which("npm"):
+            prefix = (
+                subprocess.run(["npm", "prefix", "-g"], check=True, capture_output=True)
```
bollwyvl, line 264: I don't think we want this to fail... perhaps more pithy/defensive might be to hoist this to a new function which can contain a `try` and cache the output... especially on windows, the startup time of a `nodejs` process is noticeable.

PR #732, opened 2021-12-25
Squash warnings and errors in web console

krassowski:
## References

As highlighted in few issues, squashes some warnings:
- `TypeError: Cannot read properties of null (reading 'uri')` - this warrants a further investigation, looks like we have a regression and do dispose some connections properly so diagnostics keep coming; still this highlights a logic 
- `LSP: is_visible failed TypeError: Cannot read properties of undefined (reading 'virtual_line')` - context is sometimes requested when the document is not open yet, or already closed, we should handle it
- `TypeError: n is not a function` - this was an error in handling of a default arguments.

TODO:
- [ ] let's try to enable `strictNullChecks` #130 - this will add a lot of optional chaining but should force us to think about the values being possibly disposed due to the asynchronous nature the communication.

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/adapters/adapter.ts
+++ b/packages/jupyterlab-lsp/src/adapters/adapter.ts
@@ -722,7 +722,11 @@ export abstract class WidgetAdapter<T extends IDocumentWidget> {
 
   abstract context_from_active_document(): ICommandContext | null;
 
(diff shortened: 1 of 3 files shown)


krassowski:
I will enable `strictNullChecks` in subsequent PR - and that might break some interfaces.

PR #733, opened 2021-12-26
Enable strict null checks and other strict settings

### LINKED ISSUE #130: Enable strictNullChecks

krassowski:
## References

Closes #130

## Code changes

`strictNullChecks` are now enabled for all packages; a number of bugs were caught thanks to this change and fixed:
- while logger extension was not required, if it was implicitly assumed thus logging all communication was broken if the logger extension was absent; this could have affected advanced developers debugging LSP in RetroLab/Notebook v7
- `window/showMessageRequest` server request was assuming that actions are always provided by the server which could have resulted in messages not showing up; now a "Dismiss" button will be added a a default action when no actions are provided by the server
- changing kernels always led to restarting of LSP connection, even if the kernel language did not change; now the connection will be retained if possible
- `markdownRenderer` is no longer implicitly required
- some potential failures of completion in edge cases (when `token.type` is missing, when line cannot be retrieved from editor) were fixed.
- diagnostics sorting with missing values for `source` and `severity` was improved and missing values will be consistently shown at the end
- diagnostics placeholder was split into `Diagnostics are not available` and `No issues detected, great job!` which will now show up properly.

In addition `noImplicitThis` ,`strictBindCallApply`, `noImplicitReturns`, `noFallthroughCasesInSwitch` are now enabled, and `allowUnusedLabels` was changed from default `undefined` (=warn) to stricter `false`.

## User-facing changes

More console warning were added for edge cases which might have been eluded our attention previously.

Public APIs are now annotated with `undefiend` and `null` where needed; this is a non-breaking change for dependants using default `strictNullChecks: false` and might be a breaking change for anyone else with `strictNullChecks` enabled.

- `@krassowski/code-jumpers`:
   - `ILocalPosition.index` is now always required, but it is effectively ignored outside of the notebook

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/packages/_example-extractor/src/api.spec.ts
+++ b/packages/_example-extractor/src/api.spec.ts
@@ -37,7 +37,7 @@ describe('The foo extractor', () => {
   test.each(Object.entries(FIXTURES))(
     '%s',
     (_: string, expected: IExtractedCode) => {
-      const extracted = extractor.extract_foreign_code(expected.host_code);
+      const extracted = extractor.extract_foreign_code(expected.host_code!);
       expect(extracted).to.have.length(1);
       expect(extracted[0]).to.deep.equal(expected);
     }
--- a/packages/code-jumpers/src/history.ts
+++ b/packages/code-jumpers/src/history.ts
@@ -30,7 +30,7 @@ export class JumpHistory {
     this.jump_history.push(JSON.stringify(position));
   }
 
-  recollect(): IGlobalPosition {
+  recollect(): IGlobalPosition | undefined {
     this.ensure_history_is_ready();
     if (this.jump_history.length === 0) {
       return;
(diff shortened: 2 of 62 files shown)


bollwyvl:
This is a great step to take. There are a number of additional, even more aggressive checks that can be enabled, if we're willing to deal with them, especially on newer versions of typescript...

krassowski:
I added `noImplicitThis` and `strictBindCallApply`. `useUnknownInCatchVariables` would be nice once we are on newer TypeScript. I will also give `noImplicitAny` a shot, but this would be a substantial changeset - not sure if I will do it this time.

`strictFunctionTypes` and `strictPropertyInitialization` conflict with how the code is structured and will cost more effort to refactor, so definitely not this time.

Anything I missed?

PR #763, opened 2022-03-06
Run CI on 3.x

krassowski:
Minor change to the CI config so that it runs on 3.x branch


krassowski:

PR #781, opened 2022-03-20
Bump (not yet in use) typdoc version

krassowski:
Bump typedoc version to drop some very outdated dependencies from lockfile. Typedoc results are not currently in use, but let's aim at integrating it in 4.0.


krassowski:

PR #795, opened 2022-04-25
Remove jupyterlab version constraints

MridulS:
## References

The built artifact on pypi and conda-forge takes care of the version constraints (<4) for jupyterlab. We could probably remove the explicit upper bounds from the user installation bits in the README.

## User-facing changes

A bit more cleaner installation instructions for the user.

Thanks! Let me know if I am missing something here :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,9 +113,9 @@ Use of a python `virtualenv` or a conda env is also recommended.
 1. install JupyterLab and the extensions
 
    ```bash
-   conda install -c conda-forge 'jupyterlab>=3.0.0,<4.0.0a0' jupyterlab-lsp
+   conda install -c conda-forge jupyterlab-lsp
    # or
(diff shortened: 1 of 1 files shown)


krassowski:
Thanks for the PR. Yes, a version of jupyterlab-lsp would get installed for the user, and it would be compatible with the jupyterlab version that they currently have installed, but that version would not be the same as described in README. Not a strong opinion here - I think that the range was introduced by @bollwyvl initially so curious to hear if there are other packaging concerns.

If we were to accept this change, could you please update the integrity tests? I believe there is a check verifying that this version is correct and removing the version will cause that check to fail.

krassowski:
Closing as stale. I am happy to reconsider this suggestion if we were to move the feature descriptions to documentation (so that the version in README would not need to be synced with the descriptions in README itself).

PR #800, opened 2022-05-08
Remove note about language server fork

joelostblom:
My understanding of https://github.com/jupyter-lsp/jupyterlab-lsp/issues/272#issuecomment-891286446 is that these changes how now been merged into the `python-lsp` conda package and there is no need to install the forked language server separately, is this correct? 

Does that mean that the recommended way to install `jupyterlab-lsp` including all the speed-ups is just:

```
conda install jupyterlab-lsp python-lsp-server
```

And there is no need to disable Jedi, since the updates to the pylsp also improved the slow Jedi calls? 

I have some issues with completion speed and can open a ticket with more info, but I just wanted to make sure that I followed the recommended installation procedure first.


krassowski:
Not all changes were merged but indeed for now it's better to remove this note

PR #809, opened 2022-06-22
Fix CI regressions, make tests work with JupyterLab 3.4

krassowski:
Maintenance-only PR


krassowski:

PR #778, opened 2022-03-20
Implement settings UI using native JupyterLab 3.3 UI 

krassowski:
## References

#196

Closes #417, supersedes #418.

## Code changes

- [x] update config schemas:
  - [x] pylsp (https://github.com/python-lsp/python-lsp-server/blob/develop/pylsp/config/schema.json)
  - [x] texlab (https://github.com/latex-lsp/texlab/blob/master/docs/options.md)
  - [x] bash (https://github.com/bash-lsp/bash-language-server/blob/vscode-client-1.12.1/vscode-client/package.json#L30)
  - [x] yaml (https://github.com/redhat-developer/yaml-language-server/blob/main/README.md)
  - [x] sql (some tasks remain, see follow up)
  - [x] docker (based on `package.json` and `package.nls.json` from https://github.com/microsoft/vscode-docker)
- [x] add config schemas
  - [x] pyright (https://github.com/microsoft/pyright/blob/main/packages/vscode-pyright/schemas/pyrightconfig.schema.json)
  - [x] typescript (https://github.com/typescript-language-server/typescript-language-server)
  - [x] r-language-server (based on README file)
  - [x] julia (https://github.com/julia-vscode/julia-vscode/blob/master/package.json, see https://github.com/julia-vscode/LanguageServer.jl/issues/625)
  - [x] [jedi-language-server](https://github.com/pappasam/jedi-language-server) does not support configuration, only initialization
  - [x] html/css/json - no schema for current (outdated) versions; need to migrate?
  - [x] unified - no workspace config handling yet
- [x] fix array types missing items specification, and top-level objects missing type and title
   - [x] contribute fixes back to `pylsp`
- [x] update run-time population of schema based on schemas returned by the server
   - [x] <s>do not display schema for servers which are not installed</s> but then the defaults are just ugly; instead went for adding "(this server was not detected as installed)" description; ideally we could grey out and collapse its section
   - [ ] <s>remove the defaults from user settings (like in [shortcuts](https://github.com/jupyterlab/jupyterlab/blob/f64fed3ce3ef3ecb10725e3d94725d6d9fa0f48a/packages/shortcuts-extension/src/index.ts#L175))</s> not feasible for now; explored at length see https://github.com/jupyter-lsp/jupyterlab-lsp/pull/778/commits/a91e33524aa45c0a6d03b51fbf0317fd31d56821#diff-eadb7a5104f0b088ea8c76c0b6d651644c21c51b419edc5ee6dcf54db9f5d2a3R93-R123
[...]

PR #810, opened 2022-06-23
Increase minimum Node.js to 14.0

krassowski:
Node 12 reached EOL in April.

Required for bumping of tested servers (in #778) as some dropped support for 12.0 already.


krassowski:

PR #820, opened 2022-07-08
Fix web socket url

### LINKED ISSUE #819: Wrong web socket url
## Description

Web socket url is not always the same as base url.
In my case base url hostname and web socket url hostname are different and they are specified in jupyterlab settings
I have investigated the code and saw that in jupyterlab-lsp web socket url is created from base url, which is not correct behavior for such cases ([see code](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/5d33fd2b3e306269b2926d3df19863f7ae87c693/packages/jupyterlab-lsp/src/connection_manager.ts#L394)).

## Expected behavior

I would expect web socket url to be taken from jupyterlab settings and not created from base url.

## Context

- Operating System and version: macOS 11.6.3
- Browser and version: Chrome, version 103
- JupyterLab version: 3.2.2

MikeSem:
## References

fixes #819 

## Code changes

Fixed web socket url
Used ServerConnection service to get valid web socket url

## Chores

- [x] linted

### DIFF
--- a/packages/jupyterlab-lsp/src/connection_manager.ts
+++ b/packages/jupyterlab-lsp/src/connection_manager.ts
@@ -1,4 +1,5 @@
 import { PageConfig, URLExt } from '@jupyterlab/coreutils';
+import { ServerConnection } from '@jupyterlab/services';
 import { Signal } from '@lumino/signaling';
 import type * as protocol from 'vscode-languageserver-protocol';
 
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you! Sorry about the delay here!

krassowski:

PR #829, opened 2022-08-22
Completion/cleanup when disabled

krassowski:
## References

Clean up all completer styles when disabling (noticed in https://github.com/jupyterlab/jupyterlab/issues/12830).

## Code changes

- Increase specificity of styles by adding `.lsp-completer` as needed
- when completer gets disabled
  - remove completion theme class from body
  - remove completer layout data from body
  - use upstream model for completion data (although refresh is still needed to rewire data connection) 

## User-facing changes

- styles should no longer spill when LSP completer is disabled via settings
- after re-enabling styles are restored and LSP completer works without a refresh (refresh is still needed to fully disable it, see note in "code changes").

None

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   - implement settings UI using native JupyterLab 3.3 UI ([#778])
 - bug fixes
   - use correct websocket URL if configured as different from base URL ([#820], thanks @MikeSem)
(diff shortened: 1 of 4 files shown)


krassowski:

PR #833, opened 2022-08-23
Fix LSP completion transform of path-likes when `insertText` missing

### LINKED ISSUE #832: Completion with missing insertText can get "undefined" value
## Description

Completion in typescript import names inserts "undefined". These completions have label but no `insertText`.

## Reproduce

1. Open typescript file
2. Enter `import { Signal } from '@lumino/;`
3. Press tab with cursor between `/` and `;` and select an option
4. See "undefined" getting inserted.

## Expected behavior

The label is inserted.

## Context

This line should be conditional:

https://github.com/jupyter-lsp/jupyterlab-lsp/blob/337cb02f80e4e6edea378f09fbde914093ed0ac9/packages/jupyterlab-lsp/src/features/completion/completion_handler.ts#L413

krassowski:
## References

Fixes #832

## User-facing changes

Completion of `import { Signal } from @lumino/<tab>` will work correctly.

## Chores

- [x] linted 
- [x] tested 
- [ ] documented 
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
   - implement settings UI using native JupyterLab 3.3 UI ([#778])
 - bug fixes
   - use correct websocket URL if configured as different from base URL ([#820], thanks @MikeSem)
-  - clean up all completer styles when compelter feature is disabled ([#829]).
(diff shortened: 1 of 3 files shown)


krassowski:

PR #836, opened 2022-08-23
Reduce signature and hover flickering

### LINKED ISSUE #834: Signature popup is flickering when typing
## Description

Peek 2022-08-23 13-26

krassowski:
## References

Fixes #834

## Code changes

`EditorTooltipManager` gest new methods: `showOrCreate`, `show`, `hide`

## User-facing changes

Signature tooltip will be hidden (rather than removed) if we know that we will be showing it again soon; if it has not changed (nor position, editor, etc), it will be shown as-is, if only content changed it will be updated.

Unfortunately, pylsp seems to return a message with empty signature every other time which is the direct cause of flickering in remaining instances.

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [x] changelog entry


krassowski:
Flaky completion test unrelated, fixed in #861.

Issue #828, opened 2022-08-21
Can't access files outside of jupyter root despite .lsp_symlink hack


slavistan:
Hi! I use jupyter locally and work with conda environments located outside of the jupyter content dir. I ran into the problem of being unable to access files:

![image](https://user-images.githubusercontent.com/35600954/185789101-1987b6ed-4da7-4929-9112-ea3636160eab.png)

As suggested in #750 the workaround is, inside of the jupyter content dir, to create a symlink named `.lsp_symlink` to the system's root dir: `ln -s / /home/user/my_jupyter_dir/.lsp_symlink`. However, even with the symlink in place, I get the same error despite being able to access the file via the symlink:

```
$ ls -la .lsp_symlink/home/stan/.conda/envs/pytorch_cuda3090/lib/python3.10/site-packages/transformers/models/auto/tokenization_auto.py
-rw-rw-r-- 2 stan stan 33339 Jun 18 09:36 .lsp_symlink/home/stan/.conda/envs/pytorch_cuda3090/lib/python3.10/site-packages/transformers/models/auto/tokenization_auto.py
``` 

Is there anything I've missed? The instructions for the symlink hack differ between the comments in the above issue and the [installation instructions](https://github.com/jupyter-lsp/jupyterlab-lsp#installation), so I may have misunderstood something.

## Context

- Operating System and version: Arch Linux 5.19.1
- Browser and version: Brave 1.42.97.-1
- JupyterLab version: 3.4.5
- Language server and version: python-lsp-server 1.5.0

```
$ jupyter serverextension list
# config dir: /home/stan/.conda/envs/pytorch_cuda3090/etc/jupyter
#     jupyter_lsp  enabled
#     - Validating...
#       jupyter_lsp 1.5.1 OK
#     jupyterlab  enabled
#     - Validating...
#       jupyterlab 3.4.5 OK
# config dir: /etc/jupyter
#     jupyterlab  enabled
#     - Validating...
#       jupyterlab 3.4.5 OK

$ jupyter labextension list
# JupyterLab v3.4.5
# /home/stan/.conda/envs/pytorch_cuda3090/share/jupyter/labextensions
#         @jupyter-widgets/jupyterlab-manager v3.0.0 enabled OK (python, jupyterlab_widgets)
#         @krassowski/jupyterlab-lsp v3.10.1 enabled OK (python, jupyterlab-lsp)
# 
# /usr/share/jupyter/labextensions
[...]

PR #846, opened 2022-08-27
Fix instructions for .lsp_symlink workaround

slavistan:
@krassowski As per #828, here's a draft.

I've left the unix-only part in the instructions. I am, however, unsure about why this wouldn't affect Windows users. Do you have any more insight on this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,21 +164,25 @@ Use of a python `virtualenv` or a conda env is also recommended.
    or permanently by setting `c.Completer.use_jedi = False` in your
    [`ipython_config.py` file](https://ipython.readthedocs.io/en/stable/config/intro.html?highlight=ipython_config.py#systemwide-configuration).
 
-1. (Optional, Linux/OSX-only) to enable opening files outside of the root
-   directory (the place where you start JupyterLab), create `.lsp_symlink` and
-   symlink your `/home`, or any other location which includes the files that you
-   wish to make possible to open in there:
+1. (Optional; *nix-only) As a security measure Jupyter limits file access to the Jupyter root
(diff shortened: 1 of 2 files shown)


krassowski:
> I am, however, unsure about why this wouldn't affect Windows users.

Windows users would need to enable Developer Mode in order to create symlinks and I do not have bandwidth to test if this workaround works on Windows correctly, hence the readme does not encourage it.

I added a link to a plan for a better solution which would not require a symlink and allow fine-grained access control: https://github.com/jupyter-lsp/jupyterlab-lsp/issues/850.

README.md
```diff
@@ -164,25 +164,28 @@ Use of a python `virtualenv` or a conda env is also recommended.
    or permanently by setting `c.Completer.use_jedi = False` in your
    [`ipython_config.py` file](https://ipython.readthedocs.io/en/stable/config/intro.html?highlight=ipython_config.py#systemwide-configuration).
 
-1. (Linux/OSX-only) As a security measure Jupyter limits file access to the Jupyter root
-   directory (the place where you launch the Jupyter server). Thus, in order to
-   allow `jupyterlab-lsp` to navigate to external files such as packages
+1. (Optional, Linux/OSX-only) As a security measure by default Jupyter serve only allows
```
slavistan, line 167: You meant to write "Jupyter server", correct?

krassowski: Indeed, thanks!

```suggestion
1. (Optional, Linux/OSX-only) As a security measure by default Jupyter server only allows
```

krassowski:
Thank you @slavistan!

PR #852, opened 2022-09-12
[wip] defer initialization until event loop is available

bollwyvl:
## References

- minimal fix for #851

## Code changes

- moves full initialization off critical path to starting `JupyterServerApp`
  - when it does initialize, moves potentially-expensive actions to a thread pool executor to not inadvertently block other early functions
  - when done, set a ready flag
- makes more of the handler methods asynchronous, and wait on the ready flag

## User-facing changes

- slow startup would now be deferred, allowing the server to start more quickly
- the initial request for language status might be delayed some
  - _typically_ it takes a while and many hundreds of HTTP requests, to get all the way up to a full-rendered document that could even request whether a language server is available
- the debug log message comes _after_ the server announces its location, making it harder to find the server URL

- n/a, though some of the internal methods

## Future Work

- this is still "blocks" (from an information theory perspective) the ability to start _any_ session until _all_ servers are discovered
  - theoretically with an `isComplete` flag on the response, we could handle partial setup better, and hoist more errors to the client log

## Chores

- [x] linted
- [ ] tested
  - the new server initialize code isn't under unit test, probably should be 
  - benchmark startup somehow?
- [ ] documented
- [ ] changelog entry

### DIFF
--- a/python_packages/jupyter_lsp/jupyter_lsp/handlers.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/handlers.py
@@ -25,7 +25,8 @@ class LanguageServerWebSocketHandler(  # type: ignore
 
     language_server = None  # type: Optional[Text]
 
-    def open(self, language_server):
+    async def open(self, language_server):
+        await self.manager.ready()
         self.language_server = language_server
         self.manager.subscribe(self)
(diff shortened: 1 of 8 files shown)


bollwyvl:
Continued in #882

Issue #854, opened 2022-09-13
Fails to build the binder image


yamaton:
## Description

The `jupyterlab-lsp` Binder fails to launch due to an error during the build.

![jupyterlab-lsp binder](https://user-images.githubusercontent.com/256288/189947423-4ff2f1d0-2f19-4968-a4f4-ff20ada3dca9.png)

<details>

```
lsp-ws-connection: node:internal/crypto/hash:71
lsp-ws-connection:   this[kHandle] = new _Hash(algorithm, xofLen);
lsp-ws-connection:                   ^
lsp-ws-connection: Error: error:0308010C:digital envelope routines::unsupported
lsp-ws-connection:     at new Hash (node:internal/crypto/hash:71:19)
lsp-ws-connection:     at Object.createHash (node:crypto:133:10)
lsp-ws-connection:     at module.exports (/home/jovyan/packages/lsp-ws-connection/node_modules/webpack/lib/util/createHash.js:135:53)
lsp-ws-connection:     at NormalModule._initBuildHash (/home/jovyan/packages/lsp-ws-connection/node_modules/webpack/lib/NormalModule.js:417:16)
lsp-ws-connection:     at handleParseError (/home/jovyan/packages/lsp-ws-connection/node_modules/webpack/lib/NormalModule.js:471:10)
lsp-ws-connection:     at /home/jovyan/packages/lsp-ws-connection/node_modules/webpack/lib/NormalModule.js:503:5
lsp-ws-connection:     at /home/jovyan/packages/lsp-ws-connection/node_modules/webpack/lib/NormalModule.js:358:12
lsp-ws-connection:     at /home/jovyan/packages/lsp-ws-connection/node_modules/loader-runner/lib/LoaderRunner.js:373:3
lsp-ws-connection:     at iterateNormalLoaders (/home/jovyan/packages/lsp-ws-connection/node_modules/loader-runner/lib/LoaderRunner.js:214:10)
lsp-ws-connection:     at Array.<anonymous> (/home/jovyan/packages/lsp-ws-connection/node_modules/loader-runner/lib/LoaderRunner.js:205:4)
lsp-ws-connection:     at Storage.finished (/home/jovyan/packages/lsp-ws-connection/node_modules/webpack/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:55:16)
[...]

PR #855, opened 2022-09-13
Fix an OpenSSL-related error in building the Binder image

yamaton:
## References

A PR for #854 

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


krassowski (approved):
Thanks! This is a good workaround for now. I am investigating the reason and will fix the culprit in a subsequent PR.

PR #860, opened 2022-09-25
Fix sporadic misplacement of hover tooltips

yamaton:
## References

Fixes https://github.com/jupyter-lsp/jupyterlab-lsp/issues/859

## Code changes

* Input and output types of the function `on_hover` 
  * Technically, the change in `on_hover` argument is unnecessary; it was to clarify the `Throttler`'s behavior.
* Process hover only when the virtual positions of a query and a response match

## User-facing changes

None.

None.

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/packages/jupyterlab-lsp/src/features/hover.ts
+++ b/packages/jupyterlab-lsp/src/features/hover.ts
@@ -46,11 +46,6 @@ interface IResponseData {
   ce_editor: CodeEditor.IEditor;
 }
 
-interface IOnHoverResponse {
-  hover: lsProtocol.Hover;
-  position: IVirtualPosition;
-}
-
 /**
  * Check whether mouse is close to given element (within a specified number of pixels)
  * @param what target element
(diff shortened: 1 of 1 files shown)


yamaton:
I'll investigate how the hover test fails occassionally.

packages/jupyterlab-lsp/src/features/hover.ts
```diff
@@ -443,28 +455,32 @@ export class HoverCM extends CodeMirrorIntegration {
...
+        const promise = this.debounced_get_hover.invoke(virtual_position);
         this._previousHoverRequest = promise;
         let response = await promise;
         if (this._previousHoverRequest === promise) {
           this._previousHoverRequest = null;
         }
-        if (response && this.is_useful_response(response)) {
+        if (
+          response &&
+          is_equal(response.position, virtual_position) &&
```
krassowski, line 466: [`lsProtocol.Hover`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#hover) optionally includes `range` which we use for caching purpose. If `range` is present it might be more useful than the memorised position as we could still use a response sent for another character in the same token.

yamaton: Thank you for the review and the suggestion; validating on_hover/throttler response by `range` makes sense.
The commit below is my first attempt along the line. I moved some code around to obtain `range` within `on_hover` so it's a bit ugly...

krassowski:
Is this still in progress? Asking since I will be reorganising the codebase this week for 4.0 release so if it is good enough or will be soon, then we could merge and iterate later (a test case would be nice, but can be added in another PR).

Also if you have any other local work I can wait a bit before going into full refactor mode (which will mean that all frontend PRs will need to be re-applied manually).

yamaton:
Thank you for asking. 
It seems the tests went successful so it's good to go, I suppose.

I do have a working demo of https://github.com/jupyter-lsp/jupyterlab-lsp/issues/362 so I'll send it as a PR later. But I don't mind re-working on the PR after the full refactor of the code base.

krassowski (approved):

krassowski:
Thanks!

PR #863, opened 2022-09-27
Fix more flaky completion tests

### LINKED ISSUE #830: Flaky R completion tests on CI
Two tests seem to fail randomly on CI, but more frequently on Windows:

```
Completes In R Magics                                                 | FAIL |
Element 'css:.jp-Completer.jp-HoverBox' did not appear in 35 seconds.
Completes Correctly With R Double And Triple Colon                    | FAIL |
Element 'css:.jp-Completer.jp-HoverBox' did not appear in 35 seconds.
```

There might be something off with trigger sequence.

krassowski:
## References

Fixes #830

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/atest/05_Features/Completion.robot
+++ b/atest/05_Features/Completion.robot
@@ -193,7 +193,7 @@ User Can Select Lowercase After Starting Uppercase
 Mid Token Completions Do Not Overwrite
     # `disp<tab>data` → `display_table<cursor>data`
     Place Cursor In Cell Editor At    9    line=1    character=4
-    Capture Page Screenshot    01-cursor-placed.png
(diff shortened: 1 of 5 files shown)


krassowski:
Merging since it is a sufficient improvement for now.

Remaining flake:
- Invalidates On Focus Loss 1/6
- Works When Kernel Is Shut Down 1/6
- Mid Token Completions Do Not Overwrite x 4/6
(Element 'css:.jp-Completer-item[data-value="display_table"]' not visible after 15 seconds.)
- Continuous Hinting Works 1/6
- Highlights are modified after typing
- Completes Correctly With R Double And Triple Colon 1/6

PR #867, opened 2022-10-11
Fix hover tooltip not updated after character deletions

yamaton:
## References

Fixes https://github.com/jupyter-lsp/jupyterlab-lsp/issues/866

## Code changes

Changed EditorAdapter to emit update signals when nontrivial `change.removed` exists.

## User-facing changes

None.

My concern is that the original behavior might have been intentional as was in the comment. 

```typecript
// deletion - ignore
```

If so, this PR may affect other signal receivers. 

## Chores

- [x] linted 
- [x] tested 
- [ ] documented 
- [x] changelog entry


yamaton:
FYI: I could reproduce the CI error about `jupyter serverextension list` in the following minimum environment. 

```shell
mamba create -c conda-forge -n py37-minimum-test python=3.7 jupyterlab=3.3 jupyter_server
conda activate py37-minimum-test 
jupyter serverextension list
```

krassowski:
> My concern is that the original behavior might have been intentional as was in the comment.

I think that the original intent was to avoid triggering completer on deletion, but indeed this is not where this should be handled. I think it is fine to change it - let's give it a try.

krassowski (approved):

PR #869, opened 2022-10-23
Add notebook dependency to CI to test `serverextension`, fix CI

krassowski:
An attempt to fix CI failure seen in #867

### DIFF
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -454,6 +454,9 @@ jobs:
       - name: Install JupyterLab
         shell: bash -l {0}
         run: python -m pip install 'jupyterlab${{ matrix.lab }}'
+      - name: Install Notebook (to test serverextension)
+        shell: bash -l {0}
+        run: python -m pip install 'notebook'
       - name: Install the server package
         run: cd dist && python -m pip install -vv jupyter_lsp*.whl
       - name: install the source extension
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -47,7 +47,7 @@ jobs:
         lab: ['>=3.3.0,<4.0.0a0']
         r: ['>=4']
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - uses: cschleiden/replace-tokens@v1
         with:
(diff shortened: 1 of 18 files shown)


krassowski:
The lint failure is due to robottidy changing `Test Tags` to `Force Tags` (https://github.com/MarketSquare/robotframework-tidy/issues/432) and then robocop complaining that it should be the other way. Nothing to do other than wait for a release upstream.

bollwyvl:
The individual rule could be relaxed for now...

krassowski:
Kicking CI since the new release is already available and on conda-forge (thanks for a quick merge @bollwyvl).

krassowski:
bash language server failures are gone (I did suspect this was an intermittent issue, waiting helped), but now MacOS smoke fails with cryptic:

```
+ jupyter labextension list
/Users/runner/work/_temp/03c846b0-ddaa-4fd7-be26-b0a115a5b859.sh: line 2: jupyter: command not found
```

but `pip list` shows all jupyter dependencies fine. I know we can workaround, but it sounds like an upstream problem again. Does it ring a bell?

This constant CI flake is really burning out energy.

bollwyvl:
Perhaps try `python3 -m jupyter`?

Maybe some new `jupyter_core` path vulnerability fallout? 

There is still the option of pinning the dependencies down harder with e.g. `conda-lock` so that handling issues like this can be a more intentional, single-purpose PR, if possible.

bollwyvl:
It may also be worth upgrading the various `setup-*` actions to the latest versions, as things outside our control definitely _will_ break in some places soon.

krassowski:
I will experiment in https://github.com/krassowski/jupyterlab-lsp/pull/3

bollwyvl:
Perhaps this:
```robot
css:button[title="Refresh File List"]
```
Could be pointed at the more stable:

```robot
css:button[data-command="filebrowser:refresh"]
```

bollwyvl:
```
Evaluating expression '__import__("urllib.request").request.urlopen("http://localhost:38635/@est/api/shutdown?token=jGIubKOf", data=[])' failed: URLError: <urlopen error [Errno 111] Connection refused>
```

This feels almost certainly related to the `jupyter_core` change, and might be [fixable here](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/master/atest/Keywords.resource#L43):

```robot
Create Notebok Server Config
    # ...
    ${conf} =    Set Variable    ${NOTEBOOK DIR}${/}${JPSERVER CONF JSON}
    Set Environment Variable
    ...    name=JUPYTER_CONFIG_DIR
    ...    value=${NOTEBOOK DIR}
```

bollwyvl (approved):

PR #873, opened 2022-11-08
Attempt to fix #872

### LINKED ISSUE #872: Force virtual document to use additional file extension?
## Description

I follow the instructions for the Scala Language Server (Metals) integration: https://jupyterlab-lsp.readthedocs.io/en/latest/Configuring.html#example-scala-language-server-metals-integration.

There weren't any errors, but the LSP wasn't returning any suggestions aside from the kernel ones. When testing using a Python kernel there were LSP suggestions.

## Reproduce

Follow the instructions here: https://jupyterlab-lsp.readthedocs.io/en/latest/Configuring.html#example-scala-language-server-metals-integration.

I enabled debug traces for metals by creating the file:
```
2022.11.08 13:04:00 INFO  tracing is enabled: /Users/dleen/code/jupyterlab-lsp/.metals/lsp.trace.json
2022.11.08 13:04:02 INFO  logging to file /Users/dleen/code/jupyterlab-lsp/.virtual_documents/.metals/metals.log
```

Looking at the `lsp.trace.json` file while interacting with JupyterLab showed output like this:
```
[Trace - 01:08:07 PM] Received request 'textDocument/documentHighlight - (1)'
Params: {
  "textDocument": {
    "uri": "file:///Users/dleen/code/jupyterlab-lsp/.virtual_documents/Untitled.ipynb"
  },
  "position": {
    "line": 0,
    "character": 15
  }
}


[Trace - 01:08:07 PM] Sending response 'textDocument/documentHighlight - (1)'. Processing request took 9ms
Result: []
```

Notice that the result is empty `[]` and also notice that the `uri` is: `file:///Users/dleen/code/jupyterlab-lsp/.virtual_documents/Untitled.ipynb` which has file extension `.ipynb`.

After reading this doc: https://jupyterlab-lsp.readthedocs.io/en/latest/Configuring.html#making-custom-servers-work-with-notebooks

> many language servers only handle files with specific file extensions and refuse to operate if not provided with such; the file extension of a native script for a given language (this is other than .ipynb), derived from file_extension field of language_info, will be added to the name of the notebook when communicating with the language server to satisfy the file extension check.

I guessed that the problem might be with the missing extension.

I commented out these lines: https://github.com/jupyter-lsp/jupyterlab-lsp/blob/master/packages/jupyterlab-lsp/src/virtual/document.ts#L755-L757

And the `uri` became: `  "uri": "file:///Users/dleen/code/jupyterlab-lsp/.virtual_documents/Untitled.ipynb.scala.sc",` and the LSP started working correctly.

I'm not sure what the `parent` is controlling here or how to force the addition of the file extensions otherwise.
```
    if (!this.parent) {
      return encodedPath;
    }
```

Any pointers would be great! Thanks!

dleen:
## References
See #872. This adds the file extension in certain cases, but not sure if this is right.

## Chores

- [x] linted 
- [x] tested 
- [ ] changelog entry


krassowski:
Hi sorry that I did not get to merging this. Since this code now lives [in JupyterLab repository](https://github.com/jupyterlab/jupyterlab/blob/1d989bad87e89307a336d2f671c9cc98ec37f5e9/packages/lsp/src/virtual/document.ts) I would encourage submitting a pull request (ideally with a test case) there.

PR #881, opened 2022-12-09
Use app http settings for status, WebSockets

bollwyvl:
## References

- fixes #880

## Code changes

- [x] initialize `LanguageServerManager` with `app.serviceManager.settings`is being used
- [x] use these settings for doing API requests and WebSockets
- [x] handle a race condition for very fast settings loading (no http request) for plugins with a `PromiseDelegate`
- [x] rev some node versions (not sad about retiring 14 a little early)
- [x] resolve and deduplicate `yarn.lock`: get "one true version" of `typescript`, `prettier`, `webpack`
- [x] use webpack 5 for `lsp-ws-connection`
- [x] add sourcemaps (we can take these out, but... would be _really_ nice for downstream debugging)
- [x] refresh robot tooling

## User-facing changes

- hopefully few visible changes

## Backwards-incompatible changes

- hopefully none: no signatures were changed

## Chores

- [x] linted
- [ ] tested 
- [ ] documented
- [x] changelog entry

PR #882, opened 2022-12-10
Use app's HTTP settings, handle feature settings race condition, CI/build updates, lazy server discovery

### LINKED ISSUE #880: Use the host application's server connection settings
## Elevator Pitch

Allow non-JupyterLab applications to override the server connection settings used for fetching status and creating WebSockets.

## Motivation

Apps that aren't a full JupyterLab often need to do extra things to `serverSettings` to communicate (or not) with a server properly.

Specifically, jupyterlite 

[jupyterlite-lsp](https://github.com/jupyterlite/jupyterlite-lsp) (real broken right now) can _almost_ demonstrate providing mock WebSockets, but needs nasty hacks. Unfortunately, even with these hacks, I haven't yet been able to [demonstrate one server working](https://github.com/jupyterlite/jupyterlite-lsp/issues/1) due to WebPack issues :crying_cat_face: .

## Design Ideas

- use `serverSettings: app.serverManager.serverSettings` in more places
- hand this down to the connection manager, and use that instead of `new WebSocket`

bollwyvl:
## References

- supersedes #881
- fixes #880
- progress on #851

## Code changes

- `/`
  - [x] resolve and deduplicate `yarn.lock`: get "one true version" of `typescript`, `prettier`, `webpack`
  - [x] add sourcemaps (we can take these out, but... would be _really_ nice for downstream debugging)
- `.github`
  - [x] bump more CI action versions to squash warnings
- `@jupyter-lsp/jupyterlab-lsp`
  - [x] initialize `LanguageServerManager` with `app.serviceManager.settings`is being used
  - [x] use these settings for doing API requests and WebSockets
  - [x] handle a race condition for very fast settings loading (no http request) for plugins with a `PromiseDelegate`
  - [x] update `@jupyterlab/*` dependencies to match python-level dependency
- `@jupyter-lsp/lsp-ws-connection`
  - [x] use webpack 5 for `lsp-ws-connection`
  - [x] normalize `void` for handling of promises
  - [x] bumps version 
- `jupyter_lsp`
  - [x] use `importlib(_|.)metadata` for `entrypoints`
  - [x] bumps version
  - [x] defers language server spec discovery until the event loop is ready 

## User-facing changes

- hopefully few visible changes

- python
  - downstreams which interact with `jupyter_lsp.LanguageServerManager` directly may need to `await` more things

## Chores

- [x] linted 
- [x] tested 
- [x] documented 
- [x] changelog entry


bollwyvl:
Ha, green!

I can roll back the server lazy loading changes if desired.

python_packages/jupyterlab_lsp/setup.cfg
```diff
@@ -36,5 +36,5 @@ zip_safe = False
 python_requires = >=3.7
 
 install_requires =
-    jupyter_lsp >=1.4.0
+    jupyter_lsp >=1.6.0
```
krassowski, line 39: If this is required, should we consider making a major version bump?

bollwyvl: Yeah, it's probably worth pulling this part onto a separate PR, and doing some manual reverts.

Also, of note, we might need to be very ginger with `ipykernel 6.19.*`. as i was seeing some message-related stuff that might confuse our test rig.

packages/lsp-ws-connection/src/ws-connection.ts
```diff
@@ -1,5 +1,6 @@
 import * as events from 'events';
 
+import 'setimmediate';
```
krassowski, line 3: Why is this needed? It looks as not maintained for a while now, and I am not sure where it gets used. Is this due to a setImmediate cal in one of dependencies?

python_packages/jupyterlab_lsp/setup.cfg
krassowski: Let's keep going an release 2.0 then. I hope to get to releasing stuff over the break.

packages/lsp-ws-connection/src/ws-connection.ts
bollwyvl: > a setImmediate cal in one of dependencies?

Deep in one of the `vscode-*` ones, and extremely hard to track down (hence the sourcemaps).

python_packages/jupyterlab_lsp/setup.cfg
bollwyvl: Just to clarify, which of the below:
- make a new PR with this HEAD (perhaps force push to #852) and this roll this back to not touch server
- keep going on this, but instead of bumping to 1.6, bump to 2.0

Some background:

I guess it depends on how semver we're getting: the contract with the frontend doesn't actually change at all, but I am wary of not also _testing_ against the oldest supported serverextension, whereas moving the pin makes it relatively clear.

The manager API does change a bit, to avoid paying for language server discovery. So if someone was overloading the manager, but using the stock routes, that could be an issue. 

The listener API does not change.

packages/lsp-ws-connection/src/ws-connection.ts
bollwyvl: Maybe these instead: https://www.npmjs.com/package/timers-browserify

bollwyvl: haha, uses that, too

python_packages/jupyterlab_lsp/setup.cfg
krassowski: > keep going on this, but instead of bumping to 1.6, bump to 2.0

I was thinking this one. Especially if we could also get https://github.com/jupyter-lsp/jupyterlab-lsp/pull/636 in. Otherwise maybe 1.6 is fine.

> So if someone was overloading the manager, but using the stock routes, that could be an issue.

Yes. I don't know of anyone who does that. I don't feel strong about this especially since we did not advertise support for overloading it like that.

bollwyvl: > Especially if we could also get https://github.com/jupyter-lsp/jupyterlab-lsp/pull/636 in.

That would be _great_. 

Expounding  a little further: should 2.0 be 
- `python >=3.8`
  - Ipykernel, etc. have already started to move on.
- `jupyter_server >=2.0`
  - it's just going to be really messy for a while, but it will be a lot simpler to support

bollwyvl:
Cool, glad to have this one in.

I'll have another look at working the CI assets into the downstream on `jupyterlite-lsp`...

PR #892, opened 2022-12-31
Speed up tests by not reloading page within suites

krassowski:
An attempt to speed up tests. Promising results locally (1/3rd off).

### DIFF
--- a/atest/04_Interface/BackendFailure.robot
+++ b/atest/04_Interface/BackendFailure.robot
@@ -28,3 +28,4 @@ Setup Missing Extension Test
 
 Teardown Missing Extension Test
     Set Server Extension State    enabled=${True}
+    Setup Server and Browser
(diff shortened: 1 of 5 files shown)


krassowski:
| test | before (#891) | before (master) | after |
|---|---|---|---|
| ubuntu py3.7 | 147m | 121m | 35m |
| ubuntu py3.10 | 145m | 44m | 35m |
| macos py3.7 | 104m | 69m | 56m |
| macos py3.10 | 138m | 90m | 58m |
| windows py3.7\* | 134m | 68m | 51m |
| windows py3.10\* | 79m | 122m | 54m |

\*) failing here due to https://github.com/jupyter-lsp/jupyterlab-lsp/issues/894 so in practice should be even faster

bollwyvl:
Very good progress!

It may also be worth looking at whether `pabot` can shave some time off, as a lot of the tests are just `Wait`ing for things (servers to start, kernels to be ready, jedi to do whatever it does).

I have found on windows, though, that doing too many at once is problematic, so usually only try `2` processes on that platform.

PR #893, opened 2023-01-02
Implement eliding for paths/files

### LINKED ISSUE #787: code completion missing string for very long path completion
**Description**
When I try to type in  a file path, if the path is vey long, the remaining path string will not be seen in code completion! 
image

krassowski:
## References

Elide long path completions, fixes #787 and fixes https://github.com/jupyterlab/jupyterlab/issues/12830

## Code changes

The content of the first highlighted (`<mark>`ed) part is elided for paths

- `direction: rtl` is used to reverse side of eliding
   - to prevent wrong placement of neutral BiDi characters (e.g. `.` or `/`) `<bdo>` tag wrapper is used for `<mark>` contents; the alternative, CSS `unicode-bidi` would also require an HTML wrapper, hence BDO tag is a simpler choice.
- `display:flex`/`flex-shrink: 1` is used to stretch the mark to as much as possible

I also explored a possibility of eliding of the middle part of the path based on separator, but there does not appear to be a good solution (which would have good performance with large number of matches).

## User-facing changes

Long paths are elided. User can hover over the elided part to see title with full contents.

Screenshot from 2023-01-02 03-59-58

None

## Chores

- [x] linted 
- [ ] tested 
- [ ] documented 
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   - implement jump target selector and jump to references ([#739])
   - implement settings UI using native JupyterLab 3.3 UI ([#778])
   - add option to show hover tooltip automatically ([#864], thanks @yamaton)
+  - implement eliding for long paths/files in completer ([#893])
 - bug fixes:
   - use correct websocket URL if configured as different from base URL ([#820], thanks @MikeSem)
   - clean up all completer styles when completer feature is disabled ([#829]).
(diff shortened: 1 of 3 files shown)


krassowski:
Windows failures are unrelated (https://github.com/jupyter-lsp/jupyterlab-lsp/issues/894).

PR #904, opened 2023-02-12
Adds experimental Go spec

joelschutz:
## References

Please add Issue reference here

## Code changes

- A simple initial spec for gopls

## User-facing changes

- Adds minimal support for Go
- 

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/python_packages/jupyter_lsp/jupyter_lsp/specs/__init__.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/specs/__init__.py
@@ -11,6 +11,7 @@ from .pyls import PalantirPythonLanguageServer
 from .pyright import PyrightLanguageServer
 from .python_lsp_server import PythonLSPServer
 from .r_languageserver import RLanguageServer
+from .gopls import GoLanguageServer
 from .sql_language_server import SQLLanguageServer
 from .texlab import Texlab
 from .typescript_language_server import TypescriptLanguageServer
@@ -32,6 +33,7 @@ py_palantir = PalantirPythonLanguageServer()
 py_lsp_server = PythonLSPServer()
 pyright = PyrightLanguageServer()
 r = RLanguageServer()
+go = GoLanguageServer()
 tex = Texlab()
 ts_old = JavascriptTypescriptLanguageServer()
 ts = TypescriptLanguageServer()
--- a/python_packages/jupyter_lsp/jupyter_lsp/specs/gopls.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/specs/gopls.py
@@ -1,0 +1,38 @@
+from .utils import ShellSpec
+from typing import Union
+
+TROUBLESHOOT = """\
+You probably just need to install either *gopls* or the *jupyter-gopls*
+just use the commands below.
+"""
+
+
+class GoLanguageServer(ShellSpec):
+    key = cmd = "gopls"
+    package = "jupyter_gopls" # Package with parser for golang
+    args = ["serve"]
+    is_installed_args = ["version"]
+    languages = ["go"]
+
+    spec = dict(
+        display_name=key,
+        mime_types=["text/x-go", "text/go"],
+        urls=dict(
+            home="https://pkg.go.dev/golang.org/x/tools/gopls",
+        ),
+        # config_schema=load_config_schema(key), # TODO Find this options for this server
+        troubleshoot=TROUBLESHOOT,
+        shadow_file_ext=".go", # Not implemented yet, may be removed in future. See issue #872
+        install=dict(
+            go="go get golang.org/x/tools/gopls@latest", # This may never change
+            pip="pip install {}".format(package), # This package is not published yet
+        )
+    )
+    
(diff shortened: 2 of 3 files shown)


bollwyvl:
This is... fine. But adding new heavyweight languages is... heavy, even if we _can_ get them form conda-forge. We typically require adding full unit and acceptance tests, which can be a bit of a slog, especially on windows/osx.

A quicker path to getting something working for others might be to publish this as a standalone package on PyPI of just this spec, e.g. `jupyterlab-lsp-go`, which could be linked to from the docs on this site.

joelschutz:
I'm not really sure on the best way to implement this support yet, the main question is how should the shadow files be managed. *Heavy* languages usually need a fair amount of boilerplate and the parsing between this very diferrent files needs some code to work.

> A quicker path to getting something working for others might be to publish this as a standalone package on PyPI of just this spec, e.g. `jupyterlab-lsp-go`, which could be linked to from the docs on this site.

You say a fork `jupyterlab-lsp`, right? I have other changes that I'd need, so this is a good path to start development. Other thing that i'm thinking about is to create a wrapper around `gopls` to work with the currenty format. What do you think?

bollwyvl:
> You say a fork jupyterlab-lsp, right?

No, luckily that wouldn't be necessary, or good for anyone (go ask `kite`, and the users they convinced to install their fork).

The underlying `jupyter-lsp` server is [extensible](https://jupyterlab-lsp.readthedocs.io/en/latest/Extending.html#extend-jupyter-lsp) via the relatively-unique-to-python `entry_points` mechanism and/or Jupyter's "well-known" files, e.g. `$PREFIX/share/jupyter/jupyter_server_config.d/jupyter-golsp.json`... though, sadly, the JSON format isn't usually powerful enough to handle language server configuration.

So, with e.g. [`flit`](https://flit.pypa.io/en/stable/pyproject_toml.html), you could get away with a structure like:

```
LICENSE
README.md
pyproject.toml
jupyter_golsp.py
```

Where the contents of `jupyter_golsp.py` contained the spec proposed on this PR (with absolute instead of relative imports), and advertised in `pyproject.toml`.

Of course, if you end up needing to do more (e.g. custom listeners, custom syntax highlighting in the browser), that structure would of course get more complex.

All this is to say: you're guaranteed to get something out faster published separately, rather than waiting for us to go through all the hoops. Having docs for this process would be _great_, too.

krassowski:
@joelschutz are you still interested in driving this forward? I am happy to accept a PR, just needs to include tests as mentioned above.

joelschutz:
I don't plan to keep working with this things in the near future, but I'd like to mention `GoNB` project by @janpfeifer that was able to incorporate completion using `gopls` in the kernel. I think he can help with this implementation.

janpfeifer:
hi all, <pt-BR>e obrigado por me incluir Joel!</pt-BR>(thanks for looping me in).

So I'm wondering if the python model will work for Go, or at least how GoNB works, since some assumptions work differently.

For instance GoNB compiles the whole Go code at every cell execution and memorizes definitions -- as opposed to "interpreting" it in a REPL loop. That works because Go compilation time is fast enough that it doesn't matter. But that has some implications, along with other considerations that I'll list below:

1. GoNB keeps a mapping of the line number in the assembled Go code (from all cells executed so far) to the cell ids (# of the execution) and their line number. There needs to be also a mapping of the cursor. GoNB keeps that to translate that to `gopls`.  Something like this would need to be made for Go I suppose, since `gopls` expects a Go program, not independent cells -- there are more caveats here, but I won't get into the detail.
2. Usually the Go dependency/package system works super well, and we import things and never think about something like `pip install` things. We just do `go import` and things just work -- one of the reasons I use Go actually. So GoNB does that automatically for the user, and `gopls` works with that. I think an embedded Go support in jupyterlab-lsp should also include this functionality, it's one of the big plus of using Go in my book at least.
3. Inform `gopls` of local files that are modified: both in Python and in Go, I often use Jupyter jointly developing some project, that includes separate libraries, in the local file.  I'm not sure how the language server works in Python, but `gopls` needs to be informed that some of these libraries aren't to be pulled from github, but instead used from the local disk, and which directories to monitor for changes. A good integration definitely should support this, and GoNB does this magic for the user (see *Developing Go libraries with a notebook* in the [tutorial](https://github.com/janpfeifer/gonb/blob/main/examples/tutorial.ipynb)). 

Those are from the top of my mind, but it was tricky to get right, but it seems to be work in most (not all) times: Control+I for context help, or "shift+tab".

I'm not sure about the details of what you want to achieve in `jupyterlab-lsp`, but could it try to talk to GoNB, and through it get the LSP services you need ? If you tell me the features you would like to consider, we can define a new message and I'm happy to implement GoNB side of it -- we can even use [custom messages](https://jupyter-client.readthedocs.io/en/latest/messaging.html#custom-messages) to communicate if it makes sense to you, or something else -- I'm currently working on Widgets support for GoNB using these custom messages from the front-end, and it seems to be working great.

cheers

PR #929, opened 2023-04-16
Update texlab expectation

krassowski:
## References

Address failure seen in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/927 and described in https://github.com/jupyter-lsp/jupyterlab-lsp/issues/894#issuecomment-1510410526

## Code changes

Updated expected diagnostic value

## User-facing changes

None

None


bollwyvl:
Yep, the build chain changed a fair amount, too... so I guess that's on me.
With rust stuff, if it builds and passes CLI smoke tests, I usually just
ship it. 😳

krassowski:
50 minutes mamba setup on MacOS with py3.10, worth keeping an eye if this is a pattern or one-of:

Screenshot from 2023-04-16 17-34-38

PR #939, opened 2023-05-28
Await sleep coroutine while waiting for manager to be ready

jinzhen-lin:
## References

In jupyter-lsp 2.1.0, when I open the jupyterlab page, the terminal give me a warning

```
/root/miniconda3/lib/python3.8/site-packages/jupyter_lsp/manager.py:130: RuntimeWarning: coroutine 'sleep' was never awaited
  asyncio.sleep(0.1)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

And then the process is stuck (and cpu 100%).

I notice that the coroutine `asyncio.sleep` is not sleeped here and it should be fixed.

https://github.com/jupyter-lsp/jupyterlab-lsp/blob/fee5ce6de8da23c9d00eb201d305a120a5e747a3/python_packages/jupyter_lsp/jupyter_lsp/manager.py#L128-L131

## Code changes

I add a `await` before `asyncio.sleep(0.1)` to fix it.

### DIFF
--- a/python_packages/jupyter_lsp/jupyter_lsp/manager.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/manager.py
@@ -127,7 +127,7 @@ class LanguageServerManager(LanguageServerManagerAPI):
 
     async def ready(self):
         while not self._ready:  # pragma: no cover
-            asyncio.sleep(0.1)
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you!

PR #949, opened 2023-07-02
JupyterLab 4.0 migration

krassowski:
## References

Related to https://github.com/jupyter-lsp/jupyterlab-lsp/issues/917.

Features:
- [x] signature (works)
   - [x] active argument highlighting
- [x] hover (works)
- [x] completion
   - [ ] users now need to manually enable "autocompletion" and "documentation panel" in settings - and there are two sets of completer settings; we probably should probably override the default one.
   - [x] start/end ranges are set as expected
   - [ ] model and renderer precedence:
       - wait for https://github.com/jupyterlab/jupyterlab/pull/14800
       - two upstream issues:
          - https://github.com/jupyterlab/jupyterlab/issues/15016
          - https://github.com/jupyterlab/jupyterlab/issues/15017
   - [x] results are correctly filtered (with subject to the above
   - [x] completer CSS class
   - [ ] continuous hinting detection based on trigger characters:
      - https://github.com/jupyterlab/jupyterlab/issues/14806
      - https://github.com/jupyterlab/jupyterlab/pull/15015
- [x] diagnostics
   - [x] integration
   - [x] tags (e.g. deprecated or unused) - upstream [question](https://discuss.codemirror.net/t/additional-severity-level-for-linting-and-custom-class-names/6769/1)
   - [x] robustness - upstream issue
   - [x] fix tooltip positioning https://github.com/jupyterlab/jupyterlab/issues/14857
   - [x] enable gutter?
   - [x] update CSS for new class name
   - [x] tooltip dark mode styling is off
   - [ ] bump `@codemirror/lint` version in JupyterLab
- [x] rename
- [x] highlights
- [x] syntax (implemented by swapping mimeType)
- [x] jump to (works, needs more testing)

General:
- [x] disabling features
- [x] statusbar
- [x] settings
   - [ ] when settings are open during startup the form renders generic fields; this may be resolve in lab 4.1 as we may be able to remove delayed transformation of settings

Testing/CI:
- [x] builds
- [ ] tests pass:
   - [x] unit
   - [ ] integration
- [x] documentation
   - [x] builds
   - [ ] updated

Nice to have:
- [x] re-enable immutable installs for yarn 3 (the problem is that hash for `file:` protocol includes any rubbish and needs to be recomputed on each commit; there probably should not be a hash at all, and indeed `portal:` does not have it but is not supported by learna; official recommendation is to use `workspace:`, see [comment](https://github.com/yarnpkg/berry/issues/1871#issuecomment-697986774) which we did. Related [issue](https://github.com/yarnpkg/berry/issues/3356) and [SO](https://stackoverflow.com/questions/49501749/is-it-possible-to-ignore-the-dependency-hash-validation-of-just-one-module-or-r).
- [ ]  upstream: should  `ICompletionContext.widget` be `IDocumentWidget` rather than `Widget`? (because we want to access `.context.path`); depends on where we believe completer will be used
- [ ] add a default type for `WidgetLSPAdapter<T>` generic to avoid `WidgetLSPAdapter<any>`

Upstream PRs:
- https://github.com/jupyterlab/jupyterlab/pull/14786 (required for custom settings)
- https://github.com/jupyterlab/jupyterlab/pull/14712 (needed to remove `@ts-ignore`s on non-exposed methods)
- https://github.com/jupyterlab/jupyterlab/pull/14714 (needed to catch up with fixes here, including no shadow FS for pyright)
- https://github.com/jupyterlab/jupyterlab/pull/14858
- https://github.com/jupyterlab/jupyterlab/pull/15015

Issues:
- https://github.com/jupyterlab/jupyterlab/issues/14711
- https://github.com/jupyterlab/jupyterlab/issues/14805
- https://github.com/jupyterlab/jupyterlab/issues/14710
- https://github.com/jupyterlab/jupyterlab/issues/14783
- https://github.com/jupyterlab/jupyterlab/issues/14806
- https://github.com/jupyterlab/jupyterlab/issues/15016
- https://github.com/jupyterlab/jupyterlab/issues/15017

## User-facing changes

- Diagnostics gutter can be enabled in settings (this is experimental, just exposing what `@codemirror/linter` gives us out of the box)
- Highlights now distinguish between read (variable access)/write (variable definition)/text (fuzzy matches) - please feed back on styling:
   Screenshot from 2023-07-08 14-39-29
- To enable autocomplete two settings need to be enabled, see https://github.com/jupyterlab/jupyterlab/issues/15022
- But autocomplete does not work yet (PR is open - https://github.com/jupyterlab/jupyterlab/pull/15015)
The following settings got removes:
- `disableCompletionsFrom` and `kernelResponseTimeout` - these needs to be implemented in JupyterLab itself, see [discussion in this upstream issue](https://github.com/jupyterlab/jupyterlab/issues/15022#issuecomment-1693538481)
- `showDocumentation` - an identical setting is now present in core JupyterLab

- `.lsp-completer` class got removed and instead `body[data-lsp-completer-layout]` selector is being used; this is suboptimal CSS performance-wise and may be changed.
- feature registration is now handled by `ILSPFeatureManager` from upstream `@jupyterlab/lsp` package


hbcarlos:
There is a new release of JupyterLab. See: https://github.com/jupyterlab/jupyterlab/releases/tag/v4.0.3

packages/jupyterlab-lsp/src/features/syntax_highlighting.ts
```diff
@@ -6,48 +7,70 @@ import {
...
     super(options);
-    this.virtual_document.changed.connect(this.update_mode.bind(this), this);
-    this.editors_with_active_highlight = new Set();
-  }
+    const connectionManager = options.connectionManager;
+
+    options.editorExtensionRegistry.addExtension({
+      name: 'lsp:syntaxHighlighting',
+      factory: options => {
+        const updateListener = EditorView.updateListener.of(viewUpdate => {
```
krassowski, line 48: Consider using `EditorState.transactionExtender` (https://codemirror.net/examples/config/) as there may be a performance benefit.

krassowski:
The `lerna run test` failures are odd -  there are no errors and it all passes locally.

krassowski:
I plan to merge this soon and iterate on remaining tasks after releasing an alpha. The robot tests heavily rely on diagnostics including the `title` in DOM but this is now what `lint` extension in CM6 does. One way forward would be to rely on diagnostics panel and only check for presence/absence of diagnostic ranges in the editor as needed.

krassowski:
Merging as it is getting uncomfortably long:
- unit tests, lint, smoke tests pass
- works locally
- all remaining items are tracked in issues (here or upstream)

PR #952, opened 2023-07-17
fix: handle CORS OPTIONS request

zhanba:
when deploy jupyter-lsp on server, the GET request may need handle the CORS problem, which means handle OPTIONS request properly.

## Code changes

- handle CORS Options request in the same way with jupyter-server

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/python_packages/jupyter_lsp/jupyter_lsp/handlers.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/handlers.py
@@ -75,6 +75,21 @@ class LanguageServersHandler(BaseHandler):
 
         self.finish(response)
 
+    def options(self, *args, **kwargs):
+        """Get the options."""
+        self.log.warning("handle options request: %s", self.request.path)
+        if "Access-Control-Allow-Headers" in self.settings.get("headers", {}):
+            self.set_header(
+                "Access-Control-Allow-Headers",
+                self.settings["headers"]["Access-Control-Allow-Headers"],
+            )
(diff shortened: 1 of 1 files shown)


python_packages/jupyter_lsp/jupyter_lsp/handlers.py
```diff
@@ -75,6 +75,21 @@ async def get(self):
 
         self.finish(response)
 
+    def options(self, *args, **kwargs):
```
krassowski, line 78: Since this is copy-pasted from [`APIHandler`](https://github.com/jupyter-server/jupyter_server/blob/c1d689e05169ba47e630ccaa39cebb4f847b3562/jupyter_server/base/handlers.py#L718C7-L719) could we just inherit from it? Any potential drawbacks?

zhanba: inherit from APIHandler seems like the right and clean way. I have test on my case and it works fine.

zhanba:
@krassowski can we release a new version of jupyter-lsp for this change?

krassowski:
@zhanba thanks for the patients, a new release of jupyter-lsp v2.2.1 is out (together with jupyterlab-lsp [v5.0.1](https://github.com/jupyter-lsp/jupyterlab-lsp/releases/tag/v5.0.1))

PR #955, opened 2023-08-18
Use shutil.which result to run npm

jameshurst:
## Code changes

Uses the result of `shutil.which` to run npm when locating the global prefix.

This fixes retreiving the npm global prefix on Windows. On Windows the npm command is actually named `npm.CMD`, but `shutil.which` is able to successfully resolve `npm` to `npm.CMD`. By using the result of `shutil.which` to run npm we are able to account for the Windows specific suffix.

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry


krassowski (approved):
Thank you!

Issue #556, opened 2021-03-15
Support Workspace


bollwyvl:
## Elevator Pitch

Provide a view of tokens defined in the current Document and/or the workspace.

## Motivation

A number of our already-supported language servers support [workspace](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_symbol) and [document](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentSymbol) symbols.

## Design Ideas

- Add the messages
- Add a sidebar/main area widget with search, etc

PR #970, opened 2023-08-29
Contribute symbols to ToC

krassowski:
## References

This is an exploration for #556

Limitations:
- does not replace default ToC sources for python/markdown/latex - changes upstream needed
- only for files for now, again, we would need to have a way to add a second section in ToC

An argument against merging it for 5.0 is that the code is so minimal (after removing fluff) that maybe we could just contribute it upstream directly? Something to talk about on the dev meeting.

CC @fcollonval

## Code changes

Adds new plugin

## User-facing changes

image

None


fcollonval:
:+1: to get this upstreamed

---

What do you mean by _add a second section in ToC_?
You would like to get two tree views; one from LSP symbols and one from (?). Why not using only one; for file editor I don't see what would be the difference between the poor existing one and the one coming from LSP symbols. The existing ones could be kept as fallback if no LSP server can provide symbols.

krassowski:
> What do you mean by add a second section in ToC?

Something like LibreOffice having "Headings" and "Tables":

image

For us, in context of Notebook this would be "Headings" and "Code Symbols". I would think this could be implemented using collapsible panels like the debugger/extension manager sections. Does it make sense?

fcollonval:
> For us, in context of Notebook this would be "Headings" and "Code Symbols". I would think this could be implemented using collapsible panels like the debugger/extension manager sections. Does it make sense?

Yes thanks a lot for the snapshot.

PR #978, opened 2023-09-08
Prepare extension and CI for release candidate

### LINKED ISSUE #961: Robot tests need porting to JupyterLab 4.0 (or replacing with galata)

### LINKED ISSUE #984: Syntax highlighting broken if multiple cells are pasted
## Description

Syntax highlighting broken if multiple cells are pasted. This will likely be fixed with new API for LSP editor integration in JupyterLab 4.1.

## Reproduce

paste-breaks-syntax

```
Failed to load language for 'text/x-ipython'. TypeError: Cannot read properties of null (reading 'getCursorPosition')
    at signature.ts:316:42
    at Os.update (4986.a497cdda4b7152902568.js?v=a497cdda4b7152902568:1:120638)
    at Os._dispatch (4986.a497cdda4b7152902568.js?v=a497cdda4b7152902568:1:117846)
    at Os.dispatch (4986.a497cdda4b7152902568.js?v=a497cdda4b7152902568:1:118749)
    at jlab_core.5f8e30c3b3cb076c81d8.js?v=5f8e30c3b3cb076c81d8:1:335113
```

## Expected behavior

Works

## Context

- Browser and version: Chrome
- JupyterLab version: 4.0.5

krassowski:
## References

Fixes #961

## Code changes

New and renamed keywords:
- "Click Token"
- "Open Context Menu Over Token"
- "Mouse Over Token"
- "Mouse Over And Wiggle" → "Mouse Over Token And Wiggle"
- "Mouse Over With Control" → "Mouse Over Token With Control"

Type comments are no more.

Removed no longer needed CI pins :tada:

Remaining completer bugs:
- [x] completer width increases when typing when panel is documentation open (most likely upstream bug, possibly induced by the recent PR that changed when that width is counted in, needs better tests upstream)
   - upstream issue: https://github.com/jupyterlab/jupyterlab/issues/15131#issue-1899582006
- [x] completion reconciliator fails due to insertText getter not being respected due to spread syntax (upstream, but I have workaround for now)
- [x] wrong offset in tokens (fixed locally)
- [x] elide in paths
- [ ] documentation rendering should be decided per-item and delegated to respective providers (upstream)
   - [ ] open issue upstream to discuss
   - [x] workaround by just calling parent renderer for non-LSP items
- [ ] `processedItemsCache`
   - with new changes upstream it is probably best to defer
- [x] caching result of `resolve` request
- [x] completion in paths test highlights a bug
- [x] highlights error in console when tabbing

## User-facing changes

- fixes diagnostics not showing up in text editor in certain circumstances
- fixes signature hover box not being clickable
- improves performance by not creating a temporary editor to setup linter underline style
- JSON overrides work again
- https://github.com/jupyter-lsp/jupyterlab-lsp/pull/978/commits/d08a9d8c1464bafe7ead4ffb01c0ec1522d03fb6 fixes #984

None


krassowski:
Down to 21 failing tests:
- 12 in completer (we need one more PR in and a new release upstream)
- 'Hover Does Not Trigger Automatically' - fails on test setup, passes locally
- 'Robot Framework' depends on https://github.com/MarketSquare/jupyterlab_robotmode/issues/14
- 'Docker' - rename fails, I can repro locally - it looks like the tokens are not detected properly
- 'Python' in 'Configuration of language servers' - flaky? fails on tear-down with `"ElementClickInterceptedException: Message: Element <div class="jp-icon-hover lm-TabBar-tabCloseIcon fvaq30"> is not clickable at point (489,362) because another element <code> obscures it"`
- 'Diagnostics Panel Works After Removing Foreign Document' - fix incoming
- 'Columns Can Be Hidden' - passes locally
- 'Jumps To References With Modifier Click' - reproduced locally
- 'Jumps To References From Context Menu' - reproduced locally
- 'Diagnostics Panel Works After Kernel Restart' - reproduced locally

krassowski:
- 'Jumps To References With Modifier Click' is down to an upstream issue: https://github.com/python-lsp/python-lsp-server/issues/439, PR to fix it is now open.

krassowski:
- `Docker` failure is an upstream issue: https://github.com/jupyterlab/jupyterlab/issues/15104

krassowski:
This is down to 15 failing tests, 4 of which are upstream issues.

PR #981, opened 2023-09-14
Support diagnostics panel in notebook

krassowski:
## References

An attempt to display diagnostics panel in notebook v7

## Code changes

Import `NotebookApp` and execute appropriate command.

## User-facing changes

TBD

None

### DIFF
--- a/packages/jupyterlab-lsp/src/features/diagnostics/index.ts
+++ b/packages/jupyterlab-lsp/src/features/diagnostics/index.ts
@@ -1,4 +1,4 @@
-import { INotebookShell } from '@jupyter-notebook/application';
+import { INotebookShell, NotebookApp } from '@jupyter-notebook/application';
 import {
   ILayoutRestorer,
   JupyterFrontEnd,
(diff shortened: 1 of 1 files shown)


packages/jupyterlab-lsp/src/features/diagnostics/index.ts
```diff
@@ -99,20 +99,27 @@ export const DIAGNOSTICS_PLUGIN: JupyterFrontEndPlugin<IDiagnosticsFeature> = {
...
-            });
+            if (
+              typeof NotebookApp !== 'undefined' &&
+              app instanceof NotebookApp
+            ) {
+              app.shell.add(panelWidget, 'right');
+              await app.commands.execute('application:toggle-panel', {
+                id: panelWidget.id,
+                side: 'right'
+              });
```
krassowski, line 110: This does not seem to be enough to add a Notebook panel from extension :crying_cat_face:. I am yet to investigate further, but in the meantime, @jtpio do you have thoughts on this one (given our [other conversation](https://github.com/jupyterlab/jupyterlab/pull/14920#discussion_r1350250007) on lab-notebook shell interfaces divergence)?

bollwyvl: Hmmm...

```
app instanceof NotebookApp
```

This is probably _not_ going to work, long term, as it means the notebook packages will be bundled along with the already-large (and duplicated) `jupyterlab-lsp` assets, and could very well return a false negative if webpack is feeling grumpy. 

Making it a `sharedPackage` _might_ improve things, but would really need to be in an entirely separate plugin file with a naked (not `type`) `import`...

Further, this is often not enough information, as one still has to consult the `PageConfig.getOption('notebookPage')`, to determine whether a given notebook page should _not_ be able to get a panel (e.g. `tree`, `terminal`)... it would actually be lovely if this list of documented string constants (_not_ as a TypeScript enum, for the same reason above) were hoisted as a dedicate package.

Alas, we kinda missed the boat on changing the `IFrontendShell` interface to provide a dumb, string-capability-based model on the major version revs.

PR #1007, opened 2023-11-04
Fix false “undefined name” message caused by cell magic

### LINKED ISSUE #635: Support %%time
## Elevator Pitch

Code defined in a `%%time` cell doesn't appear to be recognised so you then get spurious warnings about e.g. undefined variables.

image

## Motivation

Spurious warnings detract from the utility of a code linter

## Design Ideas

I think this is a hard problem to solve as shown by other issues about magic commands.  All I can think of is enumeratig a bunch of special cases 🤢.  Maybe that's not so bad though - just like a `.gitignore` file enumerates a bunch of patterns to ignore, `jupyterlab-lsp` could have a configuration for magics which allowed a user to specify what it should do (e.g. ignore). This would then cater for all custom user-defined magics as well.

i-aki-y:
## References

In this PR, I want to fix #635. 

## Code changes

According to the comment in #635, I implemented another transformation to the specific cell magics.

From:
```py
%%capture --no-display
x = 1
```
To:
```py
# START_CELL_MAGIC("capture", "--no-display")
x = 1
# END_CELL_MAGIC
```
Instead of:
```py
get_ipython().run_cell_magic("capture", "--no-display", """x=1""")
```

Currently, the target cell magics are hard coded (%%time, %%capture).
It would be better if the list is configurable.
Since I'm new to the project, I need more time to figure out how to make the list configurable, but I will try this in the near future.

## User-facing changes

As pointed out in #635, variables defined inside cell magics are not recognized by the lsp, and `undefined name` message shows up.
This PR tries to remove these false messages.

No

## Chores

- [x] linted 
- [x] tested 
- [ ] documented 
- [x] changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1101,8 +1101,14 @@ Requires JupyterLab `>=3.6.0,<4.0.0a0` and Python 3.8 or newer.
 ### `jupyter-lsp 0.6.0b0`
 
 - features
+
   - starts language servers on demand
   - accepts configuration via Jupyter config system (traitlets) and python
     `entry_point`s
   - autodetects language servers for bash, CSS, LESS, SASS, Dockerfile, YAML, JS,
     TypeScript, JSX, TSX, JSON, YAML
+
+- bugfixes
+  - fix issue that variables declared in cell magics(%%time, %%capture) are masked(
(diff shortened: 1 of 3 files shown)


krassowski (approved):
Thank you @i-aki-y for the PR (and patience)

PR #1015, opened 2023-11-20
Resolve traitlets type warnings, lint, remove `six`

krassowski:
- Fixes a few mypy/traitlets interaction issues due to new traitlets version as seen in #1007 and others.
- Removes implicit `six` dependency and `types-six` requirement for linting
- Lints robot scripts too

### DIFF
--- a/atest/diagnostics.py
+++ b/atest/diagnostics.py
@@ -1,59 +1,59 @@
 from functools import partial
+
 from robot.libraries.BuiltIn import BuiltIn
+from robot.utils import timestr_to_secs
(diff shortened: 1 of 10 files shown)


krassowski:
Looks like `bash-language-server` needs pinning or updating startup script as it is timing out.

krassowski:
Or maybe the typing warnings are right and this is just one of the breaking changes in latest `traitlets`?

krassowski:
Merging as it unblocks CI. The pypy smoke failure is tracked in https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1018.

PR #1022, opened 2023-11-25
Do not pre-filter completions from kernel

krassowski:
## References

- fixes https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1014
- fixes https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1009
- fixes https://github.com/jupyter-lsp/jupyterlab-lsp/issues/704

## Code changes

Moves pre-filtering to LSP model which now checks if the item came from LSP

### DIFF
--- a/packages/jupyterlab-lsp/src/features/completion/model.ts
+++ b/packages/jupyterlab-lsp/src/features/completion/model.ts
@@ -59,50 +59,6 @@ export class GenericCompleterModel<
     return item as T;
   }
 
-  setCompletionItems(newValue: T[]) {
-    super.setCompletionItems(newValue);
-
-    if (this.current && this.cursor) {
-      // set initial query to pre-filter items; in future we should use:
-      // https://github.com/jupyterlab/jupyterlab/issues/9763#issuecomment-1001603348
-
-      // note: start/end from cursor are not ideal because these get populated from fetch
-      // reply which will vary depending on what providers decide to return; we want the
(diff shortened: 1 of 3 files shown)


packages/jupyterlab-lsp/src/features/completion/model.ts
```diff
@@ -138,7 +94,11 @@ export class GenericCompleterModel<
     return super.createPatch(patch);
   }
 
-  private _sortAndFilter(query: string, items: T[]): T[] {
+  resolveQuery(userQuery: string, _item: T) {
```
krassowski, line 97:
```suggestion
  protected resolveQuery(userQuery: string, _item: T) {
```

packages/jupyterlab-lsp/src/features/completion/model.ts
```diff
@@ -267,13 +224,73 @@ type MaybeCompletionItem = Partial<CompletionItem> &
...
+        query = query.substring(0, -1);
+        trimmedQuotes = true;
+      }
+      if (this.settings.preFilterMatches || trimmedQuotes) {
+        this._preFilterQuery = query;
+      }
+    }
+  }
+
+  resolveQuery(userQuery: string, item: MaybeCompletionItem) {
```
krassowski, line 286:
```suggestion
  protected resolveQuery(userQuery: string, item: MaybeCompletionItem) {
```

PR #1046, opened 2024-02-04
Prepare 5.0.3 release

krassowski:
This is a tiny patch release with two bug fixes.


krassowski:
It would be nice to decouple running UI tests from server tests and coverage requirement as currently these do not get to run due to coverage short-circuiting the run.

PR #1057, opened 2024-02-23
Build against JupyterLab 4.1, fix tests

krassowski:
## References

Builds upon https://github.com/jupyter-lsp/jupyterlab-lsp/pull/1052

## Code changes

- uses new LSP `extensionFactory` API (JupyterLab 4.1) to remove workarounds which we used to make it kind of work in Lab 4.0.
- adds a workaround for upstream issue https://github.com/jupyterlab/jupyterlab/issues/15868
- reduces some warnings in places known to have false positives to debug statements

## User-facing changes

- requires JupyterLab 4.1
- the use of new new LSP `extensionFactory` API means that:
   - LSP features work more reliably as the correct virtual document and editor is known rather than found from heuristics
   - LSP CodeMirror extensions are not added to editors which do not have an LSP-enabled editor reducing spurious warnings and improving performance
   - LSP CodeMirror extensions are only added once the LSP server was connected

None

## Chores

- [x] linted 
- [x] tested 
- [ ] documented 
- [ ] changelog entry


.yarnrc.yml
```diff
@@ -1 +1,24 @@
+enableImmutableInstalls: false
+enableInlineBuilds: false
+enableTelemetry: false
+httpTimeout: 60000
 nodeLinker: node-modules
+npmRegistryServer: 'https://registry.npmjs.org'
+installStatePath: ./build/.cache/yarn/install-state.gz
+#globalFolder: ./build/.cache/yarn
```
krassowski, line 8: I guess this should not be the same as cache folder.

packages/jupyterlab-lsp/src/virtual/document.ts
```diff
@@ -36,7 +36,9 @@ export class VirtualDocument extends VirtualDocumentBase {
     );
     // override private `chooseForeignDocument` as a workaround for
     // https://github.com/jupyter-lsp/jupyterlab-lsp/issues/959
-    this['chooseForeignDocument'] = this._chooseForeignDocument;
+    const anyThis = this as any;
+    anyThis._chooseForeignDocument = anyThis.chooseForeignDocument =
+      this.__chooseForeignDocument;
```
krassowski, line 41: TODO: drop it, we do not need to override it anymore

packages/jupyterlab-lsp/src/virtual/document.ts
```diff
@@ -153,14 +155,13 @@ export class VirtualDocument extends VirtualDocumentBase {
     const usedDocuments = new Set<VirtualDocument>();
```
krassowski, line 155: `closeExpiredDocuments` should be good to remove too

packages/jupyterlab-lsp/webpack.config.js
```diff
@@ -10,5 +10,10 @@ module.exports = {
         use: ['source-map-loader']
       }
     ]
+  },
+  resolve: {
+    fallback: {
+      util: false
+    }
```
krassowski, line 17: We can drop it, `@rjsf/utils` 5.17.1 fixed it

packages/metapackage/package.json
```diff
@@ -36,7 +36,7 @@
     "@jupyter-lsp/theme-vscode": "workspace:*"
   },
   "devDependencies": {
-    "rimraf": "^3.0.0",
+    "rimraf": "^5.0.5",
     "typescript": "~5.0.4"
```
krassowski, line 40: 5.1.6

krassowski:
All Windows tests are failing with `After 10.0s, D:\a\jupyterlab-lsp\jupyterlab-lsp\build\reports\windows_38\atest\2\logs\lab.log.txt did not contain: ('The following Language Servers will be available',)`

krassowski:
So we are back at mostly green except for Windows + Python 3.8:

Screenshot from 2024-02-25 10-12-37

I will merge this and create follow-up issues to tackle things to address. Thanks @bollwyvl for starting this!

PR #1062, opened 2024-02-26
Minor version/debug statement cleanup in preparation for v5.1

krassowski:
Follow-up to https://github.com/jupyter-lsp/jupyterlab-lsp/pull/1061 - the two packages also needs bumping for npm upload


krassowski:
The failure of smoke tests is expected - it should pass once released.

PR #1064, opened 2024-03-01
Update `script` for `typescript-language-server`

jtpio:
## References

Looking into using `typescript-language-server` and following the [docs to install it](https://jupyterlab-lsp.readthedocs.io/en/latest/Language%20Servers.html#example-getting-all-the-node-js-based-language-servers), it looks like `jupyter_lsp` was not able to find the language server:

image

Updating the script seems to be fixing it:

image

## Code changes

Update the `script` for `typescript-language-server`.

## User-facing changes

Users can install and use `typescript-language-server` with JupyterLab LSP.

This supports the latest version of `typescript-language-server`, but might break previous versions when the file was still `cli.js`?

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "source-map-loader": "^4.0.1",
     "sql-language-server": "^1.2.1",
     "typescript": "~5.0.3",
-    "typescript-language-server": "^0.6.4",
+    "typescript-language-server": "^4.3.3",
     "unified-language-server": "^0.3.0",
     "vscode-css-languageserver-bin": "^1.4.0",
     "vscode-html-languageserver-bin": "^1.4.0",
--- a/python_packages/jupyter_lsp/jupyter_lsp/specs/typescript_language_server.py
+++ b/python_packages/jupyter_lsp/jupyter_lsp/specs/typescript_language_server.py
@@ -4,7 +4,7 @@ from .utils import NodeModuleSpec
 
 class TypescriptLanguageServer(NodeModuleSpec):
     node_module = key = "typescript-language-server"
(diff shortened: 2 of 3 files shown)


krassowski:
The test failures are relevant. It looks the location of the binary entry point has changed last year and we are pinning to an old version. I think we should bump the version, but also consider using the [`bin` field from `package.json`](https://github.com/typescript-language-server/typescript-language-server/blob/b224b878652438bcdd639137a6b1d1a6630129e4/package.json#L18-L20).

jtpio:
> I think we should bump the version

Done in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/1064/commits/f014c9e9aaf58cc37e7c3b5abfd455f31d0e12a8 :+1: 



> but also consider using the [`bin` field from `package.json`](https://github.com/typescript-language-server/typescript-language-server/blob/b224b878652438bcdd639137a6b1d1a6630129e4/package.json#L18-L20).

That makes sense, although there might be several scripts available, and the name might not match the name of the module itself?  I guess this would apply to other specs derived from `NodeModuleSpec`, so maybe something to track in a separate issue?

jtpio:
> maybe something to track in a separate issue?

Opened https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1066

krassowski (approved):
Thanks @jtpio and thanks for opening the issue!

krassowski:
@jtpio this is available in jupyter-lsp 2.2.4 - thanks again!

PR #1101, opened 2024-10-06
Fixed broken Julia LSP support + fixed broken MarkDown LSP support

alberti42:
## References

- It fixes MarkDown support that was broken. It replaces the broken `unified-language-server` with `vscode_markdown_languageserver`. See issues raised at https://discourse.jupyter.org/t/how-to-install-and-run-markdown-language-server/24286/3
- It fixes support for Julia LSP. There were several mistakes in how to LSP server was called; see notes in the commits.

## Code changes

Changes are documented in the commits.

## User-facing changes

No visual changes in the settings.

No changes to API.

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry 

I am glad to contribute to all chores if the PR will be considered and the bugs eventually fixed upstream. Before investing more time, I prefer to wait for some reactions from the maintainers. Thanks for your understanding.

Changes are anyhow very limited and small in their extent.


python_packages/jupyter_lsp/jupyter_lsp/types.py
```diff
@@ -213,17 +213,18 @@ class LanguageServerManagerAPI(LoggingConfigurable, HasListeners):
         help=_("additional absolute paths to seek node_modules first"),
     ).tag(config=True)
 
-    def find_node_module(self, *path_frag):
+    def find_node_module(self, node_module, alternatives):
```
krassowski, line 216: I think this constitutes a breaking change as language specs inheriting from `NodeModuleSpec` will no longer work. One way to add it in backward compatible way would be to add a new method with new signature and make the old method call the new one.

krassowski:
Thank you for the PR, sorry I missed it back in October!

PR #1112, opened 2025-01-08
Fix usage of `conda-incubator/setup-miniconda`

jtpio:
## References

Looking into https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1111

## Chores

- [ ] linted 
- [ ] tested 
- [ ] documented 
- [ ] changelog entry

### DIFF
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -62,7 +62,6 @@ jobs:
         uses: conda-incubator/setup-miniconda@v3
         with:
           environment-file: requirements/github-actions.yml
-          miniforge-variant: Mambaforge
           use-mamba: true
 
       - name: Install linting dependencies
@@ -141,7 +140,6 @@ jobs:
       - name: Set up Python and mamba
         uses: conda-incubator/setup-miniconda@v3
         with:
-          miniforge-variant: Mambaforge
           use-mamba: true
 
       - name: Install minimal build deps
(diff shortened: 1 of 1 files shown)


.github/workflows/job.test.yml
```diff
@@ -333,10 +328,6 @@ jobs:
           path: ./build/reports
         if: always()
 
-      - name: Rename uncached conda packages
```
jtpio, line 336: Not sure why this was needed but it was failing. Also I would expect something like this to be handled by the `conda-incubator/setup-miniconda` action directly.

jtpio:
Stopping here since the "Set up Python and mamba" steps using `conda-incubator/setup-miniconda@v3` now seems to be working again.

The remaining failures seem to be related to some robot stuff, no idea

.github/workflows/job.test.yml
krassowski: It looks like it was added in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/882, more specifically in https://github.com/jupyter-lsp/jupyterlab-lsp/pull/882/commits/a4fb737e2f2bda4bbcd3229a059c8c1f89af6e63

krassowski (approved):
Thanks!

jtpio:
> I'am not sure if I remember how to debug robot failures, but many of these look like changes in language servers, and a few of them might be due to changes in newer lab versions. Certainly can be addressed separately :)

If we want to have `jupyterlab-lsp` be a bit closer to JupyterLab (https://github.com/jupyterlab/frontends-team-compass/issues/67#issuecomment-2578363269), maybe we should consider switching to Playwright / Galata, since this is what all other lab projects are using.

PR #1114, opened 2025-01-13
Don't test for cursor position on `mousemove` if the widget adapter is not connected

### LINKED ISSUE #1113: Noisy log message
## Description

On text file with no available language server, a log message is emitted on mousemove event:

https://github.com/user-attachments/assets/13925c4c-3eb2-466d-b30a-159d9b8ac985

## Reproduce

1. Install the extension
2. Open a plain text file
3. Move your mouse on it

## Expected behavior

No log message

## Context

The log message is coming from that event callback

https://github.com/jupyter-lsp/jupyterlab-lsp/blob/f5d2b9ee99828a96e3afa2d575ce390966b69368/packages/jupyterlab-lsp/src/features/hover.ts#L197-L198

I guess we could bail early. The question is on what condition. I see two possibilities:
- Bail early if the `adapter.isConnected` is false
- Bail early if the `adapter.virtualDocument.standalone` is true (as the event handler is only for notebook according to the comment)

What would be the best fix?

> The second way would also prevent doing much actions on the file editor on a very recurrent dom event.

fcollonval:
## References

Fixes #1113

## Code changes

Bail early on mouse motion DOM event handler if the widget adapter is not connected.
This avoid doing DOM computation and displaying noisy log file on file not supporting LSP.

## User-facing changes

**Before**

https://github.com/user-attachments/assets/13925c4c-3eb2-466d-b30a-159d9b8ac985

**After**

https://github.com/user-attachments/assets/fcfb462f-e6f4-4b10-a182-a046b727f27d

None

## Chores

- [x] linted 
- [ ] tested 
- documented 
- [x] changelog entry

### DIFF
--- a/.github/workflows/job.test.yml
+++ b/.github/workflows/job.test.yml
@@ -62,7 +62,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v3
         with:
           environment-file: requirements/github-actions.yml
-          mamba-version: "*"
+          mamba-version: '*'
           channels: conda-forge,nodefaults
           channel-priority: true
 
@@ -138,7 +138,7 @@ jobs:
       - name: Set up Python and mamba
         uses: conda-incubator/setup-miniconda@v3
         with:
-          mamba-version: "*"
+          mamba-version: '*'
           channels: conda-forge,nodefaults
           channel-priority: true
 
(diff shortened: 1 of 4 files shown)


krassowski (approved):
Thanks @fcollonval!

PR #1131, opened 2025-06-25
Fix URL in docs

yangdanny97:
The old link is dead, edited from the Github web UI


krassowski (approved):
Thank you @yangdanny97!

PR #1134, opened 2025-06-27
Fix links in documentation

yangdanny97:
Summary of changes:
- fix anchor names/capitalization in links (they are always kebab case, lowercased)
- change relative references to point to .ipynb (.html relative references do not generate correctly, it seems)
- move `Contributing.md` contents into `Contributing.ipynb`. If we `{include}` the former into the latter, the documentation links will not work due to the extra `./docs/` prefix. I don't think there's an easy way to write links that will work for both, so I just moved it.

## References

closes https://github.com/jupyter-lsp/jupyterlab-lsp/issues/1132

## Chores

- [x] linted 
- [x] tested (python scripts/docs.py and clicked around the built docs)
- [x] documented 
- [ ] changelog entry


krassowski:
Yes, we need to fix some tests or mark them as XFail for now.

